PR #24, opened 2016-03-11
Created aikido_perception package

Shushman:
This is the first attempt at a perception module for AIKIDO.

- Introduction of `aikido_perception` as a package
- Abstract class in `PerceptionModule.hpp`
- Implementation in `AprilTagsModule.hpp/cpp`

---
This change is [Reviewable](https://reviewable.io/reviews/personalrobotics/aikido/24)


aikido_perception/src/AprilTagsModule.cpp
```diff
@@ -0,0 +1,187 @@
...
+			temp_pose.matrix() = frame_pose * marker_pose;
+			Eigen::Isometry3d skel_pose;
+			skel_pose.matrix() = temp_pose * skel_offset;
+			Eigen::Isometry3d link_offset = mReferenceLink->getWorldTransform();
+			skel_pose = link_offset * skel_pose;
+
+			//Check if skel in skel_list
+			//If there then update pose
+			//If not then append skeleton
+			std::string skel_name = body_name.substr(0,body_name.find('.'));
```
Shushman: Changed this so that the `YAML` file has an additional `resource` field for each tag, which points to the URDF file of the model. The `name` field now just has the object name. This obviates the need for 1) A URDF folder path  and 2) Deriving the object name from the filename.

aikido_perception/include/aikido/perception/ConfigDataLoader.hpp
```diff
@@ -0,0 +1,84 @@
...
+#include <aikido/util/CatkinResourceRetriever.hpp>
+#include <dart/common/Console.h>
+#include <dart/common/LocalResourceRetriever.h>
+#include <Eigen/Geometry>
+#include <stdexcept>
+
+namespace aikido{
+namespace perception{
+
+class ConfigDataLoader
```
Shushman: Made `ConfigDataLoader` an interface and created an instantiation called `YamlFileLoader`

aikido_perception/src/AprilTagsModule.cpp
```diff
@@ -0,0 +1,187 @@
...
+
+//================================================================================================================================
+void AprilTagsModule::detectObjects(std::vector<dart::dynamics::SkeletonPtr>& skeleton_list,double _timeout, ros::Time timestamp)
+{
+	//Looks at all detected tags, looks up config file 
+	//Appends new skeletons to skeleton list
+	visualization_msgs::MarkerArrayConstPtr marker_message
+ 			= ros::topic::waitForMessage<visualization_msgs::MarkerArray>(mMarkerTopic,mNode,ros::Duration(_timeout));
+
+	for(size_t i=0; i < marker_message->markers.size(); i++)
```
Shushman: Done

aikido_perception/src/AprilTagsModule.cpp
```diff
@@ -0,0 +1,187 @@
...
+	{
+		//Get marker, tag ID, and detection transform name 
+		visualization_msgs::Marker marker_transform = marker_message->markers[i];
+		ros::Time marker_stamp = marker_transform.header.stamp;
+		std::string tag_name = marker_transform.ns;
+		std::string detection_frame(marker_transform.header.frame_id);
```
Shushman: Also done.

aikido_perception/src/AprilTagsModule.cpp
```diff
@@ -0,0 +1,187 @@
...
+			  	}
+			);
+
+			dart::dynamics::SkeletonPtr skel_to_update;
+
+			if (it == std::end(skeleton_list)){
+				//New skeleton
+				is_new_skel = true;
+				std::string body_path(mUrdfPath);
+				body_path.append(body_name);
```
Shushman: Done, as mentioned earlier.

aikido_perception/src/AprilTagsModule.cpp
```diff
@@ -0,0 +1,187 @@
...
+				dart::utils::DartLoader urdfLoader;
+				
+				skel_to_update = 
+					urdfLoader.parseSkeleton(body_path,mResourceRetrieverPtr);
+				
+				if(!skel_to_update)
+					std::cout<<"Empty skel ptr!"<<std::endl;
```
Shushman: Removed, and replaced with warning.

aikido_perception/src/AprilTagsModule.cpp
```diff
@@ -0,0 +1,187 @@
...
+			tf::TransformListener listener;
+
+			try{
+				listener.waitForTransform(mDestinationFrame,detection_frame,
+					ros::Time(0),ros::Duration(_timeout));
+
+				listener.lookupTransform(mDestinationFrame,detection_frame,
+					ros::Time(0), transform);
+			}
+			catch(tf::TransformException ex){
```
Shushman: Changed.

aikido_perception/src/AprilTagsModule.cpp
```diff
@@ -0,0 +1,187 @@
...
+			Eigen::Quaterniond orien(marker_transform.pose.orientation.w,
+									marker_transform.pose.orientation.x,
+									marker_transform.pose.orientation.y,
+									marker_transform.pose.orientation.z);
+
+			//Get in 4x4 pose form
+			Eigen::Vector3d qpos(marker_transform.pose.position.x,marker_transform.pose.position.y,marker_transform.pose.position.z);
+			Eigen::Matrix3d qrot(orien.toRotationMatrix());
+			Eigen::Matrix4d marker_pose;
+			marker_pose<< qrot,qpos,0,0,0,1;
```
Shushman: Made a `shape_conversions.hpp/cpp` pair of files.

aikido_perception/src/AprilTagsModule.cpp
```diff
@@ -0,0 +1,187 @@
...
+
+		if(marker_stamp < timestamp){
+			dtwarn << "Marker stamp " << marker_stamp<< " for " << tag_name 
+			<< "is behind timestamp parameter: " <<timestamp <<"\n";
+
+			continue;
+		}
+
+		std::string body_name;
+		Eigen::Matrix4d skel_offset;
```
Shushman: All math in `Isometry3d now`

aikido_perception/include/aikido/perception/yaml_conversion.hpp
```diff
@@ -0,0 +1,82 @@
...
+template < typename _Scalar, int _Rows, int _Cols, int _Options, int _MaxRows, int _MaxCols>
+    struct convert< Eigen::Matrix<_Scalar, _Rows, _Cols, _Options, _MaxRows, _MaxCols> > {
+        static Node encode(const Eigen::Matrix<_Scalar, _Rows, _Cols, _Options, _MaxRows, _MaxCols>& matrix) {
+            Node node(NodeType::Sequence);
+
+            int nValues = matrix.rows() * matrix.cols();
+
+            for (int i = 0; i < nValues; ++i) {
+                node.push_back( matrix(i / matrix.cols(), i % matrix.cols() ) );
+            }
```
Shushman: Done the change after removing the `Tag()` stuff from the `muul` `yaml_utils` file

aikido_perception/CMakeLists.txt
```diff
@@ -0,0 +1,85 @@
...
+  roscpp
+  geometry_msgs
+  tf
+  std_msgs
+  visualization_msgs
+)
+
+find_package(DART REQUIRED COMPONENTS core)
+find_package(Boost REQUIRED)
+find_package(aikido REQUIRED)
```
mkoval: You'll need to change this to:
```c++
find_package(aikido REQUIRED COMPONENTS util)
```

aikido_perception/CMakeLists.txt
```diff
@@ -0,0 +1,85 @@
...
+  ${DART_INCLUDE_DIRS}
+  ${Eigen_INCLUDE_DIRS}
+  ${aikido_INCLUDE_DIRS}
+  ${catkin_INCLUDE_DIRS}
+  ${YamlCpp_INCLUDE_DIRS}
+)
+
+link_directories(
+  ${DART_LIBRARY_DIRS}
+  ${Eigen_LIBRARY_DIRS}
```
mkoval: You don't need `DART_LIBRARY_DIRS` or `Eigen_LIBRARY_DIRS`.

aikido_perception/CMakeLists.txt
```diff
@@ -0,0 +1,79 @@
...
+
+target_link_libraries(${PROJECT_NAME}
+  ${DART_LIBRARIES}
+  ${aikido_LIBRARIES}
+  ${Boost_LIBRARIES}
+  ${YamlCpp_LIBRARIES}
+  ${tf_LIBRARIES}
+)
+
+install(TARGETS ${PROJECT_NAME}
```
mkoval: This looks good. :+1:

aikido_perception/include/aikido/perception/AprilTagsModule.hpp
```diff
@@ -0,0 +1,68 @@
...
+#include <aikido/util/CatkinResourceRetriever.hpp>
+
+#include "PerceptionModule.hpp"
+
+
+namespace aikido{
+namespace perception{
+
+/// The April Tag detector 
+class AprilTagsModule : public virtual PerceptionModule
```
mkoval: Add Doxygen docstrings to all functions in this library. For functions, they should also include arguments.

aikido_perception/include/aikido/perception/AprilTagsModule.hpp
```diff
@@ -0,0 +1,68 @@
...
+{
+public:
+	AprilTagsModule(ros::NodeHandle node, std::string markerTopic, std::shared_ptr<ConfigDataLoader> configData,
+					const dart::common::ResourceRetrieverPtr& resourceRetriever,
```
mkoval: The constructor always takes ownership of `resourceRetriever.` We decided that we should take `shared_ptr`s, like `ResourceRetrieverPtr` by value and `std::move` them into a member variable in this situation.

Sorry for the mixed message. This is the convention we're going with, at least for now.

aikido_perception/include/aikido/perception/AprilTagsModule.hpp
```diff
@@ -0,0 +1,68 @@
...
+					std::string destinationFrame, dart::dynamics::Frame* referenceLink);
+	virtual ~AprilTagsModule() = default;
+
+
+	void detectObjects(std::vector<dart::dynamics::SkeletonPtr>& skeleton_list,double _timeout=10.0, ros::Time timestamp=ros::Time::now()) override;
```
mkoval: I don't think `= ros::Time::now()` will do what you want here. I believe this is evaluated **once** when `static` variables are initialized and that value is used forever after.

aikido_perception/include/aikido/perception/AprilTagsModule.hpp
```diff
@@ -0,0 +1,68 @@
...
+	///The reference frame of HERB to transform with respect to
+	dart::dynamics::Frame* mReferenceLink;
+
+	///The pointer to the loader of configuration data
+	std::shared_ptr<ConfigDataLoader> mConfigData;
+
+	///For the ROS node that will work with the April Tags module
+	ros::NodeHandle mNode;
+
+	tf::TransformListener* mListener;
```
mkoval: This shouldn't be a raw pointer. I suggest putting it directly in the class (if possible) or using a `unique_ptr` (if not).

aikido_perception/include/aikido/perception/ConfigDataLoader.hpp
mkoval: :+1: Please add Doxygen docstrings to this class and its member functions (including all arguments).

aikido_perception/include/aikido/perception/YamlFileLoader.hpp
```diff
@@ -0,0 +1,47 @@
...
+#include <dart/dart.h>
+#include <Eigen/Geometry>
+#include "ConfigDataLoader.hpp"
+#include <stdexcept>
+
+namespace aikido{
+namespace perception{
+
+/// The YAML File configuration data loader
+class YamlFileLoader : public virtual ConfigDataLoader
```
mkoval: Missing docstrings.

aikido_perception/include/aikido/perception/yaml_conversion.hpp
mkoval: I think this has been replaced with the other YAML conversion code. If so, please delete it. If not, see my comment above. :smile:

aikido_perception/src/AprilTagsModule.cpp
```diff
@@ -0,0 +1,170 @@
...
+
+AprilTagsModule::AprilTagsModule( ros::NodeHandle node, std::string markerTopic, std::shared_ptr<ConfigDataLoader> configData,
+								  const dart::common::ResourceRetrieverPtr& resourceRetriever,	
+								  std::string referenceFrameId, dart::dynamics::Frame* referenceLink):
+		mNode(node),
+		mMarkerTopic(markerTopic),
+		mConfigData(configData),
+		mResourceRetriever(resourceRetriever),
+		mReferenceFrameId(referenceFrameId),
+		mReferenceLink(referenceLink)
```
mkoval: Please `std::move` arguments that were passed by value into member variables.

aikido_perception/src/AprilTagsModule.cpp
```diff
@@ -0,0 +1,170 @@
...
+{
+	mListener = new tf::TransformListener();
```
mkoval: See my comment in the header. Either store this by value or in a `unique_ptr`. It would also be nice to use the `TransformListener` that accepts a `NodeHandle` argument. Otherwise, the TF listener won't use the same `NodeHandle` as everything else. This has implications when using `CallbackQueue`s and `AsyncSpinner`s.

aikido_perception/src/AprilTagsModule.cpp
```diff
@@ -0,0 +1,170 @@
...
+
+//================================================================================================================================
+void AprilTagsModule::detectObjects(std::vector<dart::dynamics::SkeletonPtr>& skeleton_list,double _timeout, ros::Time timestamp)
+{
+	bool transform_lookedup = false;
+	//Looks at all detected tags, looks up config file 
+	//Appends new skeletons to skeleton list
+	do{
+		visualization_msgs::MarkerArrayConstPtr marker_message
+	 			= ros::topic::waitForMessage<visualization_msgs::MarkerArray>(mMarkerTopic,mNode,ros::Duration(_timeout));
```
mkoval: Using `waitForMessage` here may miss messages if retries are necessary. I suggest manually creating a `Subscriber` and using that instead.

I would also prefer to refactor / split this function to simplify the control flow.

aikido_perception/src/AprilTagsModule.cpp
```diff
@@ -0,0 +1,170 @@
...
+
+				try{
+					mListener->waitForTransform(mReferenceFrameId,detection_frame,
+						marker_stamp,ros::Duration(_timeout));
+
+					mListener->lookupTransform(mReferenceFrameId,detection_frame,
+						marker_stamp, transform);
+				}
+				catch(const tf::ExtrapolationException& ex){
+					ROS_ERROR("%s",ex.what());
```
mkoval: Please prefix the `ExtrapoationException` error with some context about its use.

aikido_perception/src/AprilTagsModule.cpp
```diff
@@ -0,0 +1,170 @@
...
+					continue;
+				}
+				catch(const tf::TransformException& ex){
+					ROS_ERROR("%s",ex.what());
+					throw std::runtime_error("TF Error which cannot be recovered");
```
mkoval: I don't think there is any point in catching this exception. I would just let it bubble up and require the user to handle it.

aikido_perception/src/AprilTagsModule.cpp
```diff
@@ -0,0 +1,170 @@
...
+				Eigen::Isometry3d temp_pose = frame_pose * marker_pose;
+				Eigen::Isometry3d skel_pose = temp_pose * skel_offset;
+
+				Eigen::Isometry3d link_offset = mReferenceLink->getWorldTransform();
+				skel_pose = link_offset * skel_pose;
+
+				//Check if skel in skel_list
+				//If there then update pose
+				//If not then append skeleton
+				skel_name.append(std::to_string(marker_transform.id));
```
mkoval: What is the logic behind appending the `id` here? Could `getTagNameOffset` do this internally?

aikido_perception/src/AprilTagsModule.cpp
```diff
@@ -0,0 +1,170 @@
...
+
+				//Get root joint of skeleton
+				dart::dynamics::Joint* jtptr = skel_to_update->getJoint(0);
+
+				//Downcast Joint to FreeJoint
+				dart::dynamics::FreeJoint* freejtptr = dynamic_cast<dart::dynamics::FreeJoint*>(jtptr);
+
+				//Check if successful down-cast
+				if(freejtptr == NULL){
+					dtwarn << "Could not cast the joint of the body to a Free Joint so ignoring marker "<<tag_name << std::endl;
```
mkoval: The convention in DART is to prefix log output with `[ClassName::functionName]`.

aikido_perception/src/YamlFileLoader.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+namespace aikido{
+namespace perception{
+
+//================================================================================================================================
+YamlFileLoader::YamlFileLoader(const dart::common::ResourceRetrieverPtr& resourceRetriever,
+                               dart::common::Uri configDataURI)
+{
+    
+    dart::common::ResourceRetrieverPtr _resourceRetrieverPtr = resourceRetriever;
```
mkoval: What's the point of this temporary variable?

aikido_perception/src/YamlFileLoader.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+    //Read JSON file into string
+    if(!_resourceRetrieverPtr){
+        dtwarn << "[YamlFileLoader] ResourceRetrieverPtr given is null! "<<std::endl;
```
mkoval: DART convention is `[YamlFileLoader::constructor]`.

aikido_perception/src/YamlFileLoader.cpp
```diff
@@ -0,0 +1,70 @@
...
+        throw std::runtime_error("Null pointer passed");
+    }
+    
+    const dart::common::ResourcePtr resource = resourceRetriever->retrieve(configDataURI);
+    if(!resource)
+    {
+        throw std::runtime_error("Failed opening URI.");
```
mkoval: Please add `configDataUri` to the error message.

aikido_perception/src/YamlFileLoader.cpp
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    //Put file in string
+    const size_t size = resource->getSize();
+    std::string content;
+    content.resize(size);
+    if(resource->read(&content.front(), size, 1) != 1)
+    {
+        throw std::runtime_error("Failed reading from URI.");
```
mkoval: Please add `configDataUri` to the error message.

aikido_perception/src/YamlFileLoader.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+//================================================================================================================================
+bool YamlFileLoader::getTagNameOffset(const std::string& _tagName, std::string& body_name, dart::common::Uri& body_resource, Eigen::Isometry3d& body_offset)
+{
+    //Get name of object and pose for a given tag ID
+    YAML::Node name_offset = mTagData[_tagName];
+    if(name_offset){
+        body_resource.fromString(name_offset["resource"].as<std::string>());
+        body_name = name_offset["name"].as<std::string>();
+        body_offset = name_offset["offset"].as<Eigen::Isometry3d>();
```
mkoval: I suggest handling the conversion errors here and printing an informative warning. `yaml-cpp` produces notoriously bad error messages when an `as<T>` conversion fails that provides *zero information* about where the error occurred in the file.

aikido_perception/src/YamlFileLoader.cpp
```diff
@@ -0,0 +1,70 @@
...
+        return true;
+    }
+    else{
+        dtwarn << "[YamlFileLoader] Could not find Tag ID " << _tagName <<"\n";
```
mkoval: Nit: `[YamlFileLoader]` -> `[YamlFileLoader::getTagNameOffset]`. It would also be nice to print out the URI here for debugging purposes, but it isn't strictly necessary.

aikido_perception/CMakeLists.txt
Shushman: Done

aikido_perception/CMakeLists.txt
Shushman: Done

aikido_perception/include/aikido/perception/AprilTagsModule.hpp
Shushman: I've added doxygen style comments to pretty much all functions and parameters, but I saw multiple options and I may have to edit the exact choice I made later to reflect the rest of `aikido`.

aikido_perception/include/aikido/perception/AprilTagsModule.hpp
Shushman: Can you verify that I'm doing the correct thing now?

aikido_perception/include/aikido/perception/AprilTagsModule.hpp
Shushman: I'm not sure what to do here then? I've tentatively made it `ros::Time(0.0)`. What did you have in mind instead? To not have a default argument for it?

aikido_perception/include/aikido/perception/AprilTagsModule.hpp
Shushman: Changed to putting directly in the class.

aikido_perception/include/aikido/perception/ConfigDataLoader.hpp
Shushman: Same as earlier comment on Doxygen

aikido_perception/include/aikido/perception/YamlFileLoader.hpp
Shushman: Added

aikido_perception/include/aikido/perception/yaml_conversion.hpp
Shushman: Oh overlooked this! Will remove it.

aikido_perception/src/AprilTagsModule.cpp
Shushman: I've done `std::move` for all arguments in this case. I assume the `dart::dynamics::Frame*` is also by value (since the value of a pointer is the address of what it points to)?

aikido_perception/src/AprilTagsModule.cpp
Shushman: Now initialized with the `NodeHandle`, in the class

aikido_perception/src/AprilTagsModule.cpp
Shushman: I was not sure what you meant by this. I assume you meant a more detailed error message than `ROS_ERROR(ex.what())` so I've done something concatenating `ex.what()` to a context message and then sending its `.c_str()` to `ROS_ERROR`. Please have a look to confirm if that's what you meant.

aikido_perception/src/AprilTagsModule.cpp
Shushman: Changed

aikido_perception/src/AprilTagsModule.cpp
Shushman: To do that I'd have to send the `id` parameter to the `YamlFileLoader::getTagNameAndOffset` function. I was avoiding that as it is obtained for free and does not need to be used anywhere in the `yaml` context, so I did not want to send a parameter just for the sake of appending. Do you recommend that?

aikido_perception/src/AprilTagsModule.cpp
Shushman: Done.

aikido_perception/src/YamlFileLoader.cpp
Shushman: Removed that. Please confirm the current logic is correct.

aikido_perception/src/YamlFileLoader.cpp
Shushman: Changed

aikido_perception/src/YamlFileLoader.cpp
Shushman: Done

aikido_perception/src/YamlFileLoader.cpp
Shushman: Done

aikido_perception/src/YamlFileLoader.cpp
Shushman: I put two examples of what I think you mean by this in `YamlFileLoader.cpp`. If you think I am on the right track, I'll put the third one.

aikido_perception/src/YamlFileLoader.cpp
Shushman: Made the more verbose tag. To print out the URI here I'd need an additional member variable to store it. Is it worth that?

aikido_perception/src/AprilTagsModule.cpp
Shushman: This is the one change I have doubts about. The `waitForMessage` is used in the current `prpy` implementation of the `AprilTagsModule`. I think we also had a discussion a while ago that we did not want this to continuously subscribe to the topic but rather get the next message when called upon to do so. I assumed that this formulation achieved that. I could be wrong though? We should discuss in person.

mkoval:
High level comments:

- add `class` level docstrings
- remove file-level docstrings unless they contain useful information that can't be put elsewhere
- correctly annotate `\param[in]` and `\param[out]` parameters
- switch to the DART-style docstrings (`///`)
- only use `virtual` inheritance when it is absolutely necessary (I don't think it is for any of these classes)
- remove code that only exists to support the old-style yaml-cpp API (< 0.5), unless you actually intend to support it

---

Reviewed 1 of 11 files at r1, 3 of 8 files at r2, 1 of 11 files at r4, 3 of 10 files at r6, 1 of 4 files at r7, 1 of 1 files at r9, 8 of 8 files at r10.
Review status: all files reviewed at latest revision, 31 unresolved discussions.

---

*[aikido_perception/include/aikido/perception/AprilTagsModule.hpp, line 30 \[r9\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KFkTScmoMEpXh5Pss1r-r9-30:-KGKiJeZp5rJW8WxsQQ3:713930192) ([raw file](https://github.com/personalrobotics/aikido/blob/61bebce0fa1bf216ef35cd3a7bd9b68e2fa2c264/aikido_perception/include/aikido/perception/AprilTagsModule.hpp#L30)):*
- We're still missing a docstring on the `class`.
- I suggest *not* virtually inheriting unless you have a good reason to do so.
- Use the `///` docstring file as DART.
- The parameter docstring annotations are wrong - they should be `\param[in] paramName` for input parameters or `\param[out] paramName` for output parameters.

---

*[aikido_perception/include/aikido/perception/AprilTagsModule.hpp, line 34 \[r9\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KFkTScmoMEpXh5Pss1r-r9-34:-KGKjaDOM3c0uwiwLTvY:-1668913289) ([raw file](https://github.com/personalrobotics/aikido/blob/61bebce0fa1bf216ef35cd3a7bd9b68e2fa2c264/aikido_perception/include/aikido/perception/AprilTagsModule.hpp#L34)):*
Remove the `const` in front of `dart::common::ResourceRetrieverPtr` - it's not meaningful to put that in a header.

---

*[aikido_perception/include/aikido/perception/AprilTagsModule.hpp, line 7 \[r10\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KGKlRsVP7lIDO816TVl:-KGKlRsVP7lIDO816TVm:-1952843668) ([raw file](https://github.com/personalrobotics/aikido/blob/ca73c13f5d670e2cebea08ab2fb4a44d391758a5/aikido_perception/include/aikido/perception/AprilTagsModule.hpp#L7)):*
No other files in Aikido have this header, so I suggest removing it. You can put this information in the `class` docstring if you'd like.

---

*[aikido_perception/include/aikido/perception/ConfigDataLoader.hpp, line 15 \[r4\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KFkTScnQ-vjTy7BcQHU-r4-15:-KGKlppsR2dxE0TitHJ6:-8652494) ([raw file](https://github.com/personalrobotics/aikido/blob/6e26a11e94f025b0c3a3ddf33b254cd3a451ccb6/aikido_perception/include/aikido/perception/ConfigDataLoader.hpp#L15)):*
Still need a `class`-level docstring.

---

*[aikido_perception/include/aikido/perception/ConfigDataLoader.hpp, line 6 \[r10\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KGKllWcwQUzUS9C0XwL:-KGKllWdVciGZnAHM9S_:-1084201356) ([raw file](https://github.com/personalrobotics/aikido/blob/ca73c13f5d670e2cebea08ab2fb4a44d391758a5/aikido_perception/include/aikido/perception/ConfigDataLoader.hpp#L6)):*
Remove this header and move the relevant content to the `class` docstring.

---

*[aikido_perception/include/aikido/perception/ConfigDataLoader.hpp, line 29 \[r10\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KGKlzc3HDoG5vy24rCS:-KGKlzc3HDoG5vy24rCT:-1809213504) ([raw file](https://github.com/personalrobotics/aikido/blob/ca73c13f5d670e2cebea08ab2fb4a44d391758a5/aikido_perception/include/aikido/perception/ConfigDataLoader.hpp#L29)):*
- Switch to `///`-style docstrings.
- Tag `body_name`, `body_resource`, and `body_offset` as `\param[out]`.
-  Add a `\return` comment.

---

*[aikido_perception/include/aikido/perception/eigen_yaml.hpp, line 46 \[r10\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KGKmc0ePhqfA5HyiDPq:-KGKmc0ePhqfA5HyiDPr:-325529286) ([raw file](https://github.com/personalrobotics/aikido/blob/ca73c13f5d670e2cebea08ab2fb4a44d391758a5/aikido_perception/include/aikido/perception/eigen_yaml.hpp#L46)):*
Remove this `#else` clause unless you intend to support the old (< 0.5) yaml-cpp < 0 API.

---

*[aikido_perception/include/aikido/perception/eigen_yaml.hpp, line 75 \[r10\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KGKmvIVWwu8XdXZCzwh:-KGKmvIW9m3BbgcjXm8y:-325529286) ([raw file](https://github.com/personalrobotics/aikido/blob/ca73c13f5d670e2cebea08ab2fb4a44d391758a5/aikido_perception/include/aikido/perception/eigen_yaml.hpp#L75)):*
Remove this `#else` clause unless you intend to support the old (< 0.5) yaml-cpp < 0 API.

---

*[aikido_perception/include/aikido/perception/eigen_yaml.hpp, line 98 \[r10\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KGKmw0juur6sWLt1lpw:-KGKmw0juur6sWLt1lpx:-325529286) ([raw file](https://github.com/personalrobotics/aikido/blob/ca73c13f5d670e2cebea08ab2fb4a44d391758a5/aikido_perception/include/aikido/perception/eigen_yaml.hpp#L98)):*
Remove this `#else` clause unless you intend to support the old (< 0.5) yaml-cpp < 0 API.

---

*[aikido_perception/include/aikido/perception/PerceptionModule.hpp, line 6 \[r10\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KGKtVSB1S08NRBFmFg5:-KGKtVSB1S08NRBFmFg6:-1421251434) ([raw file](https://github.com/personalrobotics/aikido/blob/ca73c13f5d670e2cebea08ab2fb4a44d391758a5/aikido_perception/include/aikido/perception/PerceptionModule.hpp#L6)):*
Remove this header.

---

*[aikido_perception/include/aikido/perception/shape_conversions.hpp, line 7 \[r10\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KGKtPpVkLRwpRPO8xHD:-KGKtPpVkLRwpRPO8xHE:-1421251434) ([raw file](https://github.com/personalrobotics/aikido/blob/ca73c13f5d670e2cebea08ab2fb4a44d391758a5/aikido_perception/include/aikido/perception/shape_conversions.hpp#L7)):*
Remove this header.

---

*[aikido_perception/include/aikido/perception/shape_conversions.hpp, line 23 \[r10\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KGKtLPwW4zNUu41GFOa:-KGKtLPwW4zNUu41GFOb:1651183200) ([raw file](https://github.com/personalrobotics/aikido/blob/ca73c13f5d670e2cebea08ab2fb4a44d391758a5/aikido_perception/include/aikido/perception/shape_conversions.hpp#L23)):*
Add docstrings.

---

*[aikido_perception/include/aikido/perception/YamlFileLoader.hpp, line 25 \[r9\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KFkTScpmNxtpl6s9M8Y-r9-25:-KGKnBLxcg0iPsKIONSf:-1884841330) ([raw file](https://github.com/personalrobotics/aikido/blob/61bebce0fa1bf216ef35cd3a7bd9b68e2fa2c264/aikido_perception/include/aikido/perception/YamlFileLoader.hpp#L25)):*
- Still missing the `class`-level docstring.
- Remove `virtual` inheritance unless you actually need it.

---

*[aikido_perception/include/aikido/perception/YamlFileLoader.hpp, line 7 \[r10\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KGKn8TaMBR9d5SaIpWx:-KGKn8TaMBR9d5SaIpWy:-1838586610) ([raw file](https://github.com/personalrobotics/aikido/blob/ca73c13f5d670e2cebea08ab2fb4a44d391758a5/aikido_perception/include/aikido/perception/YamlFileLoader.hpp#L7)):*
Remove this header and write a `class` docstring.

---

*[aikido_perception/include/aikido/perception/YamlFileLoader.hpp, line 30 \[r10\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KGKnJ5yOzMHpUuNosjL:-KGKnJ5yOzMHpUuNosjM:-484069531) ([raw file](https://github.com/personalrobotics/aikido/blob/ca73c13f5d670e2cebea08ab2fb4a44d391758a5/aikido_perception/include/aikido/perception/YamlFileLoader.hpp#L30)):*
Typo.

---

*[aikido_perception/include/aikido/perception/YamlFileLoader.hpp, line 40 \[r10\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KGKnKga6-kMNTWs9zUp:-KGKnKga6-kMNTWs9zUq:-2118281774) ([raw file](https://github.com/personalrobotics/aikido/blob/ca73c13f5d670e2cebea08ab2fb4a44d391758a5/aikido_perception/include/aikido/perception/YamlFileLoader.hpp#L40)):*
Omit the docstring and write `// Documentation inherited` (note: `//`, not `///`) if you want to use the overriden method's docstring.

---

*[aikido_perception/src/AprilTagsModule.cpp, line 60 \[r4\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KFkTSclEPVU3tAaun-t-r4-60:-KGKoTOAddeg-bfOjbP4:804546575) ([raw file](https://github.com/personalrobotics/aikido/blob/6e26a11e94f025b0c3a3ddf33b254cd3a451ccb6/aikido_perception/src/AprilTagsModule.cpp#L60)):*
- I suggest *not* printing a warning here. This is expected to occur in normal operation. Only print a warning if we timeout without getting any valid `Marker` messages.
- You should only perform this check if `timestamp.isValid()`, i.e. not zero.

---

*[aikido_perception/src/AprilTagsModule.cpp, line 141 \[r4\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KFkTSclEPVU3tAaun-t-r4-141:-KGKqGKAEygTAs1GdTUK:21844905) ([raw file](https://github.com/personalrobotics/aikido/blob/6e26a11e94f025b0c3a3ddf33b254cd3a451ccb6/aikido_perception/src/AprilTagsModule.cpp#L141)):*
Prefix this with the class and function name. It is more accurate to say something like "Failed to load skeleton from URI", including `skel_resource`.

---

*[aikido_perception/src/AprilTagsModule.cpp, line 49 \[r9\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KFkTSclEPVU3tAaun-t-r9-49:-KGKoF4T4_9GM-32uOX3:-987711676) ([raw file](https://github.com/personalrobotics/aikido/blob/61bebce0fa1bf216ef35cd3a7bd9b68e2fa2c264/aikido_perception/src/AprilTagsModule.cpp#L49)):*
This is a relatively minor quibble, so I am fine with punting this to a future change. Please create an issue on Aikido to track this and link to it here.

---

*[aikido_perception/src/AprilTagsModule.cpp, line 88 \[r9\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KFkTSclEPVU3tAaun-t-r9-88:-KGKpk-W1R6QNL1572RZ:344772033) ([raw file](https://github.com/personalrobotics/aikido/blob/61bebce0fa1bf216ef35cd3a7bd9b68e2fa2c264/aikido_perception/src/AprilTagsModule.cpp#L88)):*
I didn't see this before, but it's strange to use `dtwarn` above and `ROS_ERROR` here. I suggest using `dtwarn` for consistency.

---

*[aikido_perception/src/AprilTagsModule.cpp, line 7 \[r10\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KGKnbdxjEXCmjx03yP-:-KGKnbdxjEXCmjx03yP0:-1421251434) ([raw file](https://github.com/personalrobotics/aikido/blob/ca73c13f5d670e2cebea08ab2fb4a44d391758a5/aikido_perception/src/AprilTagsModule.cpp#L7)):*
Remove this header.

---

*[aikido_perception/src/AprilTagsModule.cpp, line 45 \[r10\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KGKo0GDhtUCijtEkYnM:-KGKo0GDhtUCijtEkYnN:908170930) ([raw file](https://github.com/personalrobotics/aikido/blob/ca73c13f5d670e2cebea08ab2fb4a44d391758a5/aikido_perception/src/AprilTagsModule.cpp#L45)):*
I suggest making this a member variable. This will dramatically speed up `detectObjects` calls for the reason you saw when testing - the `TransformListener` will already have old transformations in its queue.

---

*[aikido_perception/src/AprilTagsModule.cpp, line 142 \[r10\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KGKrk0PhAqrUCBsBijA:-KGKrk0PhAqrUCBsBijB:-310667512) ([raw file](https://github.com/personalrobotics/aikido/blob/ca73c13f5d670e2cebea08ab2fb4a44d391758a5/aikido_perception/src/AprilTagsModule.cpp#L142)):*
We still need a check for `getNumDofs() > 0` before we do this access.

---

*[aikido_perception/src/AprilTagsModule.cpp, line 148 \[r10\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KGKruPq7n1os3gxVtH-:-KGKruPq7n1os3gxVtH0:-1940371456) ([raw file](https://github.com/personalrobotics/aikido/blob/ca73c13f5d670e2cebea08ab2fb4a44d391758a5/aikido_perception/src/AprilTagsModule.cpp#L148)):*
`nullptr`, not `NULL`.

---

*[aikido_perception/src/shape_conversions.cpp, line 7 \[r10\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KGKs3r7orzwd0OVsGmr:-KGKs3r7orzwd0OVsGms:-1421251434) ([raw file](https://github.com/personalrobotics/aikido/blob/ca73c13f5d670e2cebea08ab2fb4a44d391758a5/aikido_perception/src/shape_conversions.cpp#L7)):*
Remove this header.

---

*[aikido_perception/src/YamlFileLoader.cpp, line 26 \[r9\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KFkTScpmNxtpl6s9M8Z-r9-26:-KGKsAPHIc-z3bFn5ebb:1597085445) ([raw file](https://github.com/personalrobotics/aikido/blob/61bebce0fa1bf216ef35cd3a7bd9b68e2fa2c264/aikido_perception/src/YamlFileLoader.cpp#L26)):*
You don't need the `[Class::Function]` prefix in exceptions - we get a stack trace.

---

*[aikido_perception/src/YamlFileLoader.cpp, line 58 \[r9\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KFkTScpmNxtpl6s9M8Z-r9-58:-KGKsOdVh7ua_Mgd_f_3:-1054671365) ([raw file](https://github.com/personalrobotics/aikido/blob/61bebce0fa1bf216ef35cd3a7bd9b68e2fa2c264/aikido_perception/src/YamlFileLoader.cpp#L58)):*
This is a good start. If you're feeling ambitious, you can call `Mark()` and add the `line` and `column` of the error message. I suggest wrapping this is a helper function, e.g. `getAsOrThrow<T>` unless you're feeling masochistic.

---

*[aikido_perception/src/YamlFileLoader.cpp, line 7 \[r10\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KGKun11GuIKhne-nHXs:-KGKun12ppX5s1zW9dLs:-403101607) ([raw file](https://github.com/personalrobotics/aikido/blob/ca73c13f5d670e2cebea08ab2fb4a44d391758a5/aikido_perception/src/YamlFileLoader.cpp#L7)):*
Remove this.

---

*[aikido_perception/src/YamlFileLoader.cpp, line 13 \[r10\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KGKtAtmhIzCCcyZscc2:-KGKtAtnUG_VaHQo-NnW:-1409401240) ([raw file](https://github.com/personalrobotics/aikido/blob/ca73c13f5d670e2cebea08ab2fb4a44d391758a5/aikido_perception/src/YamlFileLoader.cpp#L13)):*
Use `<this/type/of/brackets.hpp>` for system includes.

---

*[aikido_perception/src/YamlFileLoader.cpp, line 75 \[r10\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KGKt4FE5PCkL8SdYJ8q:-KGKt4FFBte6rzaP5lzB:2114426259) ([raw file](https://github.com/personalrobotics/aikido/blob/ca73c13f5d670e2cebea08ab2fb4a44d391758a5/aikido_perception/src/YamlFileLoader.cpp#L75)):*
I don't think you should print a warning here. The user can check the return value and print a warning if they care to do so.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-:-KGKth9qmYRXNyN4kBSt:-1149254423)*

mkoval:
We're getting there! :sweat: 

A few *very minor* high-level comments:

- Improve the naming of the `ConfigLoader` classes.
- Write more informative docstrings for the classes and functions. Try to think about what the function does and any information the user needs to know to call it, but is not captured in the functions' signature. Default behavior, error conditions, and edge cases usually fall into this category.

---

Reviewed 2 of 8 files at r2, 1 of 1 files at r11, 8 of 9 files at r12.
Review status: all files reviewed at latest revision, 15 unresolved discussions.

---

*[aikido_perception/include/aikido/perception/AprilTagsModule.hpp, line 21 \[r13\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KGQIjN5Zi43Gys_f7_w:-KGQIjN5Zi43Gys_f7_x:-1277200103) ([raw file](https://github.com/personalrobotics/aikido/blob/734f34f78c158b904f4b6ee3910257c8012b3038/aikido_perception/include/aikido/perception/AprilTagsModule.hpp#L21)):*
The comment should describe what the class is used for. E.g. something like this:

> Perception module that detects objects using an AprilTag detector running in a separate process. This class subscribes to a `visualization_msg::Marker` ROS topic published by the AprilTag node, uses a \c AprilTagDatabase to resolve each marker to a \c dart::common::Uri, and updates the environment accordingly.

---

*[aikido_perception/include/aikido/perception/AprilTagsModule.hpp, line 36 \[r13\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KGQJR7jtjbKqLDVDl_R:-KGQJR7jtjbKqLDVDl_S:-445969105) ([raw file](https://github.com/personalrobotics/aikido/blob/734f34f78c158b904f4b6ee3910257c8012b3038/aikido_perception/include/aikido/perception/AprilTagsModule.hpp#L36)):*
This is not a useful comment - I'd prefer to remove it.

---

*[aikido_perception/include/aikido/perception/ConfigDataLoader.hpp, line 11 \[r13\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KGQILUyyIpFam3ZB0r8:-KGQILUzhRkp5zcGGqy9:598401633) ([raw file](https://github.com/personalrobotics/aikido/blob/734f34f78c158b904f4b6ee3910257c8012b3038/aikido_perception/include/aikido/perception/ConfigDataLoader.hpp#L11)):*
This is not a very descriptive name. I believe this class is only useful for AprilTag markers, so it should be named accordingly (e.g. `AprilTagDatabase`). Sorry I didn't notice this before.

---

*[aikido_perception/include/aikido/perception/ConfigDataLoader.hpp, line 17 \[r13\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KGQHqIGz76XMwcnFAzL:-KGQHqIGz76XMwcnFAzM:-1593242180) ([raw file](https://github.com/personalrobotics/aikido/blob/734f34f78c158b904f4b6ee3910257c8012b3038/aikido_perception/include/aikido/perception/ConfigDataLoader.hpp#L17)):*
Nit: The docstring should say what the function *does*, not what it *is*. For example, something like this would be ideal:

> Gets the name, resource URI, and relative offset of an Apriltag attached to a body."

A good litmus test for this is to look for verbs (e.g. "gets", "sets", "computes") and be wary of docstrings that have none.

---

*[aikido_perception/include/aikido/perception/ConfigDataLoader.hpp, line 22 \[r13\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KGQHcR8TRSETTM0B_Dg:-KGQHcR8TRSETTM0B_Dh:-2026055180) ([raw file](https://github.com/personalrobotics/aikido/blob/734f34f78c158b904f4b6ee3910257c8012b3038/aikido_perception/include/aikido/perception/ConfigDataLoader.hpp#L22)):*
- `\return` comment is inaccurate. This actually returns a `bool` that indicates success or failure.

---

*[aikido_perception/include/aikido/perception/PerceptionModule.hpp, line 17 \[r13\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KGQJomSXsh3KZu47quy:-KGQJomTkDt7sSdkUWoB:1835255959) ([raw file](https://github.com/personalrobotics/aikido/blob/734f34f78c158b904f4b6ee3910257c8012b3038/aikido_perception/include/aikido/perception/PerceptionModule.hpp#L17)):*
This should be a `\param[in,out]` if I am not mistaken.

---

*[aikido_perception/include/aikido/perception/PerceptionModule.hpp, line 18 \[r13\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KGQK1lV_-04WII_StQP:-KGQK1lWQncz0rwjnVAd:2125427058) ([raw file](https://github.com/personalrobotics/aikido/blob/734f34f78c158b904f4b6ee3910257c8012b3038/aikido_perception/include/aikido/perception/PerceptionModule.hpp#L18)):*
If we're committing to ROS types, then `duration` should probably be a `ros::Duration`. The standard behavior is generally to default to `ros::Duration(0)` and treat this as "no timeout".

---

*[aikido_perception/include/aikido/perception/PerceptionModule.hpp, line 19 \[r13\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KGQJrknNHAPkdr_fwJA:-KGQJrknNHAPkdr_fwJB:-64211792) ([raw file](https://github.com/personalrobotics/aikido/blob/734f34f78c158b904f4b6ee3910257c8012b3038/aikido_perception/include/aikido/perception/PerceptionModule.hpp#L19)):*
Document the fact that an invalid (i.e. zero) timestamp greedily takes the first available message. I suggest making that the default behavior.

---

*[aikido_perception/include/aikido/perception/PerceptionModule.hpp, line 21 \[r13\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KGQLk93ff2gvQT2hJXH:-KGQLk93ff2gvQT2hJXI:465168383) ([raw file](https://github.com/personalrobotics/aikido/blob/734f34f78c158b904f4b6ee3910257c8012b3038/aikido_perception/include/aikido/perception/PerceptionModule.hpp#L21)):*
There is currently no way to tell whether this function timed out or succeeded. I suggest making the return value `bool` to indicate this.

---

*[aikido_perception/include/aikido/perception/shape_conversions.hpp, line 23 \[r10\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KGKtLPwW4zNUu41GFOa:-KGQKPJVRJ_iJRUWC9a4:-888958099) ([raw file](https://github.com/personalrobotics/aikido/blob/ca73c13f5d670e2cebea08ab2fb4a44d391758a5/aikido_perception/include/aikido/perception/shape_conversions.hpp#L23)):*
Still missing docstrings.

---

*[aikido_perception/include/aikido/perception/YamlFileLoader.hpp, line 16 \[r13\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KGQKUe9ttpyxqM68ZP2:-KGQKUe9ttpyxqM68ZP3:636097168) ([raw file](https://github.com/personalrobotics/aikido/blob/734f34f78c158b904f4b6ee3910257c8012b3038/aikido_perception/include/aikido/perception/YamlFileLoader.hpp#L16)):*
- This is a docstring for the `class`, not the *header*.
- `YamlFileLoader` is a very generic name. Rename this to something more descriptive, e.g. `YamlAprilTagsDatabase`.

---

*[aikido_perception/src/AprilTagsModule.cpp, line 49 \[r9\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KFkTSclEPVU3tAaun-t-r9-49:-KGQLB_-wTwYVlCMte6B:926671389) ([raw file](https://github.com/personalrobotics/aikido/blob/61bebce0fa1bf216ef35cd3a7bd9b68e2fa2c264/aikido_perception/src/AprilTagsModule.cpp#L49)):*
Please make an issue to track this change.

---

*[aikido_perception/src/AprilTagsModule.cpp, line 31 \[r13\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KGQKykLObj9RAVHWBrq:-KGQKykLObj9RAVHWBrr:-1871794436) ([raw file](https://github.com/personalrobotics/aikido/blob/734f34f78c158b904f4b6ee3910257c8012b3038/aikido_perception/src/AprilTagsModule.cpp#L31)):*
- Initialize this in the initialization list unless there is a compelling reason to do otherwise.
- Why manage the `TransformListener` by `unique_ptr`, instead of by value (i.e. putting it directly in the class)?

---

*[aikido_perception/src/shape_conversions.cpp, line 7 \[r10\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KGKs3r7orzwd0OVsGmr:-KGQLX6MMK0PY5cr0GCa:-535236477) ([raw file](https://github.com/personalrobotics/aikido/blob/ca73c13f5d670e2cebea08ab2fb4a44d391758a5/aikido_perception/src/shape_conversions.cpp#L7)):*
Still need to remove this header.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-:-KGQLqYrezJLeMfnbvZ_:1712526532)*

mkoval:
Reviewed 2 of 8 files at r2, 12 of 12 files at r14.
Review status: all files reviewed at latest revision, 12 unresolved discussions.

---

*[aikido_perception/include/aikido/perception/AprilTagsDatabase.hpp, line 12 \[r14\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KGrPmg3OJU64RPeRUMr:-KGrPmg3OJU64RPeRUMs:-2096920147) ([raw file](https://github.com/personalrobotics/aikido/blob/da74cb335e08aea5257d90f8cc305faa3754c217/aikido_perception/include/aikido/perception/AprilTagsDatabase.hpp#L12)):*
Typo: last sentence is just the word "This."

---

*[aikido_perception/include/aikido/perception/PerceptionModule.hpp, line 16 \[r14\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KGrQhvtK0bliHD-vzzM:-KGrQhvtK0bliHD-vzzN:1268711331) ([raw file](https://github.com/personalrobotics/aikido/blob/da74cb335e08aea5257d90f8cc305faa3754c217/aikido_perception/include/aikido/perception/PerceptionModule.hpp#L16)):*
This still says what the function *is*, not what it *does* - there is no verb here. Please change it to something like this:

> Detect objects and updates the DART environment by creating, removing, and updating the pose of Skeletons in \c skeleton_list.

---

*[aikido_perception/include/aikido/perception/PerceptionModule.hpp, line 21 \[r14\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KGrQZ15k0b9HWQvcT7C:-KGrQZ15k0b9HWQvcT7D:1175730564) ([raw file](https://github.com/personalrobotics/aikido/blob/da74cb335e08aea5257d90f8cc305faa3754c217/aikido_perception/include/aikido/perception/PerceptionModule.hpp#L21)):*
Missing `\return` documentation. You could move the second sentence of the `timeout` parameter to that line.

---

*[aikido_perception/include/aikido/perception/YamlAprilTagsDatabase.hpp, line 16 \[r14\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KGrROTGMMVcL0oaYbk3:-KGrROTH7S1STC7cUmVT:130410330) ([raw file](https://github.com/personalrobotics/aikido/blob/da74cb335e08aea5257d90f8cc305faa3754c217/aikido_perception/include/aikido/perception/YamlAprilTagsDatabase.hpp#L16)):*
Add documentation about the file format here. Ideally, also include a snippet of example YAML in a `\code` block.

---

*[aikido_perception/include/aikido/perception/YamlAprilTagsDatabase.hpp, line 22 \[r14\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KGrRa8U9vUHbiFCqsw_:-KGrRa8U9vUHbiFCqswa:-2132570230) ([raw file](https://github.com/personalrobotics/aikido/blob/da74cb335e08aea5257d90f8cc305faa3754c217/aikido_perception/include/aikido/perception/YamlAprilTagsDatabase.hpp#L22)):*
Again: please explain what the function *does*, not what it *is*. E.g.:

> Constructs a \c YamlApriltagDatabase that uses \c resourceRetriever to load configuration data from a YAML file at the URI \c configDataURI.

---

*[aikido_perception/src/AprilTagsModule.cpp, line 30 \[r14\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KGrRva7fhlo1upiqAVH:-KGrRva8EE-SP_OjaU_P:1515966780) ([raw file](https://github.com/personalrobotics/aikido/blob/da74cb335e08aea5257d90f8cc305faa3754c217/aikido_perception/src/AprilTagsModule.cpp#L30)):*
It's not safe to use `node` here, since you `std::move`ed from it a few lines above this. You should pass `mNode` (not `node`) by value. I suspect this only works because `NodeHandle` in ROS indigo does not define a move constructor.

---

*[aikido_perception/src/AprilTagsModule.cpp, line 47 \[r14\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KGrSI7UaQ2hY5KIjBsH:-KGrSI7VuXkMHqdGI1VQ:-1847984888) ([raw file](https://github.com/personalrobotics/aikido/blob/da74cb335e08aea5257d90f8cc305faa3754c217/aikido_perception/src/AprilTagsModule.cpp#L47)):*
I'm not sure if we should treat this as a warning and return `false`. It's reasonable for the detector to detect nothing and publish an empty message. Thoughts?

If we do decide to keep this, note that we're missing a newline after the warning.

---

*[aikido_perception/src/AprilTagsModule.cpp, line 84 \[r14\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KGrSfertTS4JpH1bnAm:-KGrSfertTS4JpH1bnAn:-333613315) ([raw file](https://github.com/personalrobotics/aikido/blob/da74cb335e08aea5257d90f8cc305faa3754c217/aikido_perception/src/AprilTagsModule.cpp#L84)):*
Missing space between `timestamp` and the exception error message. Also, I don't think the `std::string` conversion is necessary here.

---

*[aikido_perception/src/YamlAprilTagsDatabase.cpp, line 59 \[r14\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KGrTCA62L-PLcXWGXhb:-KGrTCA62L-PLcXWGXhc:-1499012329) ([raw file](https://github.com/personalrobotics/aikido/blob/da74cb335e08aea5257d90f8cc305faa3754c217/aikido_perception/src/YamlAprilTagsDatabase.cpp#L59)):*
Don't add the `[class::method]` prefix to exception error messages. This is only necessary in `dtwarn` and `dterr` log messages, where a stack trace is not available.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/personalrobotics/aikido/24)*

Shushman:
Review status: 8 of 14 files reviewed at latest revision, 12 unresolved discussions.

---

*[aikido_perception/include/aikido/perception/PerceptionModule.hpp, line 16 \[r14\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KGrQhvtK0bliHD-vzzM:-KGsOHTIqV7mK9x5CSga:-1273803951) ([raw file](https://github.com/personalrobotics/aikido/blob/da74cb335e08aea5257d90f8cc305faa3754c217/aikido_perception/include/aikido/perception/PerceptionModule.hpp#L16)):*
Hmm. I had put that in the docstring for the class instead of the  but I see what you mean

---

*[aikido_perception/include/aikido/perception/YamlAprilTagsDatabase.hpp, line 16 \[r14\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KGrROTGMMVcL0oaYbk3:-KGs_oKlmjZzaDWIvbHA:-1149781515) ([raw file](https://github.com/personalrobotics/aikido/blob/da74cb335e08aea5257d90f8cc305faa3754c217/aikido_perception/include/aikido/perception/YamlAprilTagsDatabase.hpp#L16)):*
Not sure exactly how the `\code` block works. Will speak with you

---

*[aikido_perception/src/AprilTagsModule.cpp, line 30 \[r14\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KGrRva7fhlo1upiqAVH:-KGsaemnnrPCeSAg19Do:-499146009) ([raw file](https://github.com/personalrobotics/aikido/blob/da74cb335e08aea5257d90f8cc305faa3754c217/aikido_perception/src/AprilTagsModule.cpp#L30)):*
Done.

---

*[aikido_perception/src/AprilTagsModule.cpp, line 47 \[r14\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KGrSI7UaQ2hY5KIjBsH:-KGsaQppnTGS2T0hrEss:1023322479) ([raw file](https://github.com/personalrobotics/aikido/blob/da74cb335e08aea5257d90f8cc305faa3754c217/aikido_perception/src/AprilTagsModule.cpp#L47)):*
Well in terms of the `detectObjects()`, a `false` result makes sense to me because it has not actually detected any objects. And in this case it will return `false` and not affect the `skeleton_list`, so I thought that made sense.

---

*[aikido_perception/src/YamlAprilTagsDatabase.cpp, line 58 \[r9\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KFkTScpmNxtpl6s9M8Z-r9-58:-KGscUNvlXZ4x1l9gMM2:-159003148) ([raw file](https://github.com/personalrobotics/aikido/blob/61bebce0fa1bf216ef35cd3a7bd9b68e2fa2c264/aikido_perception/src/YamlAprilTagsDatabase.cpp#L58)):*
I'll think a bit more about how to do this. For now, I added an error check for the third component

---

*[aikido_perception/src/YamlAprilTagsDatabase.cpp, line 59 \[r14\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KGrTCA62L-PLcXWGXhb:-KGsaraolBjS4xSUlTbT:-499146009) ([raw file](https://github.com/personalrobotics/aikido/blob/da74cb335e08aea5257d90f8cc305faa3754c217/aikido_perception/src/YamlAprilTagsDatabase.cpp#L59)):*
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/personalrobotics/aikido/24)*

mkoval:
Reviewed 2 of 8 files at r2, 3 of 12 files at r14, 6 of 6 files at r15.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[aikido_perception/include/aikido/perception/YamlAprilTagsDatabase.hpp, line 16 \[r14\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KGrROTGMMVcL0oaYbk3:-KGsisFoYoS0NpMRjWdW:463874308) ([raw file](https://github.com/personalrobotics/aikido/blob/da74cb335e08aea5257d90f8cc305faa3754c217/aikido_perception/include/aikido/perception/YamlAprilTagsDatabase.hpp#L16)):*
Doxygen supports putting code in a `\code` ... `\endcode` block. See [here](https://github.com/personalrobotics/aikido/blob/fe985d53ea05017fde992cfef21aa18f8b911352/aikido/include/aikido/statespace/GeodesicInterpolator.hpp#L28) for an example.

---

*[aikido_perception/src/AprilTagsModule.cpp, line 49 \[r9\]](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-KFkTSclEPVU3tAaun-t-r9-49:-KGsjrooLMnVLnzCEj_A:-282351336) ([raw file](https://github.com/personalrobotics/aikido/blob/61bebce0fa1bf216ef35cd3a7bd9b68e2fa2c264/aikido_perception/src/AprilTagsModule.cpp#L49)):*
@Shushman posted an issue here: https://github.com/personalrobotics/aikido/issues/93.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/personalrobotics/aikido/24)*

mkoval:
Reviewed 2 of 8 files at r2, 2 of 12 files at r14, 5 of 5 files at r17.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/personalrobotics/aikido/24)*

mkoval:
:lgtm:

---

Review status: 13 of 14 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/personalrobotics/aikido/24#-:-KH6Z90x0W6BhELA7ZZL:1417088721)*

mkoval:
Reviewed 1 of 12 files at r14, 1 of 1 files at r18.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/personalrobotics/aikido/24)*

mkoval:
Great work @Shushman. Thanks for putting up with all of the comments!

Shushman:
Haha no worries @mkoval. It was an important learning experience for me too. And I'm sure there are many more to come :+1:

PR #75, opened 2016-04-22
Update STYLE.md in aikido

psigen:
**[IN PROGRESS]**
This PR improves the style guide documentation via simple example code and notes.

---
This change is [Reviewable](https://reviewable.io/reviews/personalrobotics/aikido/75)

### DIFF
--- a/aikido/STYLE.md
+++ b/aikido/STYLE.md
@@ -1,21 +1,295 @@
 # AIKIDO Style Guide #
 
-The code in this library should generally follow the rules of the Google C++ and Python Style Guides:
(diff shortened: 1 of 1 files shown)


jslee02:
This might be not that important, but let's consider adding `cmake` style as well.

psigen:
Great idea! Added a CMake section in e5d742c74ab5857a2dfbdcf2a90865616092950b.

mkoval:
This is a great start. Thanks for putting this together @psigen. We should be diligent about updating this when we run into new and exciting edge-cases.

PR #76, opened 2016-04-22
Renamed StateSpace and JointStateSpace classes

mkoval:
This change is [Reviewable](https://reviewable.io/reviews/personalrobotics/aikido/76)


jslee02:
In case we find the messages from coveralls annoying, we can turn it off the option `LEAVE COMMENTS` at the setup section.

PR #90, opened 2016-04-26
Improvements needed for planning with OMPL for HERB

jeking04:
This PR contains a set of improvements/bug fixes that were necessary to use the aikido framework and OMPL planners with HERB:

* Aikido state now set to a nullptr when a valid sample fails to be generated.  This includes regular state and goal sampling
* GeometricStateSpace, StateSampler, StateValidityChecker and GoalRegion updated to properly handle these states.
* Additional logic to ```aikido::constraint::NonColliding``` to pass collision options that ignore collisions between physically adjacent links by default.  Also added a constructor that exposes the option of passing custom ```CollisionOptions``` to the constraint.
* Added checks for infinite velocity or acceleration limits to ```ParabolicTimer```
* Update OMPL planner functions to use the custom ```MotionValidator``` by default
* Update OMPL planner functions to return ```Interpolated``` type paths instead of generic ```Trajectory``` type
* Return ```nullptr``` instead of an empty path when a solution cannot be found

---
This change is [Reviewable](https://reviewable.io/reviews/personalrobotics/aikido/90)

### DIFF
--- a/aikido/include/aikido/constraint/CyclicSampleable.hpp
+++ b/aikido/include/aikido/constraint/CyclicSampleable.hpp
@@ -15,6 +15,7 @@ namespace constraint {
 class CyclicSampleable : public Sampleable
 {
 public:
+
 	/// Constructor.
 	/// \param _sampleable Sampleable whose samples are to be iterated.
   explicit CyclicSampleable(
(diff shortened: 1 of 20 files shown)


mkoval:
:lgtm:

---

Reviewed 12 of 16 files at r1, 3 of 4 files at r2, 4 of 5 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/personalrobotics/aikido/90#-:-KGKyRvToAy3e5o1D8gO:1417088721)*

PR #94, opened 2016-04-29
Adding print functionality to aikido StateSpace

jeking04:
Suggestions for alternate formats for each StateSpace are welcome.

---
This change is [Reviewable](https://reviewable.io/reviews/personalrobotics/aikido/94)

### DIFF
--- a/aikido/include/aikido/statespace/CartesianProduct.hpp
+++ b/aikido/include/aikido/statespace/CartesianProduct.hpp
@@ -145,6 +145,12 @@ public:
   void logMap(const StateSpace::State *_in,
               Eigen::VectorXd &_tangent) const override;
 
+  /// Print the contents of each substate contained in the state
+  /// as a list with each substate enclosed in brackets and including its
+  /// index
+  /// Format: [0: ...] [1: ...] ... [n: ...]
+  void print(const StateSpace::State *_state, std::ostream &_os) const override;
+
 private:
   std::vector<StateSpacePtr> mSubspaces;
   std::vector<std::size_t> mOffsets;
--- a/aikido/include/aikido/statespace/Rn.hpp
+++ b/aikido/include/aikido/statespace/Rn.hpp
@@ -99,6 +99,10 @@ public:
   void logMap(const StateSpace::State *_in,
               Eigen::VectorXd &_tangent) const override;
 
+  /// Print the n-dimensional vector represented by the state
+  /// Format: [x_1, x_2, ..., x_n]
+  void print(const StateSpace::State *_state, std::ostream &_os) const override;
+
 private:
   /// Gets the mutable value stored in a Rn::State. This is
   /// used internally to implement the public \c getValue member functions.
--- a/aikido/include/aikido/statespace/SE2.hpp
+++ b/aikido/include/aikido/statespace/SE2.hpp
@@ -124,6 +124,9 @@ public:
   /// \param[out] _tangent corresponding element of the tangent space
   void logMap(const StateSpace::State *_state,
               Eigen::VectorXd &_tangent) const override;
+
+  /// Print the state. Format: [x, y, theta]
+  void print(const StateSpace::State *_state, std::ostream &_os) const override;
 };
 
 }  // namespace statespace
--- a/aikido/include/aikido/statespace/SE3.hpp
+++ b/aikido/include/aikido/statespace/SE3.hpp
@@ -124,6 +124,11 @@ public:
   /// \param[out] _tangent corresponding element of the tangent space
   void logMap(const StateSpace::State *_in,
               Eigen::VectorXd &_tangent) const override;
+
+  /// Print the quaternion followed by the translation
+  /// Format: [q.w, q.x, q.y, q.z, x, y, z] where is the quaternion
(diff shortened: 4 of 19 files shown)


psigen:
Is there some reason we implement these `print()` methods but don't implement `<<` stream operators?

mkoval:
It's not possible to make `operator <<` virtual. We could *also* implement a stream operator on the base class that calls this `virtual` function, but that does not obviate the need for this pull request.

psigen:
I guess my question is: why do we need to define a `virtual print()` function at all, if we commit to implementing the `operator <<` in the base class?

mkoval:
`print` is pure `virtual` on the base `StateSpace` class. Each concrete subclass of `StateSpace` has its own implementation of `print`. It's not possible to implement a generic `print` function on `StateSpace` because the the content of a `State` is only known by the derived type.

psigen:
Ok, maybe I need to clarify my question further here, since I'm not getting a clear answer...

Why do we need to have a `print()` function?
Why don't we _not_ have a `print()` function at all?

Then, anyplace in the code where you would use `print()`, you should use `operator <<` instead.

jeking04:
Does this help?

We would like to be able to use the operator << on an object of the base
StateSpace::State type and have it actually output using the stream
operator of its true derived type, I.e. SE2::State.  We can't make the
operator virtual on StateSpace. The best we can do is make the
StateSpace operator << call some other function that is virtual, i.e. print.

I agree that I should add a stream override on StateSpace, and possibly all
the derived statespace classes. But I think these should all just call the
classes print function. Perhaps the right thing to do is make print
protected instead of public.

On Friday, April 29, 2016, Pras Velagapudi <notifications@github.com> wrote:

> Ok, maybe I need to clarify my question further here, since I'm not
> getting a clear answer...
>
> Why do we need to have a print() function?
> Why don't we *not* have a print() function at all?
>
> Then, anyplace in the code where you would use print(), you should use operator
> << instead.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/personalrobotics/aikido/pull/94#issuecomment-215890340>
>


-- 
Jen King

mkoval:
@jeking04's explanation is correct. To reiterate, the motivation of this function is to print an opaque `StateSpace::State` type for debugging purposes. We typically *do not know* what concrete type of `StateSpace` we are working with, so a non-`virtual` function does not help.

In any case, I don't think it is possible to define `operator <<` on `State`. We can't print dynamically-sized states, like `CartesianProduct::State` and `Rn::State`, without also having a pointer to the `StateSpace` that they reside in. The best we can do is define a `virtual` function on `StateSpace` that handles this and provide `operator <<` on `StateHandle`.

psigen:
Thanks @jeking04, that makes perfect sense!  I get now that we fundamentally need these functions internally in any case.  Also the print format looks good, no issues there.

@mkoval: I think we're on the same page now. My concern isn't really the internal implementation, but simplifying the public API.  After all, it's pretty common to have the internal `print(ostream &)` method which is called by the `operator <<`, as @jeking04 mentioned.

The bigger issue, as you point out, is that we actually want to print the `StatePtr`, which is effectively raw memory and has no reference to its parent StateSpace.  So even the above indirection won't be sufficient, as the `operator << ` cannot meaningfully exist on the `StatePtr`.

So we should probably offer `operator <<` on StateHandle/ScopedState that calls the `print(StatePtr, ostream&) on the parent StateSpace.  Another option, which might just be the same thing as using a StateHandle API-wise, is to offer a `print()` overload that constructs and returns a temporary class:

```c++
class StatePrinter
{
  friend std::ostream& operator<< (std::ostream&, const StatePrinter&);
  
  StatePtr mState;
  StateSpacePtr mStateSpace;
};

class StateSpace
{
  [...]
  StatePrinter print(StatePtr);
};

int main()
{
  StateSpace statespace;
  StatePtr ptr = someMagicJunkThatMakesAState(); // Yeah, I still don't know what syntax goes here.

  std::cout << "My state was: " << statespace.print(ptr) <<  " and it's a pretty good state!" << std::cout;
}
```

There might be some usage of `StateHandle` that does much of this already, but this seems like a nice overload of `print()` to make available.

mkoval:
Reviewed 19 of 19 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/personalrobotics/aikido/94)*

mkoval:
I created #97 to track the `operator <<` overload.

jslee02:
*[aikido/include/aikido/statespace/CartesianProduct.hpp, line 152 \[r1\]](https://reviewable.io:443/reviews/personalrobotics/aikido/94#-KGrUfhHdnyDnIgkKVUs:-KGrUfhHdnyDnIgkKVUt:709094256) ([raw file](https://github.com/personalrobotics/aikido/blob/db7da4295b5165b724fc237c33fa28431c39997c/aikido/include/aikido/statespace/CartesianProduct.hpp#L152)):*
This PR is already merged, but how about passing `std::cout` as the default parameter for `_os`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/personalrobotics/aikido/94)*

PR #101, opened 2016-05-09
Feature/hand executor

gilwoolee:
This PR contains

- HandSimTrajectoryExecutor
- FingerSimStepExecutor.

Haven't tested collision part of the code yet.

---
This change is [Reviewable](https://reviewable.io/reviews/personalrobotics/aikido/101)


mkoval:
This is a good start

<details><summary><i>Previously, coveralls wrote…</i></summary>
> 
> [Coverage Status](https://coveralls.io/builds/6107327)
> 
> Coverage increased (+0.08%) to 90.03% when pulling **5f32229563668e7b6655079a73136448ade5d930 on feature/hand_executor** into **293de4551dd720a2736052d844fa8fa988b65166 on master**.
> </details>


---

Reviewed 8 of 8 files at r1.
Review status: all files reviewed at latest revision, 20 unresolved discussions, some commit checks failed.

---

*[aikido/include/aikido/control/FingerSimulationStepExecutor.hpp, line 11 \[r1\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KHSH1cZJmKjxLl71GDp:-KHSH1cZJmKjxLl71GDq:-1024159588) ([raw file](https://github.com/personalrobotics/aikido/blob/5f32229563668e7b6655079a73136448ade5d930/aikido/include/aikido/control/FingerSimulationStepExecutor.hpp#L11)):*
> ```cpp
> /// Assumes that fingers are underactuated: primal joint is actuated,
> /// distal joint moves with certain mimic ratio (set to be Barret hand ratio).
> class FingerSimulationStepExecutor
> ```

This name should reflect that this class:

- is specific to a finger on the BarrettHand
- simulates execution of a position command, not a trajectory

Perhaps something like `BarrettFingerPositionCommandExecutor`?

---

*[aikido/include/aikido/control/FingerSimulationStepExecutor.hpp, line 24 \[r1\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KHSGvMrLzKcOZeboBWV:-KHSGvMrLzKcOZeboBWW:764994276) ([raw file](https://github.com/personalrobotics/aikido/blob/5f32229563668e7b6655079a73136448ade5d930/aikido/include/aikido/control/FingerSimulationStepExecutor.hpp#L24)):*
> ```cpp
>   void execute(double _stepAngle, bool _distalOnly);
> 
>   static double getMimicRatio();
> ```

Missing docstring.

---

*[aikido/include/aikido/control/HandSimulationTrajectoryExecutor.hpp, line 18 \[r1\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KHSHmhslh1c-SejCtM4:-KHSHmhslh1c-SejCtM5:1762887540) ([raw file](https://github.com/personalrobotics/aikido/blob/5f32229563668e7b6655079a73136448ade5d930/aikido/include/aikido/control/HandSimulationTrajectoryExecutor.hpp#L18)):*
> ```cpp
> /// and distal joint moves with certain mimic ratio (set to be Barret hand) 
> /// until collision. 
> class HandSimulationTrajectoryExecutor
> ```

This name should reflect that this class:

- is specific to the to the BarrettHand
- simulates execution of a position command, not a trajectory

---

*[aikido/include/aikido/control/HandSimulationTrajectoryExecutor.hpp, line 29 \[r1\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KHSIBfvF49Y2pm_tBc2:-KHSIBfvF49Y2pm_tBc3:462562060) ([raw file](https://github.com/personalrobotics/aikido/blob/5f32229563668e7b6655079a73136448ade5d930/aikido/include/aikido/control/HandSimulationTrajectoryExecutor.hpp#L29)):*
> ```cpp
>   /// \param _collisionDetector
>   /// \param _collideWith
>   /// \param _collisionOptions 
> ```

Missing docstrings for these arguments. Also explain the default `_collisionOptions`.

---

*[aikido/include/aikido/control/HandSimulationTrajectoryExecutor.hpp, line 34 \[r1\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KHSMh8EOhjjZYBVB5z0:-KHSMh8EOhjjZYBVB5z1:-838911484) ([raw file](https://github.com/personalrobotics/aikido/blob/5f32229563668e7b6655079a73136448ade5d930/aikido/include/aikido/control/HandSimulationTrajectoryExecutor.hpp#L34)):*
> ```cpp
>     std::chrono::milliseconds _cyclePeriod,
>     ::dart::collision::CollisionDetectorPtr _collisionDetector,
>     ::dart::collision::CollisionGroupPtr _collideWith,
> ```

I am concerned about passing `_collideWith` to the constructor because the contents of the environment may change at runtime. E.g. if we detect an additional Fuze bottle, then we would need to re-create the `HandSimulationTrajectoryExecutor` before we can detect collision with it.

I suppose this is fine for now, but we should create an issue to track this. Our implementation of this may depend on the `World`/`Environment` API.

---

*[aikido/include/aikido/control/HandSimulationTrajectoryExecutor.hpp, line 46 \[r1\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KHSIMV7nEyqmNVIrFJ_:-KHSIMV7nEyqmNVIrFJa:-1586708403) ([raw file](https://github.com/personalrobotics/aikido/blob/5f32229563668e7b6655079a73136448ade5d930/aikido/include/aikido/control/HandSimulationTrajectoryExecutor.hpp#L46)):*
> ```cpp
>   ///        The trajectory may terminate early upon collision.
>   void execute(std::vector<double> _set_points, 
>     double _spread, std::chrono::milliseconds _duration);
> ```

- Typo: `_duratoin`
- Argument naming convention:`_setPoints`, not `_set_points`.
- Return a `std::future<void>` that allows you to block for the command to complete.
- Suggest merging `_setPoints` and `_spread` into one `Eigen::VectorXd` duration - I see no reason to split these.

---

*[aikido/src/control/FingerSimulationStepExecutor.cpp, line 33 \[r1\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KHSJ3dcMrmxigfB1Z_i:-KHSJ3ddBCBnS6NlHY4z:1339282400) ([raw file](https://github.com/personalrobotics/aikido/blob/5f32229563668e7b6655079a73136448ade5d930/aikido/src/control/FingerSimulationStepExecutor.cpp#L33)):*
> ```cpp
> //=============================================================================
> void FingerSimulationStepExecutor::execute(double _stepAngle, bool _distalOnly)
> {
> ```

I expected the collision detection logic to live in this function. What is the motivation behind doing all of this in `HandSimulationTrajectoryExecutor`?

---

*[aikido/src/control/FingerSimulationStepExecutor.cpp, line 36 \[r1\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KHSJY8iA0JLh2YbRmfn:-KHSJY8jFiJTU9glcKk7:932598038) ([raw file](https://github.com/personalrobotics/aikido/blob/5f32229563668e7b6655079a73136448ade5d930/aikido/src/control/FingerSimulationStepExecutor.cpp#L36)):*
> ```cpp
>   if (_stepAngle == 0)
>     return;
> ```

Also throw an exception if `isAssembled()` is `false`. This can occur if the topology of the `Skeleton` changes.

---

*[aikido/src/control/HandSimulationTrajectoryExecutor.cpp, line 18 \[r1\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KHSJnA1I1ymt2O4UFBB:-KHSJnA1I1ymt2O4UFBC:1203187480) ([raw file](https://github.com/personalrobotics/aikido/blob/5f32229563668e7b6655079a73136448ade5d930/aikido/src/control/HandSimulationTrajectoryExecutor.cpp#L18)):*
> ```cpp
> , mCollisionDetector(std::move(_collisionDetector))
> , mCollideWith(std::move(_collideWith))
> , mCollisionOptions(_collisionOptions)
> ```

Should `std::move` this as well.

---

*[aikido/src/control/HandSimulationTrajectoryExecutor.cpp, line 30 \[r1\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KHSJrUilHBDGiUNEMjP:-KHSJrUilHBDGiUNEMjQ:120162581) ([raw file](https://github.com/personalrobotics/aikido/blob/5f32229563668e7b6655079a73136448ade5d930/aikido/src/control/HandSimulationTrajectoryExecutor.cpp#L30)):*
> ```cpp
>     {
>       std::stringstream msg;
>       msg << "Finter " << i << " is null.";
> ```

Typo: `Finter`.

---

*[aikido/src/control/HandSimulationTrajectoryExecutor.cpp, line 45 \[r1\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KHSJx3b8Rm9GIvdf_G8:-KHSJx3b8Rm9GIvdf_G9:-1652926262) ([raw file](https://github.com/personalrobotics/aikido/blob/5f32229563668e7b6655079a73136448ade5d930/aikido/src/control/HandSimulationTrajectoryExecutor.cpp#L45)):*
> ```cpp
> 
>   mFingerPrimalDofExecutors.reserve(3);
>   mSubFingers.reserve(3);
> ```

Add a constant for the number of fingers to avoid hard-coding the magic numbers `2` and `3` in multiple places.

---

*[aikido/src/control/HandSimulationTrajectoryExecutor.cpp, line 48 \[r1\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KHSKF02MdYBUMNcprLo:-KHSKF02MdYBUMNcprLp:1278578787) ([raw file](https://github.com/personalrobotics/aikido/blob/5f32229563668e7b6655079a73136448ade5d930/aikido/src/control/HandSimulationTrajectoryExecutor.cpp#L48)):*
> ```cpp
> 
>   // Fingers
>   for (int i = 0; i < 2; ++i)
> ```

What's the purpose of re-creating the finger `Chain`s, instead of simply using the `Chain`s passed in by the user?

---

*[aikido/src/control/HandSimulationTrajectoryExecutor.cpp, line 54 \[r1\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KHSK4C-oNNIkFR4vwgH:-KHSK4C-oNNIkFR4vwgI:1927005441) ([raw file](https://github.com/personalrobotics/aikido/blob/5f32229563668e7b6655079a73136448ade5d930/aikido/src/control/HandSimulationTrajectoryExecutor.cpp#L54)):*
> ```cpp
>     auto subFinger = Chain::create(
>       mFingers[i]->getBodyNode(0), mFingers[i]->getBodyNode(2));
>     mSubFingers.emplace_back(subFinger);
> ```

Either `std::move` here or put the `Chain::create` inline.

---

*[aikido/src/control/HandSimulationTrajectoryExecutor.cpp, line 61 \[r1\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KHSKUU59SCzmq4hq1TK:-KHSKUU59SCzmq4hq1TL:647478230) ([raw file](https://github.com/personalrobotics/aikido/blob/5f32229563668e7b6655079a73136448ade5d930/aikido/src/control/HandSimulationTrajectoryExecutor.cpp#L61)):*
> ```cpp
>   for(int i = 0; i < 3; ++i){
>     mFingerPrimalDofExecutors.emplace_back(
>       FingerSimulationStepExecutor(mSubFingers[i]));
> ```

This should be:
```c++
mFingerPrimalDofExecutors.emplace_back(mSubFingers[i]);
```
There's no point in using `emplace_back` here if you explicitly call a constructor because it still invokes the copy constructor.

---

*[aikido/src/control/HandSimulationTrajectoryExecutor.cpp, line 65 \[r1\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KHSKPUedv6w0Zc-d7b5:-KHSKPUedv6w0Zc-d7b6:4738829) ([raw file](https://github.com/personalrobotics/aikido/blob/5f32229563668e7b6655079a73136448ade5d930/aikido/src/control/HandSimulationTrajectoryExecutor.cpp#L65)):*
> ```cpp
> 
>   if (mFingers[2]->getNumDofs() != 2)
>     throw std::invalid_argument("Third finger should have 2 dofs.");
> ```

Shouldn't this be checked by the `FingerSimulationStepExecutor` constructor?

---

*[aikido/src/control/HandSimulationTrajectoryExecutor.cpp, line 80 \[r1\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KHSKpSawyUB-nzgPvu8:-KHSKpSawyUB-nzgPvu9:395187364) ([raw file](https://github.com/personalrobotics/aikido/blob/5f32229563668e7b6655079a73136448ade5d930/aikido/src/control/HandSimulationTrajectoryExecutor.cpp#L80)):*
> ```cpp
>       mCollisionDetector->createCollisionGroup(
>         mSubFingers[i]->getBodyNode(1)));
>   }
> ```

I would prefer to encapsulate all of this logic in the finger simulator class.

---

*[aikido/src/control/HandSimulationTrajectoryExecutor.cpp, line 103 \[r1\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KHSKwjZ8cd2KDfVGzAm:-KHSKwjZ8cd2KDfVGzAn:-738950885) ([raw file](https://github.com/personalrobotics/aikido/blob/5f32229563668e7b6655079a73136448ade5d930/aikido/src/control/HandSimulationTrajectoryExecutor.cpp#L103)):*
> ```cpp
>     else if (_set_points[i] > mSubFingers[i]->getDof(0)->getPositionUpperLimit())
>       _set_points[i] = mSubFingers[i]->getDof(0)->getPositionUpperLimit();
>   }
> ```

Doesn't the same logic exist in the finger simulator?

---

*[aikido/src/control/HandSimulationTrajectoryExecutor.cpp, line 118 \[r1\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KHSLGWG_KL1XK0PneRR:-KHSLGWG_KL1XK0PneRS:-91639714) ([raw file](https://github.com/personalrobotics/aikido/blob/5f32229563668e7b6655079a73136448ade5d930/aikido/src/control/HandSimulationTrajectoryExecutor.cpp#L118)):*
> ```cpp
>   int numIntervals = std::ceil(_duration.count()/ mCyclePeriod.count());
>   double currentPrimalAngles[3];
>   double angleIncrements[3];
> ```

Nit: Use `Eigen::Vector3d` here instead of `double[3]` so we can use the math operators below.

---

*[aikido/src/control/HandSimulationTrajectoryExecutor.cpp, line 163 \[r1\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KHSLWfHmtAxMVTwVJLa:-KHSLWfHmtAxMVTwVJLb:-722918865) ([raw file](https://github.com/personalrobotics/aikido/blob/5f32229563668e7b6655079a73136448ade5d930/aikido/src/control/HandSimulationTrajectoryExecutor.cpp#L163)):*
> ```cpp
> 
>     std::this_thread::sleep_for(mCyclePeriod);
>   }
> ```

All of this should run in a separate thread and provide feedback through a `promise` and a `future`.

---

*[aikido/tests/control/test_HandSimulationTrajectoryExecutor.cpp, line 281 \[r1\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KHSMQuhxj3sWwyn1850:-KHSMQuiC1bnGWl_7pSn:-284335132) ([raw file](https://github.com/personalrobotics/aikido/blob/5f32229563668e7b6655079a73136448ade5d930/aikido/tests/control/test_HandSimulationTrajectoryExecutor.cpp#L281)):*
> ```cpp
>   EXPECT_DOUBLE_EQ(distal, startDistalDofValues[2] + mSetpoints[2]*mimicRatio);
> }
> ```

Also test the cases where collision occurs:

- with the primal link first
- with the distal link first

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-:-KHSMen1WbHhXj-5Zd3x:-1784040777)*

psigen:
This code seems very... specific to Barrett Hands and nothing else.  Is there some reason we would want it to be in `aikido` like this?

If anything, it would make sense for it to be semantically named like `MimicSimulationTrajectoryExecutor.cpp` since that seems to be what `Finger` actually means.  But I am also not sure why we have to have so many different Executor classes.  Are we composing them somehow?  Can we do this automatically from the kinematics/URDF instead of creating SimExecutors for every brand of robot arm?

lgw903:
Review status: 0 of 9 files reviewed at latest revision, 20 unresolved discussions.

---

*[aikido/include/aikido/control/HandSimulationTrajectoryExecutor.hpp, line 18 \[r1\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KHSHmhslh1c-SejCtM4:-KH_qmT1gn57wJAZcfO3:-499146009) ([raw file](https://github.com/personalrobotics/aikido/blob/5f32229563668e7b6655079a73136448ade5d930/aikido/include/aikido/control/HandSimulationTrajectoryExecutor.hpp#L18)):*
<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary>
> This name should reflect that this class:
> 
> - is specific to the to the BarrettHand
> - simulates execution of a position command, not a trajectory</details>
Done.

---

*[aikido/include/aikido/control/HandSimulationTrajectoryExecutor.hpp, line 29 \[r1\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KHSIBfvF49Y2pm_tBc2:-KHfoCtMNqivNDPM471B:-499146009) ([raw file](https://github.com/personalrobotics/aikido/blob/5f32229563668e7b6655079a73136448ade5d930/aikido/include/aikido/control/HandSimulationTrajectoryExecutor.hpp#L29)):*
<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary>
> Missing docstrings for these arguments. Also explain the default `_collisionOptions`.</details>
Done.

---

*[aikido/include/aikido/control/HandSimulationTrajectoryExecutor.hpp, line 34 \[r1\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KHSMh8EOhjjZYBVB5z0:-KH_re0ykgTRqFrXWIBJ:-499146009) ([raw file](https://github.com/personalrobotics/aikido/blob/5f32229563668e7b6655079a73136448ade5d930/aikido/include/aikido/control/HandSimulationTrajectoryExecutor.hpp#L34)):*
<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary>
> I am concerned about passing `_collideWith` to the constructor because the contents of the environment may change at runtime. E.g. if we detect an additional Fuze bottle, then we would need to re-create the `HandSimulationTrajectoryExecutor` before we can detect collision with it.
> 
> I suppose this is fine for now, but we should create an issue to track this. Our implementation of this may depend on the `World`/`Environment` API.</details>
Done.

---

*[aikido/include/aikido/control/HandSimulationTrajectoryExecutor.hpp, line 46 \[r1\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KHSIMV7nEyqmNVIrFJ_:-KHfHMyVAsnRyhEDu8oq:-499146009) ([raw file](https://github.com/personalrobotics/aikido/blob/5f32229563668e7b6655079a73136448ade5d930/aikido/include/aikido/control/HandSimulationTrajectoryExecutor.hpp#L46)):*
<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary>
> - Typo: `_duratoin`
> - Argument naming convention:`_setPoints`, not `_set_points`.
> - Return a `std::future<void>` that allows you to block for the command to complete.
> - Suggest merging `_setPoints` and `_spread` into one `Eigen::VectorXd` duration - I see no reason to split these.</details>
Done.

---

*[aikido/src/control/HandSimulationTrajectoryExecutor.cpp, line 18 \[r1\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KHSJnA1I1ymt2O4UFBB:-KHfHhLTY0TOrgZTSVmd:-499146009) ([raw file](https://github.com/personalrobotics/aikido/blob/5f32229563668e7b6655079a73136448ade5d930/aikido/src/control/HandSimulationTrajectoryExecutor.cpp#L18)):*
<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary>
> Should `std::move` this as well.</details>
Done.

---

*[aikido/src/control/HandSimulationTrajectoryExecutor.cpp, line 30 \[r1\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KHSJrUilHBDGiUNEMjP:-KHfHipXwoSR1HrNVPjB:-499146009) ([raw file](https://github.com/personalrobotics/aikido/blob/5f32229563668e7b6655079a73136448ade5d930/aikido/src/control/HandSimulationTrajectoryExecutor.cpp#L30)):*
<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary>
> Typo: `Finter`.</details>
Done.

---

*[aikido/src/control/HandSimulationTrajectoryExecutor.cpp, line 45 \[r1\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KHSJx3b8Rm9GIvdf_G8:-KHfMMssGs6LihxtEXp6:-499146009) ([raw file](https://github.com/personalrobotics/aikido/blob/5f32229563668e7b6655079a73136448ade5d930/aikido/src/control/HandSimulationTrajectoryExecutor.cpp#L45)):*
<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary>
> Add a constant for the number of fingers to avoid hard-coding the magic numbers `2` and `3` in multiple places.</details>
Done.

---

*[aikido/src/control/HandSimulationTrajectoryExecutor.cpp, line 48 \[r1\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KHSKF02MdYBUMNcprLo:-KHfMKhnP8-4UGLU86Nj:-499146009) ([raw file](https://github.com/personalrobotics/aikido/blob/5f32229563668e7b6655079a73136448ade5d930/aikido/src/control/HandSimulationTrajectoryExecutor.cpp#L48)):*
<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary>
> What's the purpose of re-creating the finger `Chain`s, instead of simply using the `Chain`s passed in by the user?</details>
Done.

---

*[aikido/src/control/HandSimulationTrajectoryExecutor.cpp, line 54 \[r1\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KHSK4C-oNNIkFR4vwgH:-KHfnMVm50vdskw0bjgP:-499146009) ([raw file](https://github.com/personalrobotics/aikido/blob/5f32229563668e7b6655079a73136448ade5d930/aikido/src/control/HandSimulationTrajectoryExecutor.cpp#L54)):*
<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary>
> Either `std::move` here or put the `Chain::create` inline.</details>
Done.

---

*[aikido/src/control/HandSimulationTrajectoryExecutor.cpp, line 61 \[r1\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KHSKUU59SCzmq4hq1TK:-KHfmsSQV6Y_DNVwizzX:-499146009) ([raw file](https://github.com/personalrobotics/aikido/blob/5f32229563668e7b6655079a73136448ade5d930/aikido/src/control/HandSimulationTrajectoryExecutor.cpp#L61)):*
<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary>
> This should be:
> ```c++
> mFingerPrimalDofExecutors.emplace_back(mSubFingers[i]);
> ```
> There's no point in using `emplace_back` here if you explicitly call a constructor because it still invokes the copy constructor.
> </details>
Done.

---

*[aikido/src/control/HandSimulationTrajectoryExecutor.cpp, line 65 \[r1\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KHSKPUedv6w0Zc-d7b5:-KHfpD0RZ4xZtU_0W7O1:-499146009) ([raw file](https://github.com/personalrobotics/aikido/blob/5f32229563668e7b6655079a73136448ade5d930/aikido/src/control/HandSimulationTrajectoryExecutor.cpp#L65)):*
<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary>
> Shouldn't this be checked by the `FingerSimulationStepExecutor` constructor?</details>
Done.

---

*[aikido/src/control/HandSimulationTrajectoryExecutor.cpp, line 80 \[r1\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KHSKpSawyUB-nzgPvu8:-KHfnTzDnK3uIHVnPqfs:-768617132) ([raw file](https://github.com/personalrobotics/aikido/blob/5f32229563668e7b6655079a73136448ade5d930/aikido/src/control/HandSimulationTrajectoryExecutor.cpp#L80)):*
<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary>
> I would prefer to encapsulate all of this logic in the finger simulator class.</details>
We discussed offline why this is handled here. Finger simulator does not keep its history, so even if collision occurred in the previous step, it will have to check again before execution. By doing it here, we have to do it only upto the point where collision happens. 

---

*[aikido/src/control/HandSimulationTrajectoryExecutor.cpp, line 103 \[r1\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KHSKwjZ8cd2KDfVGzAm:-KHfo2cYPd5lAWR8IloW:-499146009) ([raw file](https://github.com/personalrobotics/aikido/blob/5f32229563668e7b6655079a73136448ade5d930/aikido/src/control/HandSimulationTrajectoryExecutor.cpp#L103)):*
<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary>
> Doesn't the same logic exist in the finger simulator?</details>
Done.

---

*[aikido/src/control/HandSimulationTrajectoryExecutor.cpp, line 118 \[r1\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KHSLGWG_KL1XK0PneRR:-KHfo4ixie6_N6DIkAtS:-499146009) ([raw file](https://github.com/personalrobotics/aikido/blob/5f32229563668e7b6655079a73136448ade5d930/aikido/src/control/HandSimulationTrajectoryExecutor.cpp#L118)):*
<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary>
> Nit: Use `Eigen::Vector3d` here instead of `double[3]` so we can use the math operators below.</details>
Done.

---

*[aikido/src/control/HandSimulationTrajectoryExecutor.cpp, line 163 \[r1\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KHSLWfHmtAxMVTwVJLa:-KHfo5XoIaK6iazNaDwk:-499146009) ([raw file](https://github.com/personalrobotics/aikido/blob/5f32229563668e7b6655079a73136448ade5d930/aikido/src/control/HandSimulationTrajectoryExecutor.cpp#L163)):*
<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary>
> All of this should run in a separate thread and provide feedback through a `promise` and a `future`.</details>
Done.

---

*[aikido/include/aikido/control/FingerSimulationStepExecutor.hpp, line 11 \[r1\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KHSH1cZJmKjxLl71GDp:-KH_pMaVbgw2wlFo3lGd:-499146009) ([raw file](https://github.com/personalrobotics/aikido/blob/5f32229563668e7b6655079a73136448ade5d930/aikido/include/aikido/control/FingerSimulationStepExecutor.hpp#L11)):*
<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary>
> This name should reflect that this class:
> 
> - is specific to a finger on the BarrettHand
> - simulates execution of a position command, not a trajectory
> 
> Perhaps something like `BarrettFingerPositionCommandExecutor`?</details>
Done.

---

*[aikido/include/aikido/control/FingerSimulationStepExecutor.hpp, line 24 \[r1\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KHSGvMrLzKcOZeboBWV:-KH_pa-jfkLxNCBUI4BB:-499146009) ([raw file](https://github.com/personalrobotics/aikido/blob/5f32229563668e7b6655079a73136448ade5d930/aikido/include/aikido/control/FingerSimulationStepExecutor.hpp#L24)):*
<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary>
> Missing docstring.</details>
Done.

---

*[aikido/src/control/FingerSimulationStepExecutor.cpp, line 33 \[r1\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KHSJ3dcMrmxigfB1Z_i:-KHfHXpORKBhDi9sWFxY:-1998123377) ([raw file](https://github.com/personalrobotics/aikido/blob/5f32229563668e7b6655079a73136448ade5d930/aikido/src/control/FingerSimulationStepExecutor.cpp#L33)):*
<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary>
> I expected the collision detection logic to live in this function. What is the motivation behind doing all of this in `HandSimulationTrajectoryExecutor`?</details>
Discussed offline, also explained below.

---

*[aikido/src/control/FingerSimulationStepExecutor.cpp, line 36 \[r1\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KHSJY8iA0JLh2YbRmfn:-KHfp8ynBJSmT8xIuuYE:-499146009) ([raw file](https://github.com/personalrobotics/aikido/blob/5f32229563668e7b6655079a73136448ade5d930/aikido/src/control/FingerSimulationStepExecutor.cpp#L36)):*
<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary>
> Also throw an exception if `isAssembled()` is `false`. This can occur if the topology of the `Skeleton` changes.</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/personalrobotics/aikido/101)*

lgw903:
@moval Previous PR comments addressed. Major changes are: 
- Execution runs on a separate thread
- `collideWith` passed per execution

Please take a second look when you have a chance. :) 

<details><summary><i>Previously, psigen (Pras Velagapudi) wrote…</i></summary>
> This code seems very... specific to Barrett Hands and nothing else.  Is there some reason we would want it to be in `aikido` like this?
> 
> 
> 
> If anything, it seems like it would make more sense for it to be semantically named like `MimicSimulationTrajectoryExecutor.cpp`-- that seems to be what the term `Finger` actually means.
> 
> 
> 
> But I am also not sure why we have to have so many different `Executor` classes.  Are we composing them somehow?  Can we do this automatically from the kinematics/URDF instead of creating new `Executor`s for every brand of robot arm?</details>


---

Review status: 0 of 9 files reviewed at latest revision, 20 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-:-KHkw4pWa_Y41vC9em1M:-346828924)*

mkoval:
I think the functionality looks fine and am not strongly opposed to merging this pull request as it is.

However, even after our discussion offline, I am still not thrilled by the separation of responsibility between the hand and finger position command executors. The two classes are currently so tightly coupled, through the `_distalOnly` and `_stepAngle` parameters that I see dubious value in separating them.

I would prefer to manage *all state related to closing a finger,* including the current breakaway status, inside of the Finger class. The Hand class would simply glue together three instances of Finger class with additional logic to handle the spread joint. This could be elegantly implemented by having each Finger (and Spread) implement a single-DOF version of the `PositionCommand` interface. The Hand class would simply wrap a collection of finger executors (and a spread executor).

Naïvely implementing this, i.e. with a separate thread for each finger, could introduce synchronization issues between the fingers. If you are concerned about that, we could simply add a `spin()` function to each finger that is called in lockstep by the hand executor (e.g. implement a simple event loop).

<details><summary><i>Previously, lgw903 wrote…</i></summary>
> @moval Previous PR comments addressed. Major changes are: 
> - Execution runs on a separate thread
> - `collideWith` passed per execution
> 
> Please take a second look when you have a chance. :) 
</details>


---

Reviewed 12 of 15 files at r2.
Review status: 7 of 9 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[aikido/include/aikido/control/BarrettFingerPositionCommandExecutor.hpp, line 30 \[r2\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KIZgv6AX-639SGzsFjw:-KIZgv6B7p1bTTLF9FqW:1176858923) ([raw file](https://github.com/personalrobotics/aikido/blob/6642d2c868cb04e9f5c5ff1d4dfd47570eaf0e1c/aikido/include/aikido/control/BarrettFingerPositionCommandExecutor.hpp#L30)):*
> ```cpp
>   /// the primal joint. 
>   /// \return mimic ratio.
>   static double getMimicRatio();
> ```

Add a note that `getMimicRatio` is only valid when `_distalOnly` is `false`.

---

*[aikido/include/aikido/control/BarrettHandPositionCommandExecutor.hpp, line 59 \[r2\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KIZhDx_VENBwJcWNxN5:-KIZhDx_VENBwJcWNxN6:-215125162) ([raw file](https://github.com/personalrobotics/aikido/blob/6642d2c868cb04e9f5c5ff1d4dfd47570eaf0e1c/aikido/include/aikido/control/BarrettHandPositionCommandExecutor.hpp#L59)):*
> ```cpp
>     Eigen::Matrix<double, 4, 1> _positions, 
>     std::chrono::milliseconds _duration,
>     ::dart::collision::CollisionGroupPtr _collideWith);
> ```

I don't think this function should accept a `_duration` argument. We don't have control over this variable on the real BarrettHand because the timing is computed by the motor controller using its velocity and acceleration limits. We should do the same here.

---

*[aikido/include/aikido/control/BarrettHandPositionCommandExecutor.hpp, line 89 \[r2\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KIZhcbR5I9kh4frw_ce:-KIZhcbR5I9kh4frw_cf:1735488357) ([raw file](https://github.com/personalrobotics/aikido/blob/6642d2c868cb04e9f5c5ff1d4dfd47570eaf0e1c/aikido/include/aikido/control/BarrettHandPositionCommandExecutor.hpp#L89)):*
> ```cpp
>   /// Manages access on mPromise, mRunning, mSpread 
>   /// mDuration, mCollideWith, mInExecution, mSetPoints
>   std::mutex mSpinMutex;
> ```

I don't think it's true that `mSpinMutex` keeps the thread alive. Isn't that the roll of `mCv` and `mRunning`?

I agree with the "manages access" comment, though.

---

*[aikido/src/control/BarrettHandPositionCommandExecutor.cpp, line 25 \[r2\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KIZhu7HfYflnHJUz9t0:-KIZhu7IeSlFWBpcnict:1568440930) ([raw file](https://github.com/personalrobotics/aikido/blob/6642d2c868cb04e9f5c5ff1d4dfd47570eaf0e1c/aikido/src/control/BarrettHandPositionCommandExecutor.cpp#L25)):*
> ```cpp
>   if (mFingers.size() != numFingers){
>     std::stringstream msg; 
>     msg << "Need " << numFingers << "fingers."; 
> ```

Missing space before `fingers`. Also print the `mFingers.size()` for debugging purposes.

---

*[aikido/src/control/BarrettHandPositionCommandExecutor.cpp, line 219 \[r2\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KIZiZogAbKewcYESTdm:-KIZiZogAbKewcYESTdn:662822615) ([raw file](https://github.com/personalrobotics/aikido/blob/6642d2c868cb04e9f5c5ff1d4dfd47570eaf0e1c/aikido/src/control/BarrettHandPositionCommandExecutor.cpp#L219)):*
> ```cpp
>     mFingers[0]->getDof(0)->setPosition(currentSpread + spreadIncrement);
>     mFingers[1]->getDof(0)->setPosition(currentSpread + spreadIncrement);
>     currentSpread += spreadIncrement;
> ```

We should also stop changing the spread if one of the moving fingers hits something.

---

*[aikido/src/control/CMakeLists.txt, line 4 \[r2\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KIZiuGYCks_rubocklS:-KIZiuGYCks_rubocklT:1929814865) ([raw file](https://github.com/personalrobotics/aikido/blob/6642d2c868cb04e9f5c5ff1d4dfd47570eaf0e1c/aikido/src/control/CMakeLists.txt#L4)):*
> ```CMake
> 	KinematicSimulationTrajectoryExecutor.cpp
> 	BarrettHandPositionCommandExecutor.cpp
> 	BarrettFingerPositionCommandExecutor.cpp
> ```

Please use two spaces for indention in CMake files.

---

*[aikido/include/aikido/control/FingerSimulationStepExecutor.hpp, line 11 \[r1\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KIZgkg82blpZDkExZ9X:-KIZgkg9oM0n-yMauJBI:-525369320) ([raw file](https://github.com/personalrobotics/aikido/blob/5f32229563668e7b6655079a73136448ade5d930/aikido/include/aikido/control/FingerSimulationStepExecutor.hpp#L11)):*
> ```cpp
> /// Assumes that fingers are underactuated: primal joint is actuated,
> /// distal joint moves with certain mimic ratio (set to be Barret hand ratio).
> class FingerSimulationStepExecutor
> ```

This comment is only accurate when `_distalOnly` is `false`. Can you explain the full behavior in this comment?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-:-KIZj4Xc_WcJ4JgyVBkR:2120892026)*

gilwoolee:
@mkoval I made the changes we discussed:
- Actual execution lives in finger controllers, through `execute` and `step` functions.
- Added `BarrettFingerSpreadCommandExecutor`.

None of these spins of its own thread, so user is responsible for doing so and calling step function multiple times. 

<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary>
> I think the functionality looks fine and am not strongly opposed to merging this pull request as it is.
> 
> However, even after our discussion offline, I am still not thrilled by the separation of responsibility between the hand and finger position command executors. The two classes are currently so tightly coupled, through the `_distalOnly` and `_stepAngle` parameters that I see dubious value in separating them.
> 
> I would prefer to manage *all state related to closing a finger,* including the current breakaway status, inside of the Finger class. The Hand class would simply glue together three instances of Finger class with additional logic to handle the spread joint. This could be elegantly implemented by having each Finger (and Spread) implement a single-DOF version of the `PositionCommand` interface. The Hand class would simply wrap a collection of finger executors (and a spread executor).
> 
> Naïvely implementing this, i.e. with a separate thread for each finger, could introduce synchronization issues between the fingers. If you are concerned about that, we could simply add a `spin()` function to each finger that is called in lockstep by the hand executor (e.g. implement a simple event loop).
</details>


---

Review status: 1 of 12 files reviewed at latest revision, 7 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-:-KJSzVbRmmO4OKJaBb-N:2058116326)*

mkoval:
This looks great. There is one bug left: the two spread joints are actually mechanically coupled. This means that both joints should stop moving when either finger collides.

<details><summary><i>Previously, gilwoolee wrote…</i></summary>
> @mkoval I made the changes we discussed:
> - Actual execution lives in finger controllers, through `execute` and `step` functions.
> - Added `BarrettFingerSpreadCommandExecutor`.
> 
> None of these spins of its own thread, so user is responsible for doing so and calling step function multiple times. 
</details>


---

Reviewed 2 of 15 files at r2, 5 of 11 files at r4.
Review status: 6 of 12 files reviewed at latest revision, 15 unresolved discussions, some commit checks failed.

---

*[aikido/include/aikido/control/BarrettFingerSpreadCommandExecutor.hpp, line 32 \[r4\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KJnGyeEp7v6AoENR81h:-KJnGyeEp7v6AoENR81i:-1081588624) ([raw file](https://github.com/personalrobotics/aikido/blob/4ac002fb5c057b4dec3afec0af6aae47e57a78e8/aikido/include/aikido/control/BarrettFingerSpreadCommandExecutor.hpp#L32)):*
> ```cpp
>     ::dart::collision::CollisionDetectorPtr _collisionDetector,
>     ::dart::collision::Option _collisionOptions = ::dart::collision::Option(
>       false, true, 1));
> ```

I am confused by this API: Shouldn't one `Executor` control both spread joints, since they are mechanically coupled?

---

*[aikido/include/aikido/control/BarrettFingerSpreadCommandExecutor.hpp, line 34 \[r4\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KJnGmRWcOgRollVd9s_:-KJnGmRWcOgRollVd9sa:-35006214) ([raw file](https://github.com/personalrobotics/aikido/blob/4ac002fb5c057b4dec3afec0af6aae47e57a78e8/aikido/include/aikido/control/BarrettFingerSpreadCommandExecutor.hpp#L34)):*
> ```cpp
>       false, true, 1));
> 
>   /// Sets varialbes to move the spread joint by _goalPosition,
> ```

Typo: varialbes.

---

*[aikido/include/aikido/control/BarrettFingerSpreadCommandExecutor.hpp, line 43 \[r4\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KJnGqotOiZ2WAWCzIPn:-KJnGqotOiZ2WAWCzIPo:130483634) ([raw file](https://github.com/personalrobotics/aikido/blob/4ac002fb5c057b4dec3afec0af6aae47e57a78e8/aikido/include/aikido/control/BarrettFingerSpreadCommandExecutor.hpp#L43)):*
> ```cpp
>     ::dart::collision::CollisionGroupPtr _collideWith);
> 
>   /// Moves the joint of the finger by certain velocity*_timeSIncePreviousCall
> ```

Typo: _timeSIncePreviousCall

---

*[aikido/include/aikido/control/BarrettHandPositionCommandExecutor.hpp, line 32 \[r4\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KJnH8pHUq-Zm55-yeSG:-KJnH8pHUq-Zm55-yeSH:1547737781) ([raw file](https://github.com/personalrobotics/aikido/blob/4ac002fb5c057b4dec3afec0af6aae47e57a78e8/aikido/include/aikido/control/BarrettHandPositionCommandExecutor.hpp#L32)):*
> ```cpp
>   ///        Third finger should have 2 joints (primal, joint).
>    BarrettHandPositionCommandExecutor(
>     std::vector<BarrettFingerPositionCommandExecutorPtr> _positionCommandExecutors,
> ```

Nit: `std::array<BarrettFingerPositionCommandExecutor, 3>` better captures this requirement at compile time. Or take three separate arguments.

---

*[aikido/include/aikido/control/BarrettHandPositionCommandExecutor.hpp, line 50 \[r4\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KJnHJEjU71u6HZmVSr9:-KJnHJEjU71u6HZmVSrA:-2125076543) ([raw file](https://github.com/personalrobotics/aikido/blob/4ac002fb5c057b4dec3afec0af6aae47e57a78e8/aikido/include/aikido/control/BarrettHandPositionCommandExecutor.hpp#L50)):*
> ```cpp
>   /// Calls step method of finger executors.
>   /// \param _timeSincePreviousCall Time since previous call. 
>   void step(double _timeSincePreviousCall);
> ```

Add a note about locking the skeleton here.

---

*[aikido/src/control/BarrettFingerPositionCommandExecutor.cpp, line 27 \[r4\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KJnHPgFDzVKMQYJGkwj:-KJnHPgFDzVKMQYJGkwk:-1968595266) ([raw file](https://github.com/personalrobotics/aikido/blob/4ac002fb5c057b4dec3afec0af6aae47e57a78e8/aikido/src/control/BarrettFingerPositionCommandExecutor.cpp#L27)):*
> ```cpp
> 
>   if (!mPrimalDof)
>     throw std::invalid_argument("Finger does not have primal dof.");
> ```

Nit: "primal" -> "proximal"

---

*[aikido/src/control/BarrettFingerPositionCommandExecutor.cpp, line 59 \[r4\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KJnHaybivG5hMk2LR7x:-KJnHaybivG5hMk2LR7y:1462882371) ([raw file](https://github.com/personalrobotics/aikido/blob/4ac002fb5c057b4dec3afec0af6aae47e57a78e8/aikido/src/control/BarrettFingerPositionCommandExecutor.cpp#L59)):*
> ```cpp
> 
>   if (!mFinger->isAssembled())
>     throw std::runtime_error("Finger no longer linked.");
> ```

I'm not sure what "linked" means in this error message. "Finger is disassembled" might be be more clear.

---

*[aikido/src/control/BarrettFingerPositionCommandExecutor.cpp, line 113 \[r4\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KJnHwxbgvoQczMK8HcT:-KJnHwxbgvoQczMK8HcU:794207393) ([raw file](https://github.com/personalrobotics/aikido/blob/4ac002fb5c057b4dec3afec0af6aae47e57a78e8/aikido/src/control/BarrettFingerPositionCommandExecutor.cpp#L113)):*
> ```cpp
> 
>   if (distalCollision || 
>       std::abs(mDistalGoalPosition - distalPosition) < kTolerance)
> ```

This logic is incorrect on the distal link. If the proximal link stops, then the distal link will continue moving until it hits something (either an object in the environment or a joint limit).

I believe the logic would be correct if we remove `mDistalGoalPosition` entirely, including this termination check.

---

*[aikido/src/control/BarrettFingerSpreadCommandExecutor.cpp, line 29 \[r4\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KJnIQJ7L5Ifu2fP3s8Z:-KJnIQJ8SQpJGbJLa6Mt:989196428) ([raw file](https://github.com/personalrobotics/aikido/blob/4ac002fb5c057b4dec3afec0af6aae47e57a78e8/aikido/src/control/BarrettFingerSpreadCommandExecutor.cpp#L29)):*
> ```cpp
>   mSpreadCollisionGroup = mCollisionDetector->createCollisionGroup();
>   for(auto body: mFinger->getBodyNodes())
>     mSpreadCollisionGroup->addShapeFramesOf(body);
> ```

Nit: I believe `addShapeFramesOf` supports a `MetaSkeleton` argument, so you could likely replace this loop with `addShapeFramesOf(*mFinger)`.

---

*[aikido/src/control/BarrettHandPositionCommandExecutor.cpp, line 76 \[r4\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KJnIkAwuoxxgzFAfnUG:-KJnIkAwuoxxgzFAfnUH:-629894370) ([raw file](https://github.com/personalrobotics/aikido/blob/4ac002fb5c057b4dec3afec0af6aae47e57a78e8/aikido/src/control/BarrettHandPositionCommandExecutor.cpp#L76)):*
> ```cpp
>   mFingerFutures.clear();
> 
>   // mFingerFutures.reserve(kNumFingers + kNumSpreadFingers);
> ```

Either uncomment this or delete the dead code.

---

*[aikido/src/control/BarrettHandPositionCommandExecutor.cpp, line 80 \[r4\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KJnItgzwOI0wHmVhYap:-KJnItgzwOI0wHmVhYaq:1855493377) ([raw file](https://github.com/personalrobotics/aikido/blob/4ac002fb5c057b4dec3afec0af6aae47e57a78e8/aikido/src/control/BarrettHandPositionCommandExecutor.cpp#L80)):*
> ```cpp
>     mFingerFutures.push_back(std::move(
>       mPositionCommandExecutors[i]->execute(
>         mPrimalGoalPositions(i), mCollideWith)));
> ```

Nit: you could get the same behavior, without explicitly `std::move`ing, if you used `emplace_back` instead of `push_back`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-:-KJnJy9RWPYOZBTMM9Wx:-534819372)*

gilwoolee:
Review status: 1 of 12 files reviewed at latest revision, 15 unresolved discussions.

---

*[aikido/src/control/BarrettFingerPositionCommandExecutor.cpp, line 27 \[r4\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KJnHPgFDzVKMQYJGkwj:-KJvlll4s0dONuGciNA9:-499146009) ([raw file](https://github.com/personalrobotics/aikido/blob/4ac002fb5c057b4dec3afec0af6aae47e57a78e8/aikido/src/control/BarrettFingerPositionCommandExecutor.cpp#L27)):*
<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary>
> Nit: "primal" -> "proximal"
</details>
Done.

---

*[aikido/src/control/BarrettFingerSpreadCommandExecutor.cpp, line 29 \[r4\]](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-KJnIQJ7L5Ifu2fP3s8Z:-KJvqgleHwoqLNz1skV7:-1519177734) ([raw file](https://github.com/personalrobotics/aikido/blob/4ac002fb5c057b4dec3afec0af6aae47e57a78e8/aikido/src/control/BarrettFingerSpreadCommandExecutor.cpp#L29)):*
<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary>
> Nit: I believe `addShapeFramesOf` supports a `MetaSkeleton` argument, so you could likely replace this loop with `addShapeFramesOf(*mFinger)`.
</details>
It takes `Skeleton*` not `MetaSkeleton*`. :(

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/personalrobotics/aikido/101)*

mkoval:
:lgtm:

<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary>
> This looks great. There is one bug left: the two spread joints are actually mechanically coupled. This means that both joints should stop moving when either finger collides.
</details>


---

Reviewed 3 of 15 files at r2, 2 of 11 files at r4, 9 of 9 files at r5.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/personalrobotics/aikido/101#-:-KK19O-vBPwrUDSIgyAZ:1417088721)*

mkoval:
The build is failing on Travis with this linking error:
```
../../src/control/libaikido_control.so: undefined reference to `aikido::control::BarrettHandPositionCommandExecutor::kWaitPeriod'
```
I suspect that this linking error is occurring because `kWaitPeriod` is a `constexpr` variable that we're odr-using without a namespace-level definition. See this [StackOverflow post](http://stackoverflow.com/a/28446388/111426) for an explanation of this bizarre quirk of the C++ standard. The fix is simply to add an empty namespace-level definition to a `.cpp` file.

mkoval:
Reviewed 1 of 1 files at r6.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/personalrobotics/aikido/101)*

PR #102, opened 2016-05-27
Feature/kinbodyparser

Shushman:
A first pass at a Kinbody Parser under `aikido::util`. Points to note:
1. Only works if the `KinBody` has one `Body` element with 1 or 2 `Geom` elements, and `Render` and `Data` both defined, and refer to some file.
2. I've mostly followed the template in `SkelParser` 
3. A problem - The plastic bowl and glass are two files that are read in just fine, but I get a strange error (originating from `dart::dynamics::MeshShap::loadMesh`) upon trying to load say the Pop tarts or kinbody. The error message is:

```
Warning [MeshShape.cpp:349] [MeshShape::loadMesh] Failed loading mesh 'package://pr_ordata/data/objects/kinova_tool.stl 0.1'.
Error [KinBodyParser.cpp:423] [KinBodyParser] Fail to load model[kinova_tool.stl 0.1].
```

Sometimes, even without the error message, the Pop tarts looks like:

rviz_aikido_kinbodyloader

---

This change is [Reviewable](https://reviewable.io/reviews/personalrobotics/aikido/102)


mkoval:
Reviewed 2 of 3 files at r1, 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 12 unresolved discussions.

---

*[aikido/include/aikido/util/KinBodyParser.h, line 8 \[r1\]](https://reviewable.io:443/reviews/personalrobotics/aikido/102#-KIny73rsIG7hoQCkdz4:-KIny73rsIG7hoQCkdz5:-1206977591) ([raw file](https://github.com/personalrobotics/aikido/blob/f11bb86054dadcb3304db8ff253a439ea0b6acf2/aikido/include/aikido/util/KinBodyParser.h#L8)):*
> ```objc
> #include "dart/common/Uri.h"
> 
> //Currently assumes only one body node
> ```

Move this comment into the docstring of `readKinBodyXMLFile`.

---

*[aikido/include/aikido/util/KinBodyParser.h, line 14 \[r1\]](https://reviewable.io:443/reviews/personalrobotics/aikido/102#-KInyBBMPqZDUkFwTiwW:-KInyBBN-4a7UTh4lqjA:-1352075956) ([raw file](https://github.com/personalrobotics/aikido/blob/f11bb86054dadcb3304db8ff253a439ea0b6acf2/aikido/include/aikido/util/KinBodyParser.h#L14)):*
> ```objc
> namespace util {
> 
> namespace KinBodyParser {
> ```

The `namespace` hierarchy should *exactly* match the filesystem hierarchy. I suggest removing this `namespace` and leaving the file where it is unless `readKinBodyXMLFile` relies on an optional dependency.

---

*[aikido/include/aikido/util/KinBodyParser.h, line 18 \[r1\]](https://reviewable.io:443/reviews/personalrobotics/aikido/102#-KIny3zPDMRdP6xKB0TK:-KIny3zPDMRdP6xKB0TL:-1678877000) ([raw file](https://github.com/personalrobotics/aikido/blob/f11bb86054dadcb3304db8ff253a439ea0b6acf2/aikido/include/aikido/util/KinBodyParser.h#L18)):*
> ```objc
>   dart::dynamics::SkeletonPtr readKinBodyXMLFile(
>     const dart::common::Uri& _fileUri,
>     const dart::common::ResourceRetrieverPtr& _retriever = nullptr);
> ```

- Missing docstring.
- Don't indent the contents of a `namespace`.

---

*[aikido/src/util/KinBodyParser.cpp, line 159 \[r1\]](https://reviewable.io:443/reviews/personalrobotics/aikido/102#-KInyuRyo5TT-MAhECpt:-KInyuRyo5TT-MAhECpu:802692996) ([raw file](https://github.com/personalrobotics/aikido/blob/f11bb86054dadcb3304db8ff253a439ea0b6acf2/aikido/src/util/KinBodyParser.cpp#L159)):*
> ```cpp
>   {
>     dterr << "[KinBodyParser] KinBody  file[" << _fileUri.toString()
>           << "] does not contain <KinBody> as the element.\n";
> ```

Typo: "the element" should be "the root element"?

---

*[aikido/src/util/KinBodyParser.cpp, line 178 \[r1\]](https://reviewable.io:443/reviews/personalrobotics/aikido/102#-KInz-ffOwkCPgo5WQqd:-KInz-ffOwkCPgo5WQqe:-341608343) ([raw file](https://github.com/personalrobotics/aikido/blob/f11bb86054dadcb3304db8ff253a439ea0b6acf2/aikido/src/util/KinBodyParser.cpp#L178)):*
> ```cpp
> {
> 
>   assert(_KinBodyElement != nullptr);
> ```

Handle this by returning`nullptr` instead of triggering an `assert`ion.

---

*[aikido/src/util/KinBodyParser.cpp, line 185 \[r1\]](https://reviewable.io:443/reviews/personalrobotics/aikido/102#-KInz7MgFOW_zQu4C-i9:-KInz7MgFOW_zQu4C-iA:-1523763220) ([raw file](https://github.com/personalrobotics/aikido/blob/f11bb86054dadcb3304db8ff253a439ea0b6acf2/aikido/src/util/KinBodyParser.cpp#L185)):*
> ```cpp
>   // Name attribute
>   std::string name = dart::utils::getAttributeString(_KinBodyElement, "name");
>   newSkeleton->setName(name);
> ```

What happens if there is no `name` attribute?

---

*[aikido/src/util/KinBodyParser.cpp, line 209 \[r1\]](https://reviewable.io:443/reviews/personalrobotics/aikido/102#-KIo-MK44C_hNpIkS03C:-KIo-MK50Y2YuPNnw7P_:-1372966925) ([raw file](https://github.com/personalrobotics/aikido/blob/f11bb86054dadcb3304db8ff253a439ea0b6acf2/aikido/src/util/KinBodyParser.cpp#L209)):*
> ```cpp
>       Eigen::make_aligned_shared<dart::dynamics::FreeJoint::Properties>(
>             dart::dynamics::Joint::Properties("root_joint", newBodyNode.initTransform));
>   rootJoint.type = "free";
> ```

What is the point of creating this temporary `SkelJoint` structure? It seems like you could directly pass the `FreeJoint::Properties` into the `createJointAndBodyNodePair` function without all of this additional indirection.

---

*[aikido/src/util/KinBodyParser.cpp, line 235 \[r1\]](https://reviewable.io:443/reviews/personalrobotics/aikido/102#-KIo1PiOXI1J1hSW5vRf:-KIo1PiOXI1J1hSW5vRg:-300974167) ([raw file](https://github.com/personalrobotics/aikido/blob/f11bb86054dadcb3304db8ff253a439ea0b6acf2/aikido/src/util/KinBodyParser.cpp#L235)):*
> ```cpp
>   const dart::common::ResourceRetrieverPtr& _retriever)
> {
>   assert(_bodyNodeElement != nullptr);
> ```

Handle this gracefully instead of triggering an `assert`ion.

---

*[aikido/src/util/KinBodyParser.cpp, line 241 \[r1\]](https://reviewable.io:443/reviews/personalrobotics/aikido/102#-KIo1SQoAuP3Pk_so4Vs:-KIo1SQoAuP3Pk_so4Vt:-1523763220) ([raw file](https://github.com/personalrobotics/aikido/blob/f11bb86054dadcb3304db8ff253a439ea0b6acf2/aikido/src/util/KinBodyParser.cpp#L241)):*
> ```cpp
> 
>   // Name attribute
>   newBodyNode->mName = dart::utils::getAttributeString(_bodyNodeElement, "name");
> ```

What happens if there is no `name` attribute?

---

*[aikido/src/util/KinBodyParser.cpp, line 308 \[r2\]](https://reviewable.io:443/reviews/personalrobotics/aikido/102#-KK4hT4npreqPAsDXocv:-KK4hT4otEKW1e2sZKBc:-1989141141) ([raw file](https://github.com/personalrobotics/aikido/blob/f336f227d73bd112d9cda88dcea7d7b3f79280c8/aikido/src/util/KinBodyParser.cpp#L308)):*
> ```cpp
>       assert(0);
>     }
>     else if(element_count == 1){
> ```

Can you add a comment explaining the logic in here? I am not sure what this is trying to accomplish.

---

*[aikido/src/util/KinBodyParser.cpp, line 413 \[r2\]](https://reviewable.io:443/reviews/personalrobotics/aikido/102#-KK4hlUdS26Y_PgcxI86:-KK4hlUdS26Y_PgcxI87:1902845451) ([raw file](https://github.com/personalrobotics/aikido/blob/f336f227d73bd112d9cda88dcea7d7b3f79280c8/aikido/src/util/KinBodyParser.cpp#L413)):*
> ```cpp
>   {
>     std::string filename_ext = dart::utils::getValueString(vizOrColEle, fieldName);
>     std::string filename = filename_ext.substr(0,filename_ext.find(" "));
> ```

What is the purpose of this? Please add a comment.

---

*[aikido/src/util/KinBodyParser.cpp, line 416 \[r2\]](https://reviewable.io:443/reviews/personalrobotics/aikido/102#-KK4houUABHZOc31BtsQ:-KK4houUABHZOc31BtsR:96956344) ([raw file](https://github.com/personalrobotics/aikido/blob/f336f227d73bd112d9cda88dcea7d7b3f79280c8/aikido/src/util/KinBodyParser.cpp#L416)):*
> ```cpp
>     const std::string meshUri = dart::common::Uri::getRelativeUri(_baseUri, filename);
>     const aiScene* model = dart::dynamics::MeshShape::loadMesh(meshUri, _retriever);
>     const Eigen::Vector3d scale(1.0,1.0,1.0); //Default scale as kinbody does not have info
> ```

Shouldn't we read the scale from the file, if it is specified?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/personalrobotics/aikido/102)*

jslee02:
How about we add this feature to DART? I'd be happy to help to transition this PR to DART repo and resolve the remaining issues.

mkoval:
@Shushman What is the current status of this? Do you intend to address the remaining comments so we can merge this into `master`?

While I think this is a great addition, I am also fine with dropping support if you think we are at a point where we no longer need to regularly load `.kinbody.xml` files into DART. Do we have URDF versions of the critical files we need for pantry loading, etc?

jslee02:
I'm updating this PR as @Shushman almost finished. 

My little concern is that there are some discrepancies between the [spec](http://openrave.programmingvision.com/wiki/index.php/Format:XML) and kinbody files in our [repo](https://github.com/personalrobotics/pr-ordata/tree/bd22128783fbe50cbfe24d6484321e93926d18a5/data/objects). For example, the spec has `<collision>` for collision shape but it seems our kinbody files uses <data> for it instead. I'll make the parser to be able to successfully parse our kinbody files then try to obey the spec.

jslee02:
Replied in the each comment.

---

Review status: 0 of 10 files reviewed at latest revision, 12 unresolved discussions.

---

*[src/util/KinBodyParser.cpp, line 159 at r1](https://reviewable.io:443/reviews/personalrobotics/aikido/102#-KInyuRyo5TT-MAhECpt:-KfXNPskCzn1lsl-tSw8:b-7rl0jd) ([raw file](https://github.com/personalrobotics/aikido/blob/f11bb86054dadcb3304db8ff253a439ea0b6acf2/src/util/KinBodyParser.cpp#L159)):*
<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary><blockquote>

Typo: "the element" should be "the root element"?
</blockquote></details>

done.

---

*[src/util/KinBodyParser.cpp, line 178 at r1](https://reviewable.io:443/reviews/personalrobotics/aikido/102#-KInz-ffOwkCPgo5WQqd:-KfXNbM8t8gU7GKmiUpS:b-lrwkqs) ([raw file](https://github.com/personalrobotics/aikido/blob/f11bb86054dadcb3304db8ff253a439ea0b6acf2/src/util/KinBodyParser.cpp#L178)):*
<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary><blockquote>

Handle this by returning`nullptr` instead of triggering an `assert`ion.
</blockquote></details>

The `nullptr` handling should be done by the caller. Added comment for the reason.

---

*[src/util/KinBodyParser.cpp, line 185 at r1](https://reviewable.io:443/reviews/personalrobotics/aikido/102#-KInz7MgFOW_zQu4C-i9:-KfXOPdASWnNASL62Tkn:bvqbewy) ([raw file](https://github.com/personalrobotics/aikido/blob/f11bb86054dadcb3304db8ff253a439ea0b6acf2/src/util/KinBodyParser.cpp#L185)):*
<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary><blockquote>

What happens if there is no `name` attribute?
</blockquote></details>

`getAttributeStrin()` returns an empty string for that case, and empty named skeleton will not cause any problem in parsing. Will add a message here to warn people just in case that's not their intention.

---

*[src/util/KinBodyParser.cpp, line 209 at r1](https://reviewable.io:443/reviews/personalrobotics/aikido/102#-KIo-MK44C_hNpIkS03C:-KfXOwM42D_-MWPX5tW7:b-8j5tg8) ([raw file](https://github.com/personalrobotics/aikido/blob/f11bb86054dadcb3304db8ff253a439ea0b6acf2/src/util/KinBodyParser.cpp#L209)):*
<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary><blockquote>

What is the point of creating this temporary `SkelJoint` structure? It seems like you could directly pass the `FreeJoint::Properties` into the `createJointAndBodyNodePair` function without all of this additional indirection.
</blockquote></details>

Removed the indirection.

---

*[src/util/KinBodyParser.cpp, line 235 at r1](https://reviewable.io:443/reviews/personalrobotics/aikido/102#-KIo1PiOXI1J1hSW5vRf:-KfXOzmRJa0MbSGpbyra:b-bizqh8) ([raw file](https://github.com/personalrobotics/aikido/blob/f11bb86054dadcb3304db8ff253a439ea0b6acf2/src/util/KinBodyParser.cpp#L235)):*
<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary><blockquote>

Handle this gracefully instead of triggering an `assert`ion.
</blockquote></details>

Added comment.

---

*[src/util/KinBodyParser.cpp, line 241 at r1](https://reviewable.io:443/reviews/personalrobotics/aikido/102#-KIo1SQoAuP3Pk_so4Vs:-KfXP4-fNNXVji3ltIlB:buun2t1) ([raw file](https://github.com/personalrobotics/aikido/blob/f11bb86054dadcb3304db8ff253a439ea0b6acf2/src/util/KinBodyParser.cpp#L241)):*
<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary><blockquote>

What happens if there is no `name` attribute?
</blockquote></details>

An empty string will be returned. We prints error message because an empty-named BodyNode causes problems in parsing (especially for parsing <Geom>). If `bodyNodeInfo.valid == false`, the caller will return `nullptr` skeleton.

---

*[src/util/KinBodyParser.cpp, line 308 at r2](https://reviewable.io:443/reviews/personalrobotics/aikido/102#-KK4hT4npreqPAsDXocv:-KfXPW-fPLxki5hloFrm:b-hht9z3) ([raw file](https://github.com/personalrobotics/aikido/blob/f336f227d73bd112d9cda88dcea7d7b3f79280c8/src/util/KinBodyParser.cpp#L308)):*
<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary><blockquote>

Can you add a comment explaining the logic in here? I am not sure what this is trying to accomplish.
</blockquote></details>

The logic is modified with some reasonable comments.

---

*[src/util/KinBodyParser.cpp, line 413 at r2](https://reviewable.io:443/reviews/personalrobotics/aikido/102#-KK4hlUdS26Y_PgcxI86:-KfXPe4Iej6IknIhn5wE:b-2f157e) ([raw file](https://github.com/personalrobotics/aikido/blob/f336f227d73bd112d9cda88dcea7d7b3f79280c8/src/util/KinBodyParser.cpp#L413)):*
<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary><blockquote>

What is the purpose of this? Please add a comment.
</blockquote></details>

I believe the original code is to parse <Render> and <Data> that contains the mesh file name and optionally the mesh scale. Added comment.

---

*[src/util/KinBodyParser.cpp, line 416 at r2](https://reviewable.io:443/reviews/personalrobotics/aikido/102#-KK4houUABHZOc31BtsQ:-KfXPs-bLRbwwA-_U3El:b-896fix) ([raw file](https://github.com/personalrobotics/aikido/blob/f336f227d73bd112d9cda88dcea7d7b3f79280c8/src/util/KinBodyParser.cpp#L416)):*
<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary><blockquote>

Shouldn't we read the scale from the file, if it is specified?
</blockquote></details>

Done.

---

*[aikido/include/aikido/util/KinBodyParser.h, line 8 at r1](https://reviewable.io:443/reviews/personalrobotics/aikido/102#-KIny73rsIG7hoQCkdz4:-KfXPvAoFqjAW0gMfjOY:b-896fix) ([raw file](https://github.com/personalrobotics/aikido/blob/f11bb86054dadcb3304db8ff253a439ea0b6acf2/aikido/include/aikido/util/KinBodyParser.h#L8)):*
<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary><blockquote>

Move this comment into the docstring of `readKinBodyXMLFile`.
</blockquote></details>

Done.

---

*[aikido/include/aikido/util/KinBodyParser.h, line 14 at r1](https://reviewable.io:443/reviews/personalrobotics/aikido/102#-KInyBBMPqZDUkFwTiwW:-KfXPw18UFqtVTrF9rIB:bgflu89) ([raw file](https://github.com/personalrobotics/aikido/blob/f11bb86054dadcb3304db8ff253a439ea0b6acf2/aikido/include/aikido/util/KinBodyParser.h#L14)):*
<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary><blockquote>

The `namespace` hierarchy should *exactly* match the filesystem hierarchy. I suggest removing this `namespace` and leaving the file where it is unless `readKinBodyXMLFile` relies on an optional dependency.
</blockquote></details>

Done. The parser doesn't require any optional dependency.

---

*[aikido/include/aikido/util/KinBodyParser.h, line 18 at r1](https://reviewable.io:443/reviews/personalrobotics/aikido/102#-KIny3zPDMRdP6xKB0TK:-KfXPuBGRNj89o2RlgVZ:b-896fix) ([raw file](https://github.com/personalrobotics/aikido/blob/f11bb86054dadcb3304db8ff253a439ea0b6acf2/aikido/include/aikido/util/KinBodyParser.h#L18)):*
<details><summary><i>Previously, mkoval (Michael Koval) wrote…</i></summary><blockquote>

- Missing docstring.
- Don't indent the contents of a `namespace`.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/personalrobotics/aikido/102#-:-KfXQD-Mjq5j4BUDill_:bz3y568)*

jslee02:
@mkoval @Shushman I believe all the issues pointed out @mkoval are addressed in the recent commits.

The visual test is not done. I'm not familiar with using rviz so it would be grateful if someone could do that if the visualization issue @Shushman mentioned is still there. :sweat_smile:

mkoval:
> My little concern is that there are some discrepancies between the spec and kinbody files in our repo. For example, the spec has `<collision>` for collision shape but it seems our kinbody files uses for it instead. I'll make the parser to be able to successfully parse our kinbody files then try to obey the spec.

Can you compile a list of those discrepancies as you encounter them? Sadly, the spec for `.kinbody.xml` files is incomplete and actively misleading in some sections (particularly: how collision and visual geometry interact, especially for primitive shapes). The reality is that "the spec" is "whatever OpenRAVE and QtCoin do." I mostly developed `or_rviz` by tweaking the logic until it produced the same output as QtCoin on a variety of objects. 😓

jslee02:
Current KinBody parser only parses a subset of the format assuming it contains one body (with the free joint) per file (or string). 

My understandings for the relevant spec of KinBody are:
1. `<geom>` has `type` attributes with one of four values: `box`, `sphere`, `trimesh`, `cylinder`, and each type has the different set of subfields under `<geom>`.
3. `box`, `sphere`, `cylinder` selectively have the subfields among `<extents>`, `<height>`, `<radius>`, and they assumed to have both of collision and render shapes in the simulation. There is `render` attribute of `<geom>`, but this is usually omitted in our files.
4. `trimesh` has at least one of `<collision>` and `<render>`.

The discrepancies I found so far were:
* Use of `<Data>` instead of `<collision>`.
* `kinova_tool.kinbody.xml` contains `<Render>` field while the geom type is `sphere`. I modified the geom type to `trimesh` for the test file in this repo.

It's not a discrepancy but KinBody doesn't seem case sensitive. So our KinBody parser converts all the field names to lower cases and proceeds the parsing.

mkoval:
> `box`, `sphere`, `cylinder` selectively have the subfields among `<extents>`, `<height>`, `<radius>`, and **they [are] assumed to have both of collision and render shapes in the simulation. There is [a] render attribute of `<geom>`, but this is usually omitted in our files.**

The bolded part is incorrect. The `<geom>` tag directly defines collision geometry. The following logic determines what is used as the visual geometry:

1. If `<geom>`'s `render` attribute is `false`, display nothing. Otherwise...
2. If `<render>` is present, render the trimesh at the specified path. Otherwise...
3. Render the collision geometry.

The easiest way to understand this is realize that each `<geom>` tag is parsed into one instance of the [`GeometryInfo` class](https://github.com/rdiankov/openrave/blob/3fa55b84c3a6fa9d704f194ccb392db73f3024cd/include/openrave/kinbody.h#L130) by the [`GeometryInfoReader` class](https://github.com/rdiankov/openrave/blob/18521d4ee6311464522a4e98cf0eed883a2d15d3/src/libopenrave/xmlreaders.cpp#L179). Note that `GeometryInfo` has both a `_filenamerender` member variable (for case (2) above) _and_ `_meshcollision` (which is loaded from `_filenamecollision` for `type="trimesh"` collision geometry).

> There is render attribute of `<geom>`, but this is usually omitted in our files.

This is correct. However, note that the **`render="true"` attribute** is a boolean that indicates whether or not the `<geom>` should be rendered *at all* (case (1) above). The **`<render>` element** specifies a path (and, optionally, a scale) for a trimesh file that should be used for visual geometry (case (2) above). The attribute defaults to `true`, so we do not typically specify it.

> `trimesh` has at least one of `<collision>` and `<render>`.

It may have both and they have different meanings. `<collision>` (or `<data>`, see below) are used for `<geom>` with `type="trimesh"` to specify the path to a trimesh file to use for collision geometry. `<render>` is used by *all* `<geom>`s to specify a visual trimesh (case (2) above).

> Use of `<Data>` instead of `<collision>`.

These [are treated as equivalent in `GeometryInfoReader`](https://github.com/rdiankov/openrave/blob/18521d4ee6311464522a4e98cf0eed883a2d15d3/src/libopenrave/xmlreaders.cpp#L399).

> `kinova_tool.kinbody.xml` contains `<Render>` field while the `<geom>` type is `sphere`. I modified the `<geom>` type to `<trimesh>` for the test file in this repo.

This is **not** correct. From the above description, it is impossible to create visual geometry *without* any associated collision geometry. There is a `type="none"` which theoretically supports this, but (if I remember correctly) it is not handled correctly everywhere in OpenRAVE (see https://github.com/rdiankov/openrave/pull/306 for some effort I made in that direction). As a workaround, we often create a zero-radius sphere to represent "visual-only" geometry (e.g. see [this comment in `or_urdf`](https://github.com/personalrobotics/or_urdf/blob/395e2732dc8974bcaaa060eac245972ce6724a69/src/urdf_loader.cpp#L331)).

---

I hope this helps. **TL;DR:** Your best bet is to mimic `GeometryInfoReader` as closely as possible.

jslee02:
Thank you for helping me to understand the mysterious visualization logic of `<Geom>`. :sweat_smile: 

I believe the parser now correctly read KinBody files; it could read the files in `pr-ordata` without any modification.

include/aikido/util/KinBodyParser.hpp
```diff
@@ -0,0 +1,61 @@
...
+///
+/// This function only parses a subset of the format assuming only one body node
+/// in a kinbody file.
+///
+/// Followings are the available fields that this parser can read.
+/// \code
+/// kinbody - attributes: name
+///   body - attributes: name
+///     geom - attributes: name, type* (none, box, sphere, trimesh, cylinder), render
+///       data* (or collision) - file* scale (e.g., /path/to/mesh/file_collision.stl 0.5) [for trimesh]
```
mkoval: Note that `scale` may be a single `float or  three `float`s, for (respectively) the x, y, and z-axes. This is implemented in OpenRAVE in [this cryptic snippet](https://github.com/rdiankov/openrave/blob/18521d4ee6311464522a4e98cf0eed883a2d15d3/src/libopenrave/xmlreaders.cpp#L335):
```c++
_ss >> _pgeom->_filenamerender;
_ss >> _pgeom->_vRenderScale.x; _pgeom->_vRenderScale.y = _pgeom->_vRenderScale.z = _pgeom->_vRenderScale.x;
_ss >> _pgeom->_vRenderScale.y >> _pgeom->_vRenderScale.z;
```

include/aikido/util/KinBodyParser.hpp
```diff
@@ -0,0 +1,61 @@
...
+///       extents* - 3 float [for box]
+///       height* - float [for cylinder]
+///       radius* - float [for cylinder and sphere]
+///       render - file* scale (e.g., /path/to/mesh/file_render.stl 0.5)
```
mkoval: See my note above about `scale`. The same applies here (see [this snippet](https://github.com/rdiankov/openrave/blob/18521d4ee6311464522a4e98cf0eed883a2d15d3/src/libopenrave/xmlreaders.cpp#L400)).

include/aikido/util/KinBodyParser.hpp
```diff
@@ -0,0 +1,61 @@
...
+/// \endcode
+/// Elements marked with `*` are required ones.
+///
+/// The detail of the format can be found at:
+/// http://openrave.programmingvision.com/wiki/index.php/Format:XML).
```
mkoval: Extra `).` at the end of this line.

include/aikido/util/KinBodyParser.hpp
```diff
@@ -0,0 +1,61 @@
...
+///
+/// \param[in] kinBodyString The Kinbody XML string.
+/// \param[in] baseUri The base URI of the mesh files in the KinBody XML string.
```
mkoval: What does the default value of `baseUri` do?

include/aikido/util/KinBodyParser.hpp
```diff
@@ -0,0 +1,61 @@
...
+
+/// Read skeleton from a file of OpenRAVE's custom XML format
+///
+/// This function only parses a subset of the format assuming only one body node
+/// in a kinbody file.
+///
+/// The detail of the format can be found at:
+/// http://openrave.programmingvision.com/wiki/index.php/Format:XML).
+///
+/// \param[in] kinBodyFileUri The file URI ("file://...") of the KinBody file.
```
mkoval: Why must this be a `file://` URI?

src/util/KinBodyParser.cpp
```diff
@@ -0,0 +1,532 @@
...
+  jointProps.mName = DEFAULT_KINBODY_ROOT_JOINTNAME;
+  jointProps.mT_ParentBodyToJoint = bodyNodeInfo.initTransform;
+
+  auto jointAndBodyNode = skeleton->createJointAndBodyNodePair<
+      dart::dynamics::FreeJoint>(nullptr, jointProps, bodyNodeInfo.properties);
+  assert(jointAndBodyNode.first && jointAndBodyNode.second);
+
+  readGeoms(jointAndBodyNode.second, bodyEle, baseUri, retriever);
+  skeleton->resetPositions();
+  skeleton->resetVelocities();
```
mkoval: Nit: Why are these calls to `resetPositions` and `resetVelocities` necessary? We just created `Skeleton` earlier in this function.

src/util/KinBodyParser.cpp
```diff
@@ -0,0 +1,532 @@
...
+
+  // Get Body node
+  auto bodyEle = kinBodyEle->FirstChildElement("body");
+  if (bodyEle == nullptr)
+  {
+    dterr << "[KinBodyParser] KinBody document '" << baseUri.toString()
+          << "' does not contain <Body> element "
+          << "under <KinBody> element.\n";
+    return nullptr;
+  }
```
mkoval: Nit: I'd like to raise an error if there is more than one `body` element as well. Otherwise, we may end up silently loading only the first `BodyNode` in the file!

src/util/KinBodyParser.cpp
```diff
@@ -0,0 +1,532 @@
...
+  // bodyNodeElement shouldn't be nullptr at all, which always should be
+  // gauranteed because this function is only used internally to be so. If we
+  // find the nullptr case, then we should fix the caller.
+
+  BodyNodeInfo bodyNodeInfo;
+
+  dart::dynamics::BodyNode::Properties properties;
+
+  // Name attribute
+  properties.mName = "NoName";
```
mkoval: Nit: Replace this with a `DEFAULT_KINBODY_BODYNODE_NAME` constant, like you did for `KinBody` above.

src/util/KinBodyParser.cpp
```diff
@@ -0,0 +1,532 @@
...
+  const auto fileNameAndScale
+      = split(dart::utils::getValueString(geomEle, renderOrData), " ");
+  assert(!fileNameAndScale.empty());
+
+  ret.first = fileNameAndScale[0];
+
+  if (fileNameAndScale.size() > 1u)
+  {
+    ret.second = std::atof(fileNameAndScale[1].c_str());
+    // TODO(JS): warning if zero scale
```
mkoval: Add this.

src/util/KinBodyParser.cpp
```diff
@@ -0,0 +1,532 @@
...
+
+  const auto fileNameAndScale
+      = split(dart::utils::getValueString(geomEle, renderOrData), " ");
+  assert(!fileNameAndScale.empty());
+
+  ret.first = fileNameAndScale[0];
+
+  if (fileNameAndScale.size() > 1u)
+  {
+    ret.second = std::atof(fileNameAndScale[1].c_str());
```
mkoval: Check for a conversion error. I am not sure if this is possible with `atof`, so you may need to use a `stringstream` instead.

src/util/KinBodyParser.cpp
```diff
@@ -0,0 +1,532 @@
...
+
+//==============================================================================
+std::pair<std::string, double> resolveFileNameAndScale(
+    tinyxml2::XMLElement* geomEle,
+    const std::string& renderOrData)
+{
+  // <Render> or <Data> contains the relative mesh file path and optionally the
+  // mesh scale (e.g., <Render>my/mesh/file.stl<Render> or
+  // <Render>my/mesh/file.stl 0.25<Render>). If the scale is not provided then
+  // this function returns 1.0 by default.
```
mkoval: See my note above. There is a third form that accepts three scale parameters.

src/util/KinBodyParser.cpp
```diff
@@ -0,0 +1,532 @@
...
+  auto shapeNodeName = resolveShapeName(geomEle, bodyNode);
+
+  if (typeAttr == "none")
+  {
+    // Do nothing.
+  }
+  else if (typeAttr == "box")
+  {
+    auto extents = dart::utils::getValueVector3d(geomEle, "extents");
+    auto shape = std::make_shared<dart::dynamics::BoxShape>(extents);
```
mkoval: This is incorrect. I believe OpenRAVE uses `extents` to refer to *half* extents, whereas DART's `BoxShape` accepts *full* extents. The `BoxShape` should be constructed with `2.0 * extents`.

src/util/KinBodyParser.cpp
```diff
@@ -0,0 +1,532 @@
...
+
+    shapeNode = bodyNode->createShapeNodeWith<
+        dart::dynamics::CollisionAspect,
+        dart::dynamics::DynamicsAspect>(shape, shapeNodeName);
```
mkoval: I don't think we should create a `DyanmicsAspect` on all of these `ShapeNode`s, since they may not have inertial or friction properties specified.

src/util/KinBodyParser.cpp
```diff
@@ -0,0 +1,532 @@
...
+  }
+  else if (typeAttr == "sphere")
+  {
+    auto radius = dart::utils::getValueDouble(geomEle, "radius");
```
mkoval: Check for non-positive `radius`.

src/util/KinBodyParser.cpp
```diff
@@ -0,0 +1,532 @@
...
+    auto shape = std::make_shared<dart::dynamics::SphereShape>(radius);
+
+    shapeNode = bodyNode->createShapeNodeWith<
+        dart::dynamics::CollisionAspect,
+        dart::dynamics::DynamicsAspect>(shape, shapeNodeName);
+  }
+  else if (typeAttr == "cylinder")
+  {
+    auto radius = dart::utils::getValueDouble(geomEle, "radius");
+    auto height = dart::utils::getValueDouble(geomEle, "height");
```
mkoval: Check for non-positive `radius` and `height`.

src/util/KinBodyParser.cpp
```diff
@@ -0,0 +1,532 @@
...
+
+  auto shapeNodeName = resolveShapeName(geomEle, bodyNode);
+
+  if (typeAttr == "none")
+  {
+    // Do nothing.
+  }
+  else if (typeAttr == "box")
+  {
+    auto extents = dart::utils::getValueVector3d(geomEle, "extents");
```
mkoval: Check for non-positive `extents`.

src/util/KinBodyParser.cpp
```diff
@@ -0,0 +1,532 @@
...
+    else if (hasDataEle)
+    {
+      collisionOrData = "data";
+    }
+    else
+    {
+      dterr << "[KinBodyParser] <Geom> element doesn't contain neither of "
+            << "<Collision> and <Data> as the child element. Ignoring this "
+            << "<Geom>.\n";
+      return;
```
mkoval: `return`ing here also bypasses creation of visual geometry.

src/util/KinBodyParser.cpp
```diff
@@ -0,0 +1,532 @@
...
+
+    shapeNode = bodyNode->createShapeNodeWith<
+        dart::dynamics::CollisionAspect,
+        dart::dynamics::DynamicsAspect>(shape, shapeNodeName);
+  }
+  else
+  {
+    dterr << "[KinBodyParser] Attempts to parse unsupported geom type '"
+          << typeAttr << "'. Ignoring this <Geom>.\n";
+    return;
```
mkoval: `return`ing here also bypasses creation of visual geometry.

src/util/KinBodyParser.cpp
```diff
@@ -0,0 +1,532 @@
...
+  if (hasRenderEle)
+  {
+    auto fileNameAndScaleOfRender = resolveFileNameAndScale(geomEle, "render");
+    auto fileNameOfRender = fileNameAndScaleOfRender.first;
+    auto uniScaleOfRender = fileNameAndScaleOfRender.second;
+    auto renderShapeNodeName = resolveShapeName(geomEle, bodyNode);
+
+    if (typeAttr == "trimesh"
+        && fileNameOfCollision == fileNameOfRender
+        && uniScaleOfCollision == uniScaleOfRender)
```
mkoval: Very nice! 👍

src/util/KinBodyParser.cpp
```diff
@@ -0,0 +1,532 @@
...
+    tinyxml2::XMLDocument& doc)
+{
+  auto firstEle = doc.FirstChildElement();
+  if (firstEle)
+    transformElementNamesToLowerCasesRecurse(firstEle);
+}
+
+//==============================================================================
+std::vector<std::string> split(
+    const std::string& str, const std::string& delimiters)
```
mkoval: Nice helper function. I would be open to moving this to `util`, since it may be useful elsewhere.

include/aikido/util/KinBodyParser.hpp
jslee02: The default value of `baseUri` is empty. An empty base URI means that the URIs for mesh files should be absolute URIs or absolute paths. Added this to the docstring.

include/aikido/util/KinBodyParser.hpp
jslee02: It shouldn't be. Updated the comment: "The URI to a KinBody file. If the URI scheme is not file (i.e., file://), a relevant ResourceRetriever should be passed to retrieve the KinBody file."

src/util/KinBodyParser.cpp
jslee02: I thought we don't want to create visual geometry for invalid geom type but changed not to so (create visual geometry) because, technically, we still can create one as long as `<Render>` element is provided.

src/util/KinBodyParser.cpp
jslee02: I thought we don't want to create visual geometry for invalid geom type but changed not to so (create visual geometry) because, technically, we still can create one as long as `<Render>` element is provided.

jslee02:
I believe I've addressed all the @mkoval's comments. @mkoval could you do another round of review?

include/aikido/util/KinBodyParser.hpp
```diff
@@ -0,0 +1,83 @@
...
+
+namespace aikido {
+namespace util {
+
+/// Read skeleton from a string of OpenRAVE's custom XML format
+///
+/// This function only parses a subset of the format assuming only one body node
+/// in a kinbody file.
+///
+/// Followings are the available fields that this parser can read.
```
mkoval: Nit: `Followings` -> `The following`.

include/aikido/util/KinBodyParser.hpp
```diff
@@ -0,0 +1,83 @@
...
+/// kinbody - attributes: name
+///   body - attributes: name
+///     geom - attributes: name, type* (none, box, sphere, trimesh, cylinder), render
+///       data* (or collision) - file* scale [for trimesh] (see below for the detail)
+///       extents* - 3 float [for box]
+///       height* - float [for cylinder]
+///       radius* - float [for cylinder and sphere]
+///       render - file* scale (see below for the detail)
+/// \endcode
+/// Elements marked with `*` are required ones.
```
mkoval: Nit: "are required ones" -> "are required".

include/aikido/util/KinBodyParser.hpp
```diff
@@ -0,0 +1,83 @@
...
+///
+/// <render>, <data>, or <collision> contains the relative path to a mesh file
+/// and optionally a single float (for the uni-scale) or three float's (for the
+/// x, y, and z-axes) for the scale.
```
mkoval: Nits:
- "contains" -> "contain"
- "uni-scale" -> "for all three axes" (I thought the term uni-scale was a bit confusing, but it's fine to keep this if you

include/aikido/util/KinBodyParser.hpp
```diff
@@ -0,0 +1,83 @@
...
+///
+/// Example forms:
+///   <Render>my/mesh/file.stl<Render>
+///   <Render>my/mesh/file.stl 0.25<Render> <!--Unscale>
```
mkoval: Nits:
- `Unscale` typo? Also, see my comment above.
- Incorrect closing comment `-->`

include/aikido/util/string.hpp
```diff
@@ -0,0 +1,30 @@
...
+/// Example:
+/// \code
+/// std::string text = "HERB, the robot.";
+///
+/// auto res1 = split(text); // {"HERB,", "the", "robot."}
+/// auto res2 = split(text, ","); // {"HERB", " the robot."}
+/// \endcode
+///
+/// \param[in] string Input string to be splitted.
+/// \param[in] delimiters Delimiters.
```
mkoval: Add a note that *any* character this string is considered a delimiter, i.e. the string is interpreted as a set of delimiter characters, not as a single multi-character delimiter.

include/aikido/util/string.hpp
```diff
@@ -0,0 +1,30 @@
...
+/// \return The splitted substrings.
+std::vector<std::string> split(
+    const std::string& string, const std::string& delimiters = " ");
```
mkoval: Add `\t` as a default delimiter as well?

test_resources/kinbody/objects/bowl.kinbody.xml
```diff
@@ -0,0 +1,16 @@
+<KinBody name="bowl">
+  <Body type="static" name="bowl">
+    <Geom type="trimesh">
+      <!--<Data>bowl.iv</Data>-->
```
mkoval: Nit: Remove comment.

test_resources/kinbody/objects/kinova_tool.kinbody.xml
```diff
@@ -0,0 +1,17 @@
...
+      <diffuseColor>0.0 0.0 0.0</diffuseColor>
+      <ambientColor>0.0 0.0 0.0</ambientColor>
+      <specularColor>0.0 0.0 0.0</specularColor>
+      <shininess>0.2</shininess>
+      <transparency>0.0</transparency>
+    </Geom>      
+    <Geom type="trimesh" modifiable="false" render="false">
+        <Data>kinova_tool.stl 0.1</Data>
+    </Geom>
+    <!-- TODO: Add a sphere approximation. -->
```
mkoval: Nit: Remove comment.

test_resources/kinbody/objects/smallsphere.kinbody.xml
```diff
@@ -0,0 +1,11 @@
...
+
+<KinBody>
+	<Body type="dynamic">
+		<Geom type="sphere">
+			<Translation>0  0  0</Translation>
+			<diffuseColor>0 0.5 0</diffuseColor>
+			<radius>0.013</radius>
+		</Geom>
+	</Body>
+</KinBody>
```
mkoval: Nit: Change this file to use the same spacing as the other `.kinbody.xml` files.

PR #105, opened 2016-06-28
Workaround for VisualAspect default color bug

mkoval:
This is a workaround for https://github.com/dartsim/dart/issues/739 pending an upstream fix.

---

This change is [Reviewable](https://reviewable.io/reviews/personalrobotics/aikido/105)


psigen:
@mkoval: what were your thoughts on https://github.com/dartsim/dart/issues/739#issuecomment-229189512

mkoval:
I will merge this workaround for now, since most of our visual meshes (e.g. HERB) are textured. We can revisit the topic in the future to come up with a more convincing fix.

PR #106, opened 2016-06-29
Added helper functions for printing CollisionResult

mkoval:
I found this to be very useful when debugging spurious collisions, e.g. due to missing `<disable_collision>` pairs in SRDF.

---

This change is [Reviewable](https://reviewable.io/reviews/personalrobotics/aikido/106)


aikido/src/util/CMakeLists.txt
```diff
@@ -4,6 +4,7 @@ aikido_add_library("${PROJECT_NAME}_util"
   VanDerCorput.cpp
   StepSequence.cpp
   RNG.cpp
+  stream.cpp
```
psigen: I don't know how I feel about other cpp source files being camel case because they define classes and this one being all lowercase.  Is there a convention for this?

aikido/include/aikido/util/stream.hpp
```diff
@@ -0,0 +1,38 @@
...
+
+/// Operator that calls printCollisionObject.
+std::ostream& operator <<(std::ostream& stream,
+  const dart::collision::CollisionObject& collisionObject);
+
+/// Operator that calls printCollisionReport.
+std::ostream& operator <<(std::ostream& stream,
+  const dart::collision::CollisionResult& collisionResult);
+
+} // namespace operators
```
psigen: Is the assumption here that users would do something like:
```c++
using aikido::util::operators::operator<<;
```
to activate these stream overloads?

mkoval: I'm not sure if that will work since `operator <<` is overloaded. I typically use:
```c++
using namespace aikido::util::operators
```
Putting the operators in their own namespace avoids the namespace pollution that comes along with `using namespace`.

aikido/src/util/CMakeLists.txt
mkoval: I thought naming it `Stream.hpp` would imply that the file contains a class named `Stream`. However, we already break this rule in `PseudoInverse.hpp`, so that ship has already sailed. 😁

jslee02:
I made some mistakes in merging master into this PR. Please do "Squash and merge" (as we usually do) when someone merges this PR for the cleaner history.

PR #113, opened 2016-11-21
Support DART 6.1

jslee02:
This change is [Reviewable](https://reviewable.io/reviews/personalrobotics/aikido/113)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -102,7 +102,7 @@ find_package(Boost COMPONENTS filesystem)
 include_directories(${Boost_INCLUDE_DIRS})
 
 # TODO: utils-urdf should only be required for aikido::perception.
-find_package(DART REQUIRED COMPONENTS optimizer-nlopt utils-urdf CONFIG)
+find_package(DART 6.1.0 REQUIRED COMPONENTS optimizer-nlopt utils-urdf CONFIG)
 include_directories(${DART_INCLUDE_DIRS})
 
 #==============================================================================
--- a/src/constraint/TSR.cpp
+++ b/src/constraint/TSR.cpp
@@ -72,6 +72,9 @@ private:
 
 
   friend class TSR;
+
+public:
+  EIGEN_MAKE_ALIGNED_OPERATOR_NEW
 };
 
 //=============================================================================
--- a/src/statespace/dart/SO2Joint.cpp
+++ b/src/statespace/dart/SO2Joint.cpp
@@ -5,8 +5,7 @@ namespace statespace {
 namespace dart {
 
 //=============================================================================
-SO2Joint::SO2Joint(
-      ::dart::dynamics::SingleDofJoint* _joint)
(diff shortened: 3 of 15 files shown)


CMakeLists.txt
```diff
@@ -102,7 +102,7 @@ find_package(Boost COMPONENTS filesystem)
 include_directories(${Boost_INCLUDE_DIRS})
 
 # TODO: utils-urdf should only be required for aikido::perception.
-find_package(DART REQUIRED COMPONENTS optimizer-nlopt utils-urdf CONFIG)
+find_package(DART 6.1.0 REQUIRED COMPONENTS optimizer-nlopt utils-urdf CONFIG)
```
mkoval: I think this will break with *any* release other than DART 6.1.0. Since we only break backwards compatibility on major releases, I think we're better off `find_package(DART ...)` and then checking that `DART_VERSION` is greater than or equal to `6.1.0` and less than `7.0.0`. Thoughts?

src/constraint/TSR.cpp
```diff
@@ -72,6 +72,9 @@ class TSRSampleGenerator : public SampleGenerator
 
 
   friend class TSR;
+
+public:
+  EIGEN_MAKE_ALIGNED_OPERATOR_NEW
```
mkoval: 👍 Good catch!

src/statespace/dart/SO2Joint.cpp
```diff
@@ -5,8 +5,7 @@ namespace statespace {
 namespace dart {
 
 //=============================================================================
-SO2Joint::SO2Joint(
-      ::dart::dynamics::SingleDofJoint* _joint)
+SO2Joint::SO2Joint(::dart::dynamics::GenericJoint<::dart::math::R1Space> *_joint)
```
mkoval: `SO2` is a different topology than `R1`. Is this mismatch intentional?

CMakeLists.txt
jslee02: I believe the decision of compatibility is determined by the upstream version check script. In particular, the compatibility policy of DART is specified as [_same major version_](https://github.com/dartsim/dart/blob/master/CMakeLists.txt#L363). So this should work for DART 6.1.0 or any greater.

src/statespace/dart/SO2Joint.cpp
jslee02: It's intentional for now. DART currently doesn't distinct R and SO(2) using different classes anyways. Instead, we have used a member function that says if the joint configuration space is SO(2) by checking the position limits. I'm thinking of introducing real SO(2) configuration space in the future, but not in DART 6.1.

jslee02:
This PR got an approval, but I don't have an access to merge PRs of aikido. I would appreciate if someone who has the write access merges this.

psigen:
@mkoval: good to merge?

CMakeLists.txt
mkoval: I didn't know this was an option! That's great. 👍

mkoval:
Merged! Thanks @jslee02 for taking the time to do this.

Issue #53, opened 2016-04-20
Configure Travis to build with -Werror


ClintLiddick:
We should have no build warnings

jslee02:
I've thought to do the same thing for DART. I should create the same issue now.

psigen:
Go team!
https://github.com/dartsim/dart/issues/677

jslee02:
Resolved by #114

PR #114, opened 2016-11-22
Configure Travis to build treating warnings as errors

jslee02:
This PR addresses #53. 

Depends on: #113 (merged)

---
This change is [Reviewable](https://reviewable.io/reviews/personalrobotics/aikido/114)


jslee02:
@mkoval Do we need to add `-DTREAT_WARNINGS_AS_ERRORS=ON` to [here](https://github.com/personalrobotics/aikido/blob/master/.travis.yml#L24) as well?

CMakeLists.txt
```diff
@@ -102,10 +111,68 @@ find_package(Boost COMPONENTS filesystem)
...
+  endif()
+
+elseif("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
+
+  if(CMAKE_CXX_COMPILER_VERSION VERSION_LESS 3.3)
+    message(FATAL_ERROR "The installed Clang version is
+        ${CMAKE_CXX_COMPILER_VERSION}. ${PROJECT_NAME} requires clang 3.3 or
+        greater."
+    )
+  endif()
```
mkoval: Please add a comment about why we require this version.

CMakeLists.txt
```diff
@@ -102,10 +111,68 @@ find_package(Boost COMPONENTS filesystem)
...
+
+elseif("${CMAKE_CXX_COMPILER_ID}" STREQUAL "AppleClang")
+
+  if(CMAKE_CXX_COMPILER_VERSION VERSION_LESS 6.1)
+    message(FATAL_ERROR "The installed AppleClang version is
+        ${CMAKE_CXX_COMPILER_VERSION}. ${PROJECT_NAME} requires AppleClang 6.1
+        or greater."
+    )
+  endif()
```
mkoval: Please add a comment about why we require this version.

CMakeLists.txt
```diff
@@ -102,10 +111,68 @@ find_package(Boost COMPONENTS filesystem)
...
+
+  if(TREAT_WARNINGS_AS_ERRORS)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Werror")
+  endif()
+
+elseif(MSVC)
+
+  if(MSVC_VERSION VERSION_LESS 1900)
+     message(FATAL_ERROR "${PROJECT_NAME} requires VS 2015 or greater.")
```
mkoval: Please add a comment about why we require this version.

CMakeLists.txt
```diff
@@ -102,10 +111,68 @@ find_package(Boost COMPONENTS filesystem)
...
+   endif()
+
+  if(TREAT_WARNINGS_AS_ERRORS)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /WX")
+  endif()
+
+else()
+
+  message(SEND_ERROR "Compiler[${CMAKE_CXX_COMPILER_ID}] not supported.")
```
mkoval: Is it necessary to make this a fatal error? I don't see why we couldn't continue the build without enabling warnings as errors.

I'd also prefer to put all of the logic to set `-Werror` into a helper file, e.g. `WarningsAsErrors.cmake`.

mkoval:
> @mkoval Do we need to add `-DTREAT_WARNINGS_AS_ERRORS=ON` to here as well?

I don't have a strong opinion either way. If the unit tests currently build without warnings (or it is trivial to make them do so), then I would go ahead and add the flag. Otherwise, it doesn't seem worth the effort.

CMakeLists.txt
jslee02: > Is it necessary to make this a fatal error? I don't see why we couldn't continue the build without enabling warnings as errors.

It has nothing to do the option, but for compilers that we don't support. I thought we support only GCC, clang, and MSVC. If it's not the case, we can remove this error.

> I'd also prefer to put all of the logic to set -Werror into a helper file, e.g. WarningsAsErrors.cmake.

:+1:

CMakeLists.txt
jslee02: All the compiler requirements are to ensure that C++ is supported by them. I'll add the comment.

jslee02:
I'm not familiar with the build process with catkin. My preference is enabling the option for every build for sure hoping that it doesn't require to change the build script a lot. What is the most effective way to do so?

CMakeLists.txt
jslee02: I actually prefer to have all the compiler settings in a helper file like `CompilerSettings.cmake` as it will be a big block of script. What do you think?

mkoval:
The [`internal-build.sh`](https://github.com/personalrobotics/pr-cleanroom/blob/master/internal-build.sh#L5) script uses the `CATKIN_CONFIG_OPTIONS` environment variable to populate the default build flags. You could add this to the `.travis.yml` file:
```yaml
env:
  global:
  - "CATKIN_CONFIG_OPTIONS=-DCMAKE_BUILD_TYPE=Release -DTREAT_WARNINGS_AS_ERRORS=ON"
```

I added `-DCMAKE_BUILD_TYPE=Release` because that is the default if no other flags are specified. This should be overridden (as expected) when we run tests:
```bash
./scripts/internal-run.sh catkin build --no-status --no-deps -p 1 -i --cmake-args -DCMAKE_BUILD_TYPE=Debug -DCOVERALLS=ON -- --make-args all tests -- aikido
```

mkoval:
> All the compiler requirements are to ensure that C++ is supported by them. I'll add the comment.

Do you mean C++11? If so, can we put this into a separate that *just* sets the flags necessary to enable C++11 support?

jslee02:
Hm, this PR has become to have several changes:

(1) Compiler version checking
(2) Adding a compiler flag for enabling C++11
(3) Passing the compiler flag for enabling C++11 to the consumer libraries using `target_compiler_options()`
(4) Adding a compiler flag for treating warnings as errors

> Do you mean C++11? If so, can we put this into a separate that just sets the flags necessary to enable C++11 support?

I think we need both of (1) and (2) because old compilers wouldn't be able to enable C++11 even though we set a flag (e.g., `-std=c++11`) for it. 

> You could add this to the .travis.yml file

I'll do this.

> This should be overridden (as expected) when we build the tests for Coveralls support:

Are the whole default options are overridden or just options that are specified in both places?

mkoval:
> Are the whole default options are overridden or just options that are specified in both places?
View changes

Overriden. The options are merged into one list of command-line arguments that are passed to `cmake`. If you set `-DOPTION=1` using `catkin config` (i.e. through the `CATKIN_CONFIG_OPTIONS` environment variable) and pass `-DOPTION=2` to `catkin build`, then `cmake` is invoked with `-DOPTION=1 -DOPTION=2`.  CMake uses the last value, so the end result should be equivalent to passing `-DOPTION=2`.

> I think we need both of (1) and (2) because old compilers wouldn't be able to enable C++11 even though we set the flag (e.g., `-std=c++11`) for it.

I understand that. I am advocating for encapsulating the "check if the compiler supports C++11 and, if so, enable it" logic into a standalone `.cmake` file that we can easily remove later. Note that CMake 3.1, which we can hopefully switch to soon, [adds native support for C++11](https://crascit.com/2015/03/28/enabling-cxx11-in-cmake/) by setting:
```cmake
set(CMAKE_CXX_STANDARD 11)
set(CMAKE_CXX_STANDARD_REQUIRED ON)
```

mkoval:
I'd like to get this merged, but I am concerned that we are adding the flags to each target with `PUBLIC` visibility.

First, my understanding is that `PUBLIC` flags will be attached to the `EXPORT`ed targets that are `include`d in the auto-generated `aikidoConfig.cmake` file. This is troubling because it means that users who link against a version of Aikido that was built with `-DTREAT_WARNINGS_AS_ERRORS=ON` will inadvertently enable enable `-Werror` on their own targets.

Second, it adds a lot of boilerplate to Aikido's [already quite verbose] `CMakeLists.txt` files.

What is the advantage of this over setting `CMAKE_CXX_FLAGS` inside Aikido's top-level `CMakeLists.txt` file?

jslee02:
The first concern would be fine because this PR doesn't add `-DTREAT_WARNINGS_AS_ERRORS=ON` but the flag to the Aikido targets that enables building with C++11 (depending on the compilers) so that the consumer project inherently enable C++11.

For the second point, as mentioned above, it passes the compiler options to the consumer projects so that the consumer project doesn't need to figure out what are the required flags such as [this](https://github.com/personalrobotics/dartpy/blob/master/CMakeLists.txt#L4).

mkoval:
> The first concern would be fine because this PR doesn't add `-DTREAT_WARNINGS_AS_ERRORS=ON` but the flag to the Aikido targets that enables building with C++11 (depending on the compilers) so that the consumer project inherently enable C++11.

It looks like this pull request *does* add the `TREAT_WARNINGS_AS_ERRORS` CMake `option` to Aikido. What am I missing here?

In any case, I am fine with having that option. I am only concerned that the `-Werror` flag is set as a `PUBLIC` property on Aikido's targets.

> For the second point, as mentioned above, it passes the compiler options to the consumer projects so that the consumer project doesn't need to figure out what are the required flags such as this.

This makes sense for flags that are *required* for a library to work as expected, e.g. `--std=c++11` or `-frounding-math`. I think it's a bad idea modify compiler warnings and errors for dependent projects, unless doing so is absolutely necessary to `#include` Aikido's headers. Adding `-Werror` to a dependency project could break that project's build!

jslee02:
We're adding two flags `-Werror` and `-std=c++11` to Aikido, but _only_ `-std=c++11` is added to dependency projects. Please note that `-Werror` is added using `**add**_compile_options()` while `-std=c++11` is added using `**target**_compile_options()`.

jslee02:
No worries. Thanks!

Issue #21, opened 2016-03-10
Set up automated style-checking for c++ code.


psigen:
It can be really hard to validate style conventions when reviewing PRs.  Let's try to set up automated style checking for `aikido` before the codebase gets too large.

Some promising options:
- https://github.com/alexmurray/cstyle
- http://astyle.sourceforge.net/
- https://bitbucket.org/verateam/vera/wiki/Home
- https://github.com/kunaltyagi/nsiqcppstyle

jslee02:
Resolved by #116

PR #116, opened 2016-11-22
Setup clang-format for automatic code formatting

jslee02:
Related issue: #21

This change is [Reviewable](https://reviewable.io/reviews/personalrobotics/aikido/116)


mkoval:
I'm generally skeptical of automatic formatting tools. They often make questionable decisions on edge cases that make the code less readable. I'd prefer a linter that *checks* the formatting of pull requests, but does not change it.

Pinging @psigen here to see what he thinks.

jslee02:
> They often make questionable decisions on edge cases that make the code less readable.

I basically agree with you. This PR adds two targets one for formatting and the other for checking the format. We could use the first one as a guideline, and then we could discard it if it doesn't make sense.

One possible issue would be that the checking will keep complaining about the discarded change that we don't think it makes sense.

psigen:
I am in agreement with @mkoval here, we should automatically perform *lint* checks, but not do any automated formatting unless specifically called.

Automated formatting might be useful as a one time operation if we pull in large chunk of code.  We can autoformat it, then correct the formatting to make it more readable.

If I'm reading this correctly, this adds two non-default targets for linting and formatting, neither of which is directly linked to anything yet. It seems like we would use the linting target for automated testing or pre-commit, and largely ignore the formatting unless called by a developer locally, right?

jslee02:
> If I'm reading this correctly, this adds two non-default targets for linting and formatting, neither of which is directly tied to anything yet. It seems like we would use the linting target in our automated testing or as a pre-commit hook, and largely ignore the formatting target except as a convenience method for a developer locally.

Exactly!

mkoval:
That sounds perfect. Nice work @jslee02!

ClintLiddick:
I'm going to voice my disagreement here. I think all code should be autoformatted, otherwise people will forget, someone else will do it, and diff-hell will emerge. If there is a block of code that doesn't look right with the style configuration, you should manually  wrap it with a `// clang-format off` block so that it blocks warnings in the future, and it shows others the formatting is intentional to improve readability/your intention.

That said, I'm not sure if we want to add this to CI, or just yell at people to do it during PR reviews. Thoughts?

psigen:
The issue is that autoformatting is _underspecified_.  No autoformatter knows anything about where to put line breaks to semantically make sense, or how to group nested arrays.  There can be many ways to format code that are _all_ compliant with the style specification.

> I think all code should be autoformatted, otherwise people will forget, someone else will do it, and diff-hell will emerge.
I don't see how this is an issue:  the autoformatter by definition doesn't care who is running it.

Is there some misunderstanding here?  The proposal here is that code style checks are like any other unit test:  failing them means that a PR is automatically flagged as not ready for a merge.  How would this cause any of the issues you are mentioning?

psigen:
Is the `lab style format` the same as `DART style format`?  Just wondering.

jslee02:
> Is the lab clang-format mostly the same as DART clang-format? Just wondering.

This [diff](https://github.com/personalrobotics/aikido/pull/116/commits/ea4fe8776cf859f4da65f67dd3f160c6923b0338) shows the differences between the style configurations. It seems the key configurations I believe are pretty similar, but as you can see there are some differences. By the way, I wonder if the lab clang-format is made upon some consensus.

psigen:
I'm not sure, some of the differences in that diff seem to go against what I would have expected.  @ClintLiddick do you have some insight into this?

ClintLiddick:
@psigen I think the differences were mostly me going back through some of our code and seeing what style we (mostly @mkoval and I?) used and what I know to be common practice. Definitely open to specific suggestions/changes.

Yeah I guess I misunderstood the intent then. If the autoformatter is run as a check then I'd be ok. I just don't want a situation where I autoformat and because the last person to commit didn't I now have a tragic diff. I don't think there's going to be much difference from what I suggested however, because in order to pass the test you will still have to either accept the autoformatting, or wrap with an ignore block comment.

psigen:
To be honest, I'm generally against using any automatic formatting vs. automatic style checking.  It leads to automatically creating stuff like this for no reason: https://github.com/personalrobotics/herbpy/blob/feature/linting/scripts/calibration_transmission_ratios.py#L184 which is really not improving the readability of the code.

If we have an automated style test, then it should not be possible to add code to the repo that does not comply with the style guide.  If you run an auto-formatter in your PR and it destroys the formatting of some unrelated code _that was already compliant with the style guide_, that seems like an error in your PR, not the existing codebase.  If it was _not_ compliant, then it should not have passed the check to be accepted in the first place.

I see no reason to force an auto-formatter to be run at any point in time in the PR process.  I do not trust them to produce readable or semantic code, and they are not in themselves a test that can be passed.

ClintLiddick:
I'm not aware of any tool that is a style checker and not an auto formatter. Or any tool that is flexible enough to case-by-case change rules to make some code you like "pass." Either your tool is the style guide, and you have some exception comments for special cases, or you will never get a clean run of the style checker. If the tool is the style guide then there is no difference from running it automatically and fixing anything that looks really horrible with an escape block. I'm not saying we have to run it automatically, but I am saying there is no point in having it if it's not equivalent anyway.

Unless there's a tool/setup I'm unaware of that is far more powerful than clang-format for checking style on a per-change level (not a codebase level)?

[Not trying to be argumentative, but just literally don't understand the practical difference here.]

psigen:
No worries, I think we are mostly on the same page here.

Tools like [cpplint](https://github.com/google/styleguide/tree/gh-pages/cpplint), [cppcheck](https://sourceforge.net/projects/cppcheck/), and [Coverity](https://www.synopsys.com/software-integrity/products/static-code-analysis.html#) are examples of _checkers/linters_ that detect potential issues, but do not attempt to automatically fix them.  **clang-format** can either _check_ for errors, or _rewrite_ code to try and comply with a style guide, and does a pretty good job of it, although we may still use another tool for other static analysis checks in the future.

Suppose you open a PR with a piece of non-compliant code like a matrix that exceeds 80 characters:
```cpp
// Block A
const int[10][4] my_matrix = {{0, 1, 2, 3, 4, 5, 6, 7, 8, 9}, {0, 1, 2, 3, 4, 5, 6, 7, 8, 9}, {0, 1, 2, 3, 4, 5, 6, 7, 8, 9}, {0, 1, 2, 3, 4, 5, 6, 7, 8, 9}};
```

It makes a ton of sense that this PR should fail automated tests with something like:
```bash
>> Line 1: Failed style check. Too many characters in this line (Max: 80)
"const int[10][4] my_matrix = {{0, 1, 2, 3, 4, 5, 6, 7, 8, 9}, {0, 1, 2, 3, 4, 5, 6, 7, 8, 9}, {0, 1, 2, 3, 4, 5, 6, 7, 8, 9}, {0, 1, 2, 3, 4, 5, 6, 7, 8, 9}};"
```

Then we have flagged it just like any other unit test error, and the author can adjust the PR to look like something reasonable:
```cpp
// Block B
const int[10][4] my_matrix = {
  {0, 1, 2, 3, 4, 5, 6, 7, 8, 9},
  {0, 1, 2, 3, 4, 5, 6, 7, 8, 9},
  {0, 1, 2, 3, 4, 5, 6, 7, 8, 9},
  {0, 1, 2, 3, 4, 5, 6, 7, 8, 9}
};
```

What I am arguing is _not_ ideal is forcing us to run an automatic formatter on this code which might easily produce something like this:
```cpp
// Block C
const int[10][4] my_matrix = {
  {
    0,
    1,
    2,
    3,
    4,
    5,
    6,
    7,
    8,
    9
  },
  {
    0,
    1,
    2,
    3,
    4,
    5,
    6,
    7,
    8,
    9
  },
  {
    0,
    1,
    2,
    3,
    4,
    5,
    6,
    7,
    8,
    9
  },
  {
    0,
    1,
    2,
    3,
    4,
    5,
    6,
    7,
    8,
    9
  }
};
```

While this particular line might not do something this pathological in `clang-format`, this type of problem is rampant in automatic formatting.  **Block B** is _just as compliant_ to the style guide as **Block C** but is much more readable.

The auto-formatter has no way to consider these options, it is simply applying _some_ whitespace that it knows is compliant with the style guide, but not the _only_ or _best_ whitespace solution.

The second part of this is that we don't want to implement any process that takes **Block B**, which is a _compliant_ piece of code, and turns it into **Block C**, a _compliant but less readable_ version of the same code, as part of the process of a later PR.  Some autoformatters will happily convert **Block B** to **Block C** if run over the entire codebase, and I would argue this is _wrong_ behavior, even if it is _easier_ behavior.

Hopefully that makes more sense.

jslee02:
By unnecessary whitespace, you mean the empty lines like
```cmake
if (CLANG_FORMAT_EXECUTABLE)
  # here?
  get_property(formatting_files GLOBAL PROPERTY CODE_FORMATTING_SOURCES)
  list(LENGTH formatting_files formatting_files_length)
  
  if (formatting_files)
  # here?
    add_custom_target(format
        COMMAND ${CMAKE_COMMAND} -E echo "Formatting code style of"
            ...
```
?

CMakeLists.txt
```diff
@@ -235,6 +258,56 @@ if(COVERALLS)
...
+        COMMAND ${CMAKE_SOURCE_DIR}/tools/check_code_format.sh
+            ${CLANG_FORMAT_EXECUTABLE} ${formatting_files}
+        COMMAND ${CMAKE_COMMAND} -E echo "Done."
+        DEPENDS ${CLANG_FORMAT_EXECUTABLE}
+        WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
+    )
+
+  else()
+
+    add_custom_target(code-formatting
```
psigen: Why is this target named `code-formatting` but the target in the other branch of this `if-else` block called `format`?

jslee02: That's just an oversight. Will fix it!

psigen:
@mkoval: Yup, the issues I raised weren't actually in this PR, it's all good.

mkoval:
@jslee02 Yes, I meant the empty lines that you identified.

jslee02:
Unnecessary empty lines are removed in the last commit.

PR #121, opened 2016-12-27
Add custom target for displaying the headers in IDEs

jslee02:
Some IDEs including Qt Creator doesn't show the headers when they are not in a build target. This PR adds a dummy custom target that contains the headers for displaying them in IDEs.

---
This change is [Reviewable](https://reviewable.io/reviews/personalrobotics/aikido/121)


jslee02:
I believe we can merge this if there is no objection.

PR #133, opened 2017-01-26
fix missing argument to VanDerCorput constructor

ClintLiddick:
This compilation error wasn't caught by Travis CI, so I suspect a silent exclusion of the planner_ompl module.

---
This change is [Reviewable](https://reviewable.io/reviews/personalrobotics/aikido/133)


ClintLiddick:
Correction. planner_ompl is being built, but only with GCC. clang-3.8 caught this error.

PR #135, opened 2017-02-10
Adds SE2BoxConstraint

gilwoolee:
This PR adds:
- `SE2BoxConstraint` which is `Testable`, `Sampleable`, and `Projectable`.
- helper methods in `JoinStateSpaceHelpers-Impl.hpp` so that dart's SE2 joints with x, y constraints can use be tested, sampled, and projected. 
- SE2 distance metric `distance::SE2` which takes l2 norm of (dx, dy, dw). (Happy to take other suggestions too.)

To be consistent with SO2 and SO3 samplers, this does not allow limit on rotational joint, but only on x, y dimensions.

---
This change is [Reviewable](https://reviewable.io/reviews/personalrobotics/aikido/135)


mkoval:
Why did you close this? These look like a great set of changes.

gilwoolee:
I realized that I wanted to make some changes, e.g. making the distance
function weighted sum of radial distance and R^2 norm , and add at least
some basic tests. Not sure I could do it in the next a few days (given IROS
deadline), so I'll re open it as soon as its ready!

Thanks for noticing.

- Gilwoo

gilwoolee:
This is not demo-priority but we'll need it eventually for base planning. I haven't made any changes I promised, but it'd be great if someone could take a quick pass and comment/agree on the general structure.

src/constraint/uniform/SE2BoxConstraint.cpp
```diff
@@ -0,0 +1,213 @@
...
+      std::shared_ptr<statespace::SE2> _space,
+      std::unique_ptr<util::RNG> _rng,
+      const Eigen::Vector2d& _lowerLimits,
+      const Eigen::Vector2d& _upperLimits)
+  : mSpace(std::move(_space))
+  , mRng(std::move(_rng))
+  , mRnDimension(2)
+  , mDimension(3)
+{
+  mLowerLimits[0] = -M_PI;
```
brianhou: Nit: can we explicitly document that the limits are `[theta, x, y]`? It seems that this is a result of `SE2::logMap`.

src/distance/SE2.cpp
```diff
@@ -0,0 +1,50 @@
...
+{
+  return mStateSpace;
+}
+
+//=============================================================================
+double SE2::distance(
+    const aikido::statespace::StateSpace::State* _state1,
+    const aikido::statespace::StateSpace::State* _state2) const
+{
+  Eigen::VectorXd tangent1;
```
brianhou: Should these be `Vector3d`?

jslee02:
> It'd be great if you have time to add the weighted norm and tests, but it might be easier to address it in a future PR and merge this now!

:+1: If we merge this without those great changes, then it would be less great to add issues to track this issues. :smile:

src/distance/SE2.cpp
jslee02: `VectorXd` is correct here. `logMap`, which is a virtual function of `StateSpace`, takes `VectorXd` and resize it if needed.

src/constraint/uniform/SE2BoxConstraint.cpp
jslee02: :+1: to adding documentation for this.

Also, I wonder why we don't take the bounds for the rotation angle of SE(2) here. @gilwoolee Do we have a reason not to do so?

include/aikido/constraint/uniform/SE2BoxConstraint.hpp
```diff
@@ -65,7 +65,10 @@ class SE2BoxConstraint
   /// and the last two elements encode the translational limits.
   Eigen::Vector3d mUpperLimits;
 
+  // DOF of joint that have limits, in this case translational DOF.
+  // TODO: Confirm this with Gilwoo
   size_t mRnDimension;
```
aditya-vk: @gilwoolee : is the docstring for this variable right?

tests/planner/ompl/test_MotionValidator.cpp
```diff
@@ -29,7 +29,7 @@ class MotionValidatorTest : public ::testing::Test
         std::make_shared<MockTranslationalRobotConstraint>(
             stateSpace, Eigen::Vector3d(-0.1, -0.1, -0.1),
             Eigen::Vector3d(0.1, 0.1, 0.1));
-    auto vchecker =
+    ::ompl::base::StateValidityCheckerPtr vchecker =
```
aditya-vk: @jslee02 not specifying explicitly the type was resulting in a build error due to ambiguity of the variable type, between `StateValidityCheckerPtr` and `StateValidityCheckerFn`. I thought you might want to confirm this change.

aditya-vk:
> If we merge this without those great changes, then it would be less great to add issues to track this issues. 

Introduced the weighted norm and added tests. Please do review. Thanks!

src/distance/SE2Weighted.cpp
```diff
@@ -0,0 +1,73 @@
...
+  Eigen::Vector2d linearDistance;
+  double angularDistance;
+
+  // Difference between angles
+  angularDistance = tangent2[0] - tangent1[0];
+  angularDistance = std::fmod(std::abs(angularDistance), 2.0 * M_PI);
+  if (angularDistance > M_PI)
+  {
+    angularDistance -= 2.0 * M_PI;
+    angularDistance = -angularDistance;
```
jslee02: @aditya-vk I added this line to make `angularDistance` to be always positive. Previously, `angluarDistance` could be negative if we get in this if-statement.

jslee02:
Unrelated issue: It seems [the Travis tests on ubuntu isn't working](https://travis-ci.org/personalrobotics/aikido/jobs/277028954#L1683) [due to `http://packages.personalrobotics.ri.cmu.edu` is down](https://travis-ci.org/personalrobotics/aikido/jobs/277028954#L1596). Let me resolve this by switching to using PPA instead of running our own debian package server, as discussed.

src/distance/SE2Weighted.cpp
```diff
@@ -58,10 +58,8 @@ double SE2Weighted::distance(
   angularDistance = tangent2[0] - tangent1[0];
   angularDistance = std::fmod(std::abs(angularDistance), 2.0 * M_PI);
   if (angularDistance > M_PI)
-  {
-    angularDistance -= 2.0 * M_PI;
-    angularDistance = -angularDistance;
-  }
+    angularDistance = 2*M_PI - angularDistance;
```
aditya-vk: @jslee02 thanks for the change! Simplified it into a single statement.

tests/constraint/test_SE2BoxConstraint.cpp
```diff
@@ -24,7 +24,7 @@ class SE2BoxConstraintTests : public ::testing::Test
   static constexpr size_t NUM_X_TARGETS { 10 };
   static constexpr size_t NUM_Y_TARGETS { 10 };
   static constexpr size_t NUM_SAMPLES { 1000 };
-  static constexpr double DISTANCE_THRESHOLD { 0.15 };
+  static constexpr double DISTANCE_THRESHOLD { 0.8 };
```
aditya-vk: @jslee02  For the number of samples taken and the resolution, the maximum distance between a sample and a target in this case is ~0.74. increasing the threshold to 0.8 is sensible in my opinion.

aditya-vk:
@jslee02 I agree, the Weighted class needs to be renamed. Pushed the relevant changes and issued a warning message suggesting the new class, in the `Weighted.hpp` file. Thanks!

include/aikido/distance/Weighted.hpp
```diff
@@ -1,58 +1,8 @@
...
-      const statespace::StateSpace::State* _state2) const override;
-
-private:
-  std::shared_ptr<statespace::CartesianProduct> mStateSpace;
-  std::vector<std::pair<DistanceMetricPtr, double>> mMetrics;
-};
-
-} // namespace distance
-} // namespace aikido
+#warning Weighted is deprecated. Please use CartesianProductWeighted instead by including "CartesianProductWeighted".
```
jslee02: Please add `using Weighted = CartesianProductWeighted`.

src/constraint/uniform/SE2BoxConstraint.cpp
brianhou: @jslee02 was this documented somewhere? I don't seem to see it.

brianhou: Never mind, found it!

PR #136, opened 2017-03-10
Addes TSR visualizer to aikido/rviz.

gilwoolee:
This change is [Reviewable](https://reviewable.io/reviews/personalrobotics/aikido/136)

### DIFF
--- a/include/aikido/rviz/InteractiveMarkerViewer.hpp
+++ b/include/aikido/rviz/InteractiveMarkerViewer.hpp
@@ -9,6 +9,7 @@
 #include <interactive_markers/interactive_marker_server.h>
 #include "SmartPointers.hpp"
 #include <aikido/constraint/TSR.hpp>
+#include "TSRMarker.hpp"
 
 namespace aikido {
 namespace rviz {
@@ -34,8 +35,8 @@ public:
     dart::dynamics::SkeletonPtr const &skeleton);
 
   /// Visualizes tsr with at most n samples
-  void visualizeTSR(const aikido::constraint::TSR& _tsr, int nSamples = 10);
-  void cleanTSR();
+  TSRMarkerPtr visualizeTSR(aikido::constraint::TSR const &_tsr, int nSamples = 10,
+    std::string basename = "");
 
   void setAutoUpdate(bool _flag);
   void update();
@@ -47,7 +48,6 @@ private:
   interactive_markers::InteractiveMarkerServer mMarkerServer;
   std::set<SkeletonMarkerPtr> mSkeletonMarkers;
   std::set<FrameMarkerPtr> mFrameMarkers;
-  std::vector<std::shared_ptr<dart::dynamics::SimpleFrame>> mSimpleFrames;
 
   std::atomic_bool mRunning;
   std::atomic_bool mUpdating;
--- a/include/aikido/rviz/InteractiveMarkerViewer.hpp
+++ b/include/aikido/rviz/InteractiveMarkerViewer.hpp
@@ -35,9 +35,9 @@ public:
     dart::dynamics::SkeletonPtr const &skeleton);
 
   /// Visualizes tsr with at most n samples
-  TSRMarkerPtr visualizeTSR(
-    aikido::constraint::TSR const &_tsr,
-    int nSamples = 10, std::string basename = "");
+  TSRMarkerPtr addTSRMarker(
+    aikido::constraint::TSR const &tsr,
+    int nSamples = 10, const std::string &basename = "");
 
   void setAutoUpdate(bool _flag);
   void update();
--- a/include/aikido/rviz/InteractiveMarkerViewer.hpp
+++ b/include/aikido/rviz/InteractiveMarkerViewer.hpp
@@ -8,6 +8,7 @@
 #include <dart/dynamics/SmartPointer.hpp>
 #include <interactive_markers/interactive_marker_server.h>
 #include "SmartPointers.hpp"
+#include <aikido/constraint/TSR.hpp>
 
 namespace aikido {
 namespace rviz {
@@ -32,6 +33,10 @@ public:
   SkeletonMarkerPtr CreateSkeletonMarker(
     dart::dynamics::SkeletonPtr const &skeleton);
 
+  /// Visualizes tsr with at most n samples
+  void visualizeTSR(const aikido::constraint::TSR& _tsr, int nSamples = 10);
+  void cleanTSR();
+
   void setAutoUpdate(bool _flag);
   void update();
 
@@ -42,6 +47,7 @@ private:
   interactive_markers::InteractiveMarkerServer mMarkerServer;
   std::set<SkeletonMarkerPtr> mSkeletonMarkers;
   std::set<FrameMarkerPtr> mFrameMarkers;
+  std::vector<std::shared_ptr<dart::dynamics::SimpleFrame>> mSimpleFrames;
 
   std::atomic_bool mRunning;
   std::atomic_bool mUpdating;
--- a/include/aikido/rviz/InteractiveMarkerViewer.hpp
+++ b/include/aikido/rviz/InteractiveMarkerViewer.hpp
@@ -34,7 +34,11 @@ public:
   SkeletonMarkerPtr CreateSkeletonMarker(
     dart::dynamics::SkeletonPtr const &skeleton);
 
-  /// Visualizes tsr with at most n samples
+  /// Visualizes a TSR.
+  /// \param tsr TSR constraint
+  /// \param nSamples Max number of samples to be used in visualization
+  /// \param basename Basename for markers
+  /// \return TSRMarkerPtr contains sampled frames of TSR.
   TSRMarkerPtr addTSRMarker(
     aikido::constraint::TSR const &tsr,
     int nSamples = 10, const std::string &basename = "");
--- a/include/aikido/rviz/InteractiveMarkerViewer.hpp
+++ b/include/aikido/rviz/InteractiveMarkerViewer.hpp
@@ -35,8 +35,9 @@ public:
     dart::dynamics::SkeletonPtr const &skeleton);
 
   /// Visualizes tsr with at most n samples
-  TSRMarkerPtr visualizeTSR(aikido::constraint::TSR const &_tsr, int nSamples = 10,
-    std::string basename = "");
+  TSRMarkerPtr visualizeTSR(
+    aikido::constraint::TSR const &_tsr,
+    int nSamples = 10, std::string basename = "");
 
   void setAutoUpdate(bool _flag);
   void update();
--- a/include/aikido/rviz/InteractiveMarkerViewer.hpp
+++ b/include/aikido/rviz/InteractiveMarkerViewer.hpp
@@ -1,38 +1,42 @@
-#ifndef AIKIDO_RVIZ_INTERACTIVEMARKERVIEWER_H_
-#define AIKIDO_RVIZ_INTERACTIVEMARKERVIEWER_H_
+#ifndef AIKIDO_RVIZ_INTERACTIVEMARKERVIEWER_HPP_
+#define AIKIDO_RVIZ_INTERACTIVEMARKERVIEWER_HPP_
+
 #include <atomic>
 #include <set>
 #include <thread>
+
 #include <ros/ros.h>
 #include <dart/dynamics/Frame.hpp>
 #include <dart/dynamics/SmartPointer.hpp>
 #include <interactive_markers/interactive_marker_server.h>
-#include "SmartPointers.hpp"
+
 #include <aikido/constraint/TSR.hpp>
-#include "TSRMarker.hpp"
+#include <aikido/rviz/SmartPointers.hpp>
+#include <aikido/rviz/TSRMarker.hpp>
 
 namespace aikido {
 namespace rviz {
 
-class InteractiveMarkerViewer {
+class InteractiveMarkerViewer
+{
 public:
-  InteractiveMarkerViewer(std::string const &topicNamespace);
+  InteractiveMarkerViewer(const std::string& topicNamespace);
   virtual ~InteractiveMarkerViewer();
 
-  InteractiveMarkerViewer(InteractiveMarkerViewer const &) = delete;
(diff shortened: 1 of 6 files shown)


src/rviz/InteractiveMarkerViewer.cpp
```diff
@@ -49,6 +49,39 @@ FrameMarkerPtr InteractiveMarkerViewer::addFrame(
...
+  using dart::dynamics::Frame;
+  using dart::dynamics::SimpleFrame;
+  using aikido::constraint::TSR;
+
+  auto sampler = _tsr.createSampleGenerator();
+  auto state = _tsr.getSE3()->createState();
+
+  for(int i = 0; i < nSamples && sampler->canSample(); ++i)
+  {
+    sampler->sample(state);
```
mkoval: Check the return value of `sample`. In principle it should not return `false` if `canSample()` returned `true`, so an assertion would be fine.

src/rviz/InteractiveMarkerViewer.cpp
```diff
@@ -49,6 +49,39 @@ FrameMarkerPtr InteractiveMarkerViewer::addFrame(
...
+
+    std::stringstream sstm;
+    sstm << "tsr" << i;
```
mkoval: Suggestion: Make the base name a `std::string` parameter. I suggest making it default to something that includes the `&tsr` (or, even better, the memory address of the return value) to avoid name collisions.

src/rviz/InteractiveMarkerViewer.cpp
```diff
@@ -49,6 +49,39 @@ FrameMarkerPtr InteractiveMarkerViewer::addFrame(
...
+
+    auto tsrFrame = std::make_shared<SimpleFrame>(
+      Frame::World(), sstm.str(), state.getIsometry());
+    mSimpleFrames.push_back(tsrFrame);
+    this->addFrame(tsrFrame.get());
```
mkoval: Nit: We don't typically explicitly specify `this->` unless there is a name conflict.

Also, I'd prefer to aggregate all of the `FrameMarkerPtr`s returned by `addFrame` into an RAII object (perhaps `TSRMarker` 😉) that we return from this function. That would also make `clearTSRs` and the `mSimpleFrames` member variable superfluous, since they could be replaced by the RAII object.

include/aikido/rviz/InteractiveMarkerViewer.hpp
```diff
@@ -32,6 +33,10 @@ class InteractiveMarkerViewer {
   SkeletonMarkerPtr CreateSkeletonMarker(
     dart::dynamics::SkeletonPtr const &skeleton);
 
+  /// Visualizes tsr with at most n samples
+  void visualizeTSR(const aikido::constraint::TSR& _tsr, int nSamples = 10);
```
mkoval: Please add a full docstring.

src/rviz/InteractiveMarkerViewer.cpp
```diff
@@ -49,6 +50,42 @@ FrameMarkerPtr InteractiveMarkerViewer::addFrame(
   return marker;
 }
 
+TSRMarkerPtr InteractiveMarkerViewer::visualizeTSR(
```
mkoval: Rename this to `addTSRMarker` to be consistent with `addFrameMarker`.

src/rviz/InteractiveMarkerViewer.cpp
```diff
@@ -49,6 +50,42 @@ FrameMarkerPtr InteractiveMarkerViewer::addFrame(
...
+  aikido::constraint::TSR const &_tsr,
+  int nSamples,
+  std::string basename)
```
mkoval: Take `basename` by `const std::string&`.

src/rviz/InteractiveMarkerViewer.cpp
```diff
@@ -49,6 +50,42 @@ FrameMarkerPtr InteractiveMarkerViewer::addFrame(
...
+  using aikido::constraint::TSR;
+
+  auto sampler = _tsr.createSampleGenerator();
+  auto state = _tsr.getSE3()->createState();
+  auto tsrMarker = std::make_shared<TSRMarker>();
+
+  if (basename == "")
+  {
+      std::ostringstream ost;
+      ost << &tsrMarker;
```
mkoval: `tsrMarker` is a `smart_ptr`, so this should be `tsrMarker.get()` to get the memory address of the heap memory. It's possible that repeated calls of this function in a loop could allocate `tsrMarker` itself at the same memory address on the stack.

src/rviz/InteractiveMarkerViewer.cpp
```diff
@@ -49,6 +50,42 @@ FrameMarkerPtr InteractiveMarkerViewer::addFrame(
...
+{
+  using dart::dynamics::Frame;
+  using dart::dynamics::SimpleFrame;
+  using aikido::constraint::TSR;
+
+  auto sampler = _tsr.createSampleGenerator();
+  auto state = _tsr.getSE3()->createState();
+  auto tsrMarker = std::make_shared<TSRMarker>();
+
+  if (basename == "")
```
mkoval: Nit: `basename.empty()`.

src/rviz/InteractiveMarkerViewer.cpp
```diff
@@ -49,6 +50,42 @@ FrameMarkerPtr InteractiveMarkerViewer::addFrame(
...
+      std::ostringstream ost;
+      ost << &tsrMarker;
+      basename = ost.str();
+  }
+  for(int i = 0; i < nSamples && sampler->canSample(); ++i)
+  {
+    assert(sampler->sample(state));
+
+    std::stringstream ss;
+    ss << "Marker[" << basename << "].frame[" << i << "]";
```
mkoval: Nit: `Marker` -> `TSRMarker` to avoid collisions with other markers we add in the future.

include/aikido/rviz/TSRMarker.hpp
```diff
@@ -0,0 +1,26 @@
...
+
+namespace aikido {
+namespace rviz {
+
+class TSRMarker {
+public:
+  TSRMarker();
+  ~TSRMarker(){};
+
+  void addFrameMarker(FrameMarkerPtr const &marker);
```
mkoval: I don't think this should be part of the public API. The simplest option would be to modify `TSRMarker`'s constructor to directly accept the `std::set<FrameMarkerPtr>`.

include/aikido/rviz/TSRMarker.hpp
```diff
@@ -0,0 +1,26 @@
...
+#include "FrameMarker.hpp"
+#include "SmartPointers.hpp"
+
+namespace aikido {
+namespace rviz {
+
+class TSRMarker {
+public:
+  TSRMarker();
+  ~TSRMarker(){};
```
mkoval: Nit: Use `~TSRMarker = default` instead.

include/aikido/rviz/TSRMarker.hpp
```diff
@@ -0,0 +1,26 @@
...
+#define AIKIDO_RVIZ_TSRMARKER_H_
+
+#include <set>
+#include "FrameMarker.hpp"
+#include "SmartPointers.hpp"
+
+namespace aikido {
+namespace rviz {
+
+class TSRMarker {
```
mkoval: Nit: Either mark this as `final` or mark the destructor as `virtual`.

include/aikido/rviz/TSRMarker.hpp
```diff
@@ -0,0 +1,26 @@
...
+public:
+  TSRMarker();
+  ~TSRMarker(){};
+
+  void addFrameMarker(FrameMarkerPtr const &marker);
+
+private:
+  std::set<FrameMarkerPtr> mFrameMarkers;
```
mkoval: Nit: Why does this need to be an `std::set`? I'd prefer an `std::vector` here, since linear memory is a lot simpler than a red-black tree, if we don't require uniqueness.

gilwoolee:
@mkoval I think I addressed all of your comments. Please take a look. Thanks for the help!

src/rviz/InteractiveMarkerViewer.cpp
```diff
@@ -49,6 +50,52 @@ FrameMarkerPtr InteractiveMarkerViewer::addFrame(
...
+    auto sampled = sampler->sample(state);
+    assert(sampled);
+
+    std::stringstream ss;
+    ss << "TSRMarker[" << name << "].frame[" << i << "]";
+
+    auto tsrFrame = std::make_shared<SimpleFrame>(
+      Frame::World(), ss.str(), state.getIsometry());
+    auto frameMarker = addFrame(tsrFrame.get());
+    tsrFrames.emplace_back(tsrFrame);
```
mkoval: Nit: Could we use `unique_ptr` instead of `shared_ptr` here?

gilwoolee:
@mkoval Would you mind taking a last look before I merge this? It just changes `shared_ptr` to `unique_ptr`, taking your last suggestion. :)

include/aikido/rviz/SmartPointers.hpp
```diff
@@ -17,6 +17,10 @@ class SkeletonMarker;
 typedef std::shared_ptr<SkeletonMarker> SkeletonMarkerPtr;
 typedef std::shared_ptr<SkeletonMarker const> SkeletonMarkerConstPtr;
 
+class TSRMarker;
+typedef std::shared_ptr<TSRMarker> TSRMarkerPtr;
+typedef std::shared_ptr<TSRMarker const> TSRMarkerConstPtr;
```
mkoval: Nit: Prefer `using` aliases to `typedef`s.

src/rviz/TSRMarker.cpp
```diff
@@ -0,0 +1,14 @@
+#include <aikido/rviz/TSRMarker.hpp>
+
+using dart::dynamics::SimpleFrame;
+
+namespace aikido {
+namespace rviz {
+
+TSRMarker::TSRMarker(std::vector<std::unique_ptr<SimpleFrame>> tsrFrames)
```
jslee02: We could use rvalue reference (`std::vector<std::unique_ptr<SimpleFrame>>&& tsrFrames`) here to prevent unnecessary copy.

include/aikido/rviz/InteractiveMarkerViewer.hpp
```diff
@@ -32,6 +34,15 @@ class InteractiveMarkerViewer {
   SkeletonMarkerPtr CreateSkeletonMarker(
     dart::dynamics::SkeletonPtr const &skeleton);
 
+  /// Visualizes a TSR.
+  /// \param tsr TSR constraint
+  /// \param nSamples Max number of samples to be used in visualization
+  /// \param basename Basename for markers
+  /// \return TSRMarkerPtr contains sampled frames of TSR.
+  TSRMarkerPtr addTSRMarker(
+    aikido::constraint::TSR const &tsr,
```
jslee02: I believe [the convention](https://github.com/personalrobotics/aikido/blob/master/STYLE.md#c-style) is to place `const` before type and space after `&` as `const aikido::constraint::TSR& tsr`.

include/aikido/rviz/InteractiveMarkerViewer.hpp
```diff
@@ -32,6 +34,15 @@ class InteractiveMarkerViewer {
...
+    int nSamples = 10, const std::string &basename = "");
```
jslee02: Please put a space after & not not before. See above.

include/aikido/rviz/SmartPointers.hpp
jslee02: Please place `const` before type. See above.

src/rviz/TSRMarker.cpp
mkoval: This already doesn't make any copies if the caller uses `std::move`, e.g.:
```c++
// Caller std::move's from local variable to argument:
std::vector<std::unique_ptr<SimpleFrame>> frames = /* ... */;
auto marker = TSRMarker{std::move(frames)};

// Constructor std::moves from argument to member variable:
TSRMarker::TSRMarker(std::vector<std::unique_ptr<SimpleFrame>> frames)
  : markers{std::move(frames)} {}
```
Note that `std::vector<std::unique_ptr<...>>` is not copyable, so we already **must not** be introducing an extra copy. 😉 

My general understanding is that rvalue references are only useful for: (1) move constructors and (2) perfect forwarding in templates due to [reference collapsing](http://stackoverflow.com/a/13726754/111426).

jslee02: > This already doesn't make any copies if the caller uses `std::move`

That is true, but the explicit rvalue argument could prevent passing lvalue by accident (if I'm missing something here). For example:
```cpp
class TSRMarker
{
public:
  TSRMarker(std::vector<std::unique_ptr<SimpleFrame>>&& tsrFrames);
  // ...
};
std::vector<std::unique_ptr<SimpleFrame>> frames = /* ... */;
auto marker1 = TSRMarker{frames}; // error
auto marker2 = TSRMarker{std::move(frames)}; // okay
```

jslee02: > Note that std::vector<std::unique_ptr<...>> is not copyable, so we already must not  be introducing an extra copy.

Hm, this is what I missed. It still prevents passing lvalue even without making it rvalue reference but in a different way (we get different error messages).

I still prefer to make the argument _explicitly_ rlvalue reference because we always expect to pass rlvalue here. However, I am not strongly against leaving it as it is since it works anyways as we intend.

mkoval: I disagree! 😬 I prefer to leave copy versus move up to the caller (i.e. leave the code as it is).

I don't feel particularly strongly either way, so let's leave it up to @gilwoolee to decide. 😄

jslee02: > I don't feel particularly strongly either way, so let's leave it up to @gilwoolee to decide.

:+1: 

My only concern is that the caller is _not able to pass lvalue anyway_ (whether or not the argument is rvalue reference because copy assignment of `std::unique_ptr` is not defined), but explicit rvalue reference argument could give us hint why the caller cannot pass lvalue.

Alright, let's pass the baton to @gilwoolee. :smiling_imp:

PR #146, opened 2017-04-02
Adding WeldJoint and ConstantSampler

gilwoolee:
This PR adds the following:
  - dart/statespace/WeldJoint 
  - constraint/uniform/ConstantSampler

and uses ConstantSampler as a sampler for WeldJoint.

---
This change is [Reviewable](https://reviewable.io/reviews/personalrobotics/aikido/146)


gilwoolee:
@mkoval I initially had that, which works fine, but it would unnecessarily try to "sample" when it can simply return a fixed value, so I thought this would be cleaner? I can switch back to using `RnBoxConstraint` if necessary.

include/aikido/constraint/uniform/ConstantSampler.hpp
```diff
@@ -0,0 +1,39 @@
...
+#include "../../statespace/Rn.hpp"
+#include "../Sampleable.hpp"
+
+namespace aikido {
+namespace constraint {
+
+/// ConstantSampler for RealVectorStates.
+/// Stub sampler for WeldJoint or any fixed constant state space.
+class ConstantSampler
+  : public constraint::Sampleable
```
mkoval: Rename `ConstantSampler` to `RnConstantSampler` since this only works on an `Rn` state space.

src/constraint/uniform/ConstantSampler.cpp
```diff
@@ -0,0 +1,105 @@
+#include <stdexcept>
+#include <aikido/constraint/uniform/ConstantSampler.hpp>
+
+namespace aikido {
+namespace constraint {
+
+//=============================================================================
+class ConstantSamplerSampleGenerator
+  : public constraint::SampleGenerator
```
mkoval: - Rename to `RnConstantSamplerSampleGenerator` (see above).
- Define this class in an anonymous `namespace` since it is only used in this file.

src/statespace/dart/WeldJoint.cpp
```diff
@@ -0,0 +1,32 @@
...
+  : Rn(_joint->getNumDofs())
+  , JointStateSpace(_joint)
+{
+}
+
+//=============================================================================
+void WeldJoint::convertPositionsToState(
+  const Eigen::VectorXd& _positions, StateSpace::State* _state) const
+{
+  setValue(static_cast<State*>(_state), _positions);
```
jslee02: I believe we don't need to convert any value here because the dimension of WeldJoint is zero.

src/statespace/dart/WeldJoint.cpp
```diff
@@ -0,0 +1,32 @@
...
+}
+
+//=============================================================================
+void WeldJoint::convertStateToPositions(
+  const StateSpace::State* _state, Eigen::VectorXd& _positions) const
+{
+  _positions = getValue(static_cast<const State*>(_state));
```
jslee02: Nothing to get values from `_state`, which is zero dimensional. It would be sufficient to resize `_positions` to zero as `_positions.resize(0);` (see above)

jslee02:
Addressed @mkoval 's comments.

gilwoolee:
@jslee02 Sorry to ask you, but would you mind writing unit tests for the two classes? Thanks a lot!!

PR #111, opened 2016-08-31
Aikido ROS control

ClintLiddick:
@Shushman @mkoval can you compare this with https://github.com/personalrobotics/aikido/pull/103 and see which needs to be worked on? I branched from the `mkoval` branch and made the fix to use the component refactor.

---

This change is [Reviewable](https://reviewable.io/reviews/personalrobotics/aikido/111)

PR #147, opened 2017-04-02
Function to convert trajectory_msgs::JointTrajectory into aikido::trajectory::Spline

gilwoolee:
This PR branches from #111. 
I have one question. It seems like the use of DOF and Joint is mixed here. `convertJointTrajectory` asks for `_space->getNumSubSpaces()` which would return the number of Joints, but then considers these as DOFs. This would work fine (except for name-matching) if all joints are single-dof joints. Is that the case since we're using `JointTrajectory`, not `MutliDOFJointTrajectory`? 

@mkoval @ClintLiddick Please confirm. If that's the case, I'll change the variable names, docstring, and name-checking part to clarify this.


ClintLiddick:
I took a look and I'm not exactly sure. I think this is a latent bug or, like you said, implicitly assuming a single-DOF joint (for `JointTrajectory` not `MultiDOFJointTrajectory`). I think if we gave this a true multi-DOF joint it would throw an incorrect error. Correct thing is probably to change the strings + docs + error message, and add a comment that this only works for single-DOF joints (i.e. no SO3/SE3 I think). I'm not quite sure what the right logic would be for handling multi-DOF joints, or if it could live in the same conversion functions. @mkoval thoughts?

src/control/ros/Conversions.cpp
```diff
@@ -0,0 +1,215 @@
...
+//=============================================================================
+std::unique_ptr<SplineTrajectory> convertJointTrajectory(
+  const std::shared_ptr<MetaSkeletonStateSpace>& _space,
+  const trajectory_msgs::JointTrajectory& _jointTrajectory)
+{
+  if (!_space)
+    throw std::invalid_argument{"StateSpace must be non-null."};
+
+  const auto numControlledDofs = _space->getNumSubspaces();
+  if (_jointTrajectory.joint_names.size() != numControlledDofs)
```
mkoval: Rename `numControlledDofs` to `numControlledJoints`.

src/control/ros/Conversions.cpp
```diff
@@ -0,0 +1,215 @@
...
+    message << "Incorrect number of joints: expected "
+        << numControlledDofs << ", got "
+        << _jointTrajectory.joint_names.size() << ".";
+    throw std::invalid_argument{message.str()};
+  }
+
+  if (_jointTrajectory.points.size() < 2)
+  {
+    throw std::invalid_argument{
+      "Trajectory must contain at least two or more waypoints."};
```
mkoval: Nit: "Trajectory must contain two or more waypoints."

src/control/ros/Conversions.cpp
```diff
@@ -0,0 +1,215 @@
...
+    auto joint = _space->getJointSpace(i)->getJoint();
+    if (_jointTrajectory.joint_names[i] != joint->getName())
+    {
+      std::stringstream message;
+      message << "Unknown joint name: expected "
+          << joint->getName() << ", got "
+          << _jointTrajectory.joint_names[i] << ".";
+      throw std::invalid_argument{message.str()};
+    }
+  }
```
mkoval: This check will fail if `_jointTrajectory` and `_space` have the same joints, but in a different order. We should build a map from joint name to subspace and use it here.

Also, add a check that throws an exception if any of the joints are not single DOF.

src/control/ros/Conversions.cpp
```diff
@@ -0,0 +1,215 @@
...
+      "Velocity is required since acceleration is specified."};
+  }
+
+  int numCoefficients;
+  if (isAccelerationRequired)
+    numCoefficients = 6; // quintic
+  else if (isVelocityRequired)
+    numCoefficients = 4; // cubic
+  else
+    numCoefficients = 2; // linear;
```
mkoval: Nit: Remove extra `;`.

src/control/ros/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
+#==============================================================================
+# Dependencies
+#
+find_package(trajectory_msgs REQUIRED)
+aikido_check_package(trajectory_msgs "aikido::control::ros" "trajectory_msgs")
```
mkoval: The `control/ros` namespace is intended to be optional, so `traectory_msgs` should *not* be marked as `REQUIRED`. We should change this to:
```
find_package(trajectory_msgs QUIET)
aikido_check_package(geometry_msgs "control::ros" "trajectory_msgs")
```

tests/control/ros/CMakeLists.txt
```diff
@@ -0,0 +1,6 @@
+
```
mkoval: Add a check here to only build these tests if the `ros::control` component was built.

Shushman:
FYI, the `test_Conversions` is failing here with the following error:

```
[ RUN      ] ConvertJointTrajectoryTests.LinearTrajectory
unknown file: Failure
C++ exception with description "Unknown joint name: expected Joint1, got DegreeOfFreedom1." thrown in the test body.
[  FAILED  ] ConvertJointTrajectoryTests.LinearTrajectory (0 ms)
```

and similarly for `CubicTrajectory` and `QuinticTrajectory`

gilwoolee:
@Shushman Yeah that's because it tries to match joint names with dof names.  Fixing it now. :D

mkoval:
> Yeah that's because it tries to match joint names with dof names.

I don't know what that means, but I'm glad @gilwoolee figured it out! 😈

src/control/ros/Conversions.cpp
gilwoolee: @mkoval  I am a bit confused about your first comment. If we can't assume that they are given in the same order, how do we build the initial mapping? Should we compare each `_jointTrajectory.joint_names[i]`with all joint names in `_space->getJointSpace(j)->getJoint()->getName()`? 

One more question: is it  _joint_name_ or _dof_name_ in `_jointTrajectory` when we get it from ROS? (Assuming single dofs. Are they the same?)

tests/control/ros/CMakeLists.txt
```diff
@@ -0,0 +1,6 @@
+if(${FOUND_${PROJECT_NAME}_control_ros})
```
jslee02: `aikido_control_ros` is a local target so `FOUND_aikido_control_ros` wouldn't be defined. Instead, please use `if(TARGET "${PROJECT_NAME}_control_ros")`.

> **if(TARGET target-name)**
True if the given name is an existing logical target name such as those created by the add_executable(), add_library(), or add_custom_target() commands.

src/control/ros/Conversions.cpp
```diff
@@ -0,0 +1,228 @@
...
+  }
+
+  // Check that all joints are single DOF.
+  for (size_t i = 0; i < _space->getNumSubspaces(); ++i)
+  {
+    auto n = _space->getJointSpace(i)->getJoint()->getNumDofs();
+    if (n != 1)
+    {
+      std::stringstream message;
+      message << "[Conversion] Expected 1 dof. Joint "
```
jslee02: Nit: Only this warning message has the heading `[Conversion]`. It would good to be consistent by either removing the heading from here or adding the heading to all other messages.

src/control/ros/Conversions.cpp
mkoval: > If we can't assume that they are given in the same order, how do we build the initial mapping? Should we compare each `_jointTrajectory.joint_names[i]` with all joint names in `_space->getJointSpace(j)->getJoint()->getName()`?

That was my idea. None of these solutions are perfect, but I trust the names to be correct more than I trust the indices to be.

> One more question: is it `joint_name` or `dof_name` that we get from `_jointTrajectory.joint_names[]` when we get it from actual HERB?

There is no correct answer here, since the `JointTrajectory` message has no distinction between joints and DOFs. I would strongly prefer `Joint`, since that: (1) it seems that `DegreeOfFreedom` names are typically (but not always) constructed from their `Joint`'s name, (2) URDF only provides a mechanism for specifying joint names, and (3) it semantically matches the `joint_name` field in the message.

> (Assuming single dofs. Are they the same?)

They are not guaranteed to be the same in DART, but I believe they are by default. @jslee02 Is that correct?

src/control/ros/Conversions.cpp
mkoval: :+1: For removing the heading here.

gilwoolee:
@jslee02 Thanks for a quick review! @mkoval Would you mind giving it another pass? :D Thanks a lot!!

src/control/ros/Conversions.cpp
```diff
@@ -0,0 +1,283 @@
...
+  // Check that all joints are single DOF.
+  for (size_t i = 0; i < space->getNumSubspaces(); ++i)
+  {
+    auto n = space->getJointSpace(i)->getJoint()->getNumDofs();
+    if (n != 1)
+    {
+      std::stringstream message;
+      message << "Expected 1 dof. Joint " << i << " has " << n << " dofs.";
+      throw std::invalid_argument{message.str()};
+    }
```
mkoval: Check that `getJointSpace(i)`is `Rn` or `SO2`.

src/control/ros/Conversions.cpp
```diff
@@ -0,0 +1,283 @@
...
+  }
+
+  // Check that all joints are single DOF.
+  for (size_t i = 0; i < space->getNumSubspaces(); ++i)
+  {
+    auto n = space->getJointSpace(i)->getJoint()->getNumDofs();
+    if (n != 1)
+    {
+      std::stringstream message;
+      message << "Expected 1 dof. Joint " << i << " has " << n << " dofs.";
```
mkoval: Nit: It would be helpful to include `joint->getName()` and `joint->getType()` to `message`, e.g.:

> Only single-DOF joints are supported. Joint "my_joint1" (index: 1) is a SphericalJoint with 3 DOFs.

src/control/ros/Conversions.cpp
```diff
@@ -0,0 +1,283 @@
...
+        {
+          std::stringstream message;
+          message << "Both subspace[" << search->second << "] and ["
+            << j << "] map to joint " << i << " in jointTrajectory.";
+          throw std::invalid_argument{message.str()};
+        }
+        rosJointToMSSSJoint.emplace(std::make_pair(i, j));
+        found_match = true;
+        break;
+      }
```
mkoval: Encapsulate this in a helper function. Putting this search inline means that we can't easily improve it's efficiency later (e.g. using `unordered_map`) and makes it harder to understand this function.

I suggest writing a helper function to get a `Joint` by name from a `MetaSkeleton`. If the name is ambiguous, which can occur in a `MetaSkeleton` that contain `Joint`s from multiple `Skeleton`s, `throw` an exception. Then, this code simply becomes:

        auto metaSkeleton = space->getMetaSkeleton();
        auto joint = getJointByName(*metaSkeleton, jointName);
        auto index = metaSkeleton->getIndex(joint);

src/control/ros/Conversions.cpp
```diff
@@ -0,0 +1,283 @@
...
+  Eigen::MatrixXd conversionPermutation;
+  convertToPermutationMatrix(rosJointToMSSSJoint, &conversionPermutation);
+  const auto& waypoints = jointTrajectory.points;
+  for (size_t iwaypoint = 1; iwaypoint < waypoints.size(); ++iwaypoint)
+  {
+    Eigen::VectorXd nextPosition, nextVelocity, nextAcceleration;
+    extractJointTrajectoryPoint(jointTrajectory, iwaypoint, numControlledJoints,
+      &nextPosition, isPositionRequired,
+      &nextVelocity, isVelocityRequired,
+      &nextAcceleration, isAccelerationRequired);
```
mkoval: I'd like to perform the `reorder` operation here. These two operations are closely related because they convert from a ROS data structure to the Aikido data structure that we expect. The rest of this loop purely performs operations on Aikido data structures, e.g. fitting splines.

src/control/ros/Conversions.cpp
```diff
@@ -0,0 +1,283 @@
...
+  Eigen::MatrixXd* permutationMatrix)
+{
+  *permutationMatrix = Eigen::MatrixXd::Zero(map.size(), map.size());
+  for(auto it = map.begin(); it != map.end(); ++it)
+  {
+    (*permutationMatrix)(it->second, it->first) = 1.0;
+  }
+}
+
+} // namespace
```
mkoval: I suspect that using a permutation matrix here will be _a lot_ slower than operating directly on the list of `std::pair<size_t, size_t>`s. I also think it is is more complicated to implement this way, since it involves an intermediate data structure (the permutation matrix) that would not be otherwise necessary.

Since this is an entirely an implementation detail, I am fine keeping it as it is for now. However, we may want to consider refactoring this in the future if performance becomes an issue.

tests/control/ros/test_Conversions.cpp
```diff
@@ -0,0 +1,275 @@
...
+  EXPECT_EIGEN_EQUAL(make_vector(4.), values, kTolerance);
+
+  trajectory->evaluateDerivative(0., 2, values);
+  EXPECT_EIGEN_EQUAL(make_vector(5.), values, kTolerance);
+
+  trajectory->evaluateDerivative(1., 2, values);
+  EXPECT_EIGEN_EQUAL(make_vector(6.), values, kTolerance);
+}
+
+TEST_F(ConvertJointTrajectoryTests, LinearTrajectoryWithDifferentOrdering)
```
mkoval: Nit: The fact that you are using a linear trajectory in this test is an implementation detail. I suggest renaming the test to just `DifferentJointOrdering`.

tests/control/ros/test_Conversions.cpp
```diff
@@ -0,0 +1,275 @@
...
+  trajectory->evaluate(1., state);
+  mStateSpace2Joints->convertStateToPositions(state, values);
+  EXPECT_EIGEN_EQUAL(make_vector(3., 2.), values, kTolerance);
+
+  trajectory->evaluateDerivative(0.5, 1, values);
+  EXPECT_EIGEN_EQUAL(make_vector(1., 1.), values, kTolerance);
+
+  trajectory->evaluateDerivative(0.5, 2, values);
+  EXPECT_EIGEN_EQUAL(make_vector(0., 0.), values, kTolerance);
+}
```
mkoval: - Add a test for a `MetaSkeleton` that has multiple joints with the same name. This is possible by creating a `Group` that contains the `Joint` with the same name in two different `Skeleton`s.
- Add a test for a `MetaSkeleton` that is missing one of the joints named in the `JointTrajectoryMessage`.

src/control/ros/Conversions.cpp
```diff
@@ -0,0 +1,288 @@
...
+
+//=============================================================================
+dart::dynamics::Joint* findJointByName(
+    dart::dynamics::MetaSkeleton* metaSkeleton, const std::string& jointName)
+{
+  for (size_t i = 0; i < metaSkeleton->getNumJoints(); ++i)
+  {
+    if (metaSkeleton->getJoint(i)->getName() == jointName)
+      return metaSkeleton->getJoint(i);
+  }
```
mkoval: Check if the `MetaSkeleton` has more than one joint with this name.

src/control/ros/Conversions.cpp
```diff
@@ -0,0 +1,288 @@
...
+  *outVector = Eigen::VectorXd::Zero(inVector.size());
+  for (auto it = indexMap.begin(); it != indexMap.end(); ++it)
+  {
+    (*outVector)(it->second) = inVector(it->first);
+  }
+}
+
+//=============================================================================
+dart::dynamics::Joint* findJointByName(
+    dart::dynamics::MetaSkeleton* metaSkeleton, const std::string& jointName)
```
mkoval: Nit: `metaSkeleton` is not an out parameter, so we should take it by reference. In this case, I think we can actually take it `const &`.

src/control/ros/Conversions.cpp
```diff
@@ -0,0 +1,288 @@
...
+  if (jointTrajectory.joint_names.size() != numControlledJoints)
+  {
+    std::stringstream message;
+    message << "Incorrect number of joints: expected "
+        << numControlledJoints << ", got "
+        << jointTrajectory.joint_names.size() << ".";
+    throw std::invalid_argument{message.str()};
+  }
+
+  // Check that all joints are single DOF.
```
mkoval: Nit: "...RnJoint or SO2Joint state spaces."

src/control/ros/Conversions.cpp
```diff
@@ -0,0 +1,288 @@
...
+  }
+
+  // Map joint indices between jointTrajectory and space subspaces.
+  std::map<size_t, size_t> rosJointToMetaSkeletonJoint;
+
+  auto metaSkeleton = space->getMetaSkeleton();
+
+  for (size_t i = 0; i < jointTrajectory.joint_names.size(); ++i)
+  {
+    std::string jointName = jointTrajectory.joint_names[i];
```
mkoval: Nit: `const auto&`.

src/control/ros/Conversions.cpp
```diff
@@ -0,0 +1,288 @@
...
+    {
+      std::stringstream message;
+      message << "Joint " << jointName << " (index: " << i << ")"
+        << " does not exist in metaSkeleton.";
+      throw std::invalid_argument{message.str()};
+    }
+    auto index = metaSkeleton->getIndexOf(joint);
+    assert(index != dart::dynamics::INVALID_INDEX);
+
+    rosJointToMetaSkeletonJoint.emplace(std::make_pair(i, index));
```
mkoval: Verify that this is not a duplicate, i.e. that `rosJointToMetaSkeletonJoint[i]` does not already exist. You can do this by checking the `bool` flag in the `std::pair` returned by `emplace`.

Nit: Also, it would be nice to more descriptive variable names than `i` and `index`. 😉 Maybe `metaSkeletonIndex` and `trajectoryIndex`?

src/control/ros/Conversions.cpp
jslee02: Hm, I know polymorphisms work both for pointers and references, but I thought our convention is to take a const pointer.

src/control/ros/Conversions.cpp
jslee02: As a note, I'm thinking of implementing this to `dart::dynamics::MetaSkeleton`. That function would return a list of joint pointers whoes names are all `jointName` because `MetaSkeleton` possibly has multiple joints with the same name.

mkoval: @jslee02 That would be a great addition. 👍

In fact, it would be nice to have two functions: (1) one that returns a list of joints and (2) one that returns a single joint and prints a warning if multiple joints have the same name.

src/control/ros/Conversions.cpp
mkoval: The convention I've seen (and generally follow) is to prefer references and _only_ use pointers for: (1) output parameters or (2) optional parameters that may be `nullptr`.

@psigen Want to break the tie? 😬

jslee02: I haven't decided my preference here yet. If we want to use references for this case, then I see there are tons of changes in Aikido code base we need to make for the consistency. Just saying. :sweat_smile:

mkoval: Are you sure? Note that `State` variables are an exception to this rule because I did some sketchy business to reduce their memory footprint and am not sure [whether or not calling `*state` to create a reference invokes undefined behavior](http://stackoverflow.com/q/36457841/111426).

src/control/ros/Conversions.cpp
```diff
@@ -0,0 +1,322 @@
...
+  const dart::dynamics::MetaSkeleton& metaSkeleton,
+  const std::string& jointName)
+{
+  std::vector<const dart::dynamics::Joint*> joints;
+
+  for (size_t i = 0; i < metaSkeleton.getNumJoints(); ++i)
+  {
+    if (metaSkeleton.getJoint(i)->getName() == jointName)
+    {
+      joints.emplace_back(metaSkeleton.getJoint(i));
```
mkoval: Nit: I'd prefer to store `metaSkeleton->getJoint(i)` in a variable so we don't have to call it twice.

src/control/ros/Conversions.cpp
```diff
@@ -0,0 +1,322 @@
...
+  for (size_t i = 0; i < numControlledJoints - 1; ++i)
+  {
+    if (joint_names[i] == joint_names[i+1])
+    {
+      std::stringstream message;
+      message << "JointTrajectory has multiple joints with same name ["
+        << joint_names[i] << "].";
+      throw std::invalid_argument{message.str()};
+    }
+  }
```
mkoval: Nit: it would be cleaner to replace this loop with:
```c++
auto duplicate_it = std:: adjacent_find(std::begin(joint_names), std::end(joint_names));
if (duplicate_it != std::end(joint_names)) { /* ... */ }
```

src/control/ros/Conversions.cpp
```diff
@@ -0,0 +1,322 @@
...
+    throw std::invalid_argument{message.str()};
+  }
+
+  // Check that the names in jointTrajectory are unique.
+  std::vector<std::string> joint_names;
+  joint_names.reserve(numControlledJoints);
+  for (size_t i = 0; i < numControlledJoints; ++i)
+  {
+    joint_names.emplace_back(jointTrajectory.joint_names[i]);
+  }
```
mkoval: Nit: Why is this loop necessary? I believe this is equivalent to:
```c++
auto joint_names = jointTrajectory.joint_names;
```

src/control/ros/Conversions.cpp
```diff
@@ -0,0 +1,322 @@
...
+      "Trajectory must contain two or more waypoints."};
+  }
+
+  // Map joint indices between jointTrajectory and space subspaces.
+  std::map<size_t, size_t> rosJointToMetaSkeletonJoint;
+
+  auto metaSkeleton = space->getMetaSkeleton();
+  auto nJoints = jointTrajectory.joint_names.size();
+
+  for (size_t trajIndex = 0; trajIndex < nJoints; ++trajIndex)
```
mkoval: Nit: The name `trajIndex` is ambiguous, since it could refer to a waypoint index or a joint index.

src/control/ros/Conversions.cpp
```diff
@@ -0,0 +1,322 @@
...
+  {
+    const auto& jointName = jointTrajectory.joint_names[trajIndex];
+    auto joints = findJointByName(*metaSkeleton, jointName);
+
+    if (joints.size() == 0)
```
mkoval: Nit: `joints.empty()`.

src/control/ros/Conversions.cpp
```diff
@@ -0,0 +1,322 @@
...
+    {
+      std::stringstream message;
+      message << "Joint " << jointName << " (index: " << trajIndex << ")"
+        << " does not exist in metaSkeleton.";
```
mkoval: Nit: This index is ambiguous because it could refer to an index in the `JointTrajectory` or an index in the `MetaSkeleton`. I suggest explicitly saying "trajectory index."

src/control/ros/Conversions.cpp
```diff
@@ -0,0 +1,322 @@
...
+      throw std::invalid_argument{message.str()};
+    }
+    else if  (joints.size() > 1)
+    {
+      std::stringstream message;
+      message << "Multiple (" << joints.size()
+        << ") joints have the same name [" << jointName << "].";
```
mkoval: Nit: This message is ambiguous because it could refer to duplicate joints in the `JointTrajectory` or the `MetaSkeleton`. I suggest explicitly saying "in the JointTrajectory."

gilwoolee:
Closing. Thanks a lot @mkoval @jslee02 @ClintLiddick !

src/control/ros/Conversions.cpp
jslee02: https://github.com/dartsim/dart/pull/879

mkoval: Thanks @jslee02!

PR #149, opened 2017-04-03
ros_control/RosTrajectoryExecutor

gilwoolee:
Need to review and write unit-tests. 

I have some issue compiling this. (I can only compile it if I comment out test-relevant lines in root CMakeLists.txt.)


include/aikido/control/ros/RosTrajectoryExecutor.hpp
```diff
@@ -0,0 +1,79 @@
...
+#include <ros/callback_queue.h>
+#include <control_msgs/FollowJointTrajectoryAction.h>
+#include <dart/dart.hpp>
+#include <aikido/control/TrajectoryExecutor.hpp>
+#include <aikido/trajectory/Trajectory.hpp>
+
+// actionlib and DART both #define this macro.
+#undef DEPRECATED
+#include <actionlib/client/action_client.h>
+#undef DEPRECATED
```
mkoval: @jslee02 Is this still necessary? I believe DART switched to the `DART_DEPRECATED` macro to avoid a conflict just like this.

src/control/ros/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+  ${actionlib_LIBRARIES}
+  ${control_msgs_LIBRARIES}
+  ${roscpp_LIBRARIES}
+)
+
+add_component(${PROJECT_NAME} control_ros)
+add_component_targets(${PROJECT_NAME} control_ros "${PROJECT_NAME}_control_ros")
+add_component_dependencies(${PROJECT_NAME} control_ros control statespace trajectory)
+
+# coveralls_add_sources(${sources})
```
mkoval: Unless there is a strong reason not to, we should uncomment this.

src/control/ros/RosTrajectoryExecutor.cpp
```diff
@@ -0,0 +1,290 @@
...
+      return "old header timestamp";
+
+    case Result::PATH_TOLERANCE_VIOLATED:
+      return "path tolerance violated";
+
+    case Result::GOAL_TOLERANCE_VIOLATED:
+      return "goal tolerance violated";
+
+    default:
+      return "unknown";
```
mkoval: Nit: Include `errorCode` here, e.g.:
```c++
std::stringstream errorString;
errorString << "unknown (" << errorCode << ")";
return errorString.str()
```

include/aikido/control/ros/RosTrajectoryExecutor.hpp
```diff
@@ -0,0 +1,79 @@
...
+  RosTrajectoryExecutor(
+    ::dart::dynamics::MetaSkeletonPtr skeleton, 
+    ::ros::NodeHandle node,
+    const std::string& serverName,
+    double timestep,
+    double goalTimeTolerance,
+    std::chrono::milliseconds connectionTimeout
+      = std::chrono::milliseconds{1000},
+    std::chrono::milliseconds connectionPollingRate
+      = std::chrono::milliseconds{20}
```
mkoval: Nit: Rename this to `connectionPollingPeriod`.

src/control/ros/RosTrajectoryExecutor.cpp
```diff
@@ -0,0 +1,290 @@
...
+
+  if (mGoalTimeTolerance <= 0)
+    throw std::invalid_argument("Goal time tolerance must be positive.");
+}
+
+//=============================================================================
+RosTrajectoryExecutor::~RosTrajectoryExecutor()
+{
+  // Do nothing.
+  // TODO: Should we wait for the current trajectory to finish executing?
```
mkoval: @psigen @ClintLiddick Any thoughts here?

src/control/ros/RosTrajectoryExecutor.cpp
```diff
@@ -0,0 +1,290 @@
...
+  }
+
+  // Setup the goal properties.
+  // TODO: Also set the goal properties.
+  control_msgs::FollowJointTrajectoryGoal goal;
+  goal.trajectory.header.stamp = _startTime;
+  goal.goal_time_tolerance = ::ros::Duration(mGoalTimeTolerance);
+
+  // Convert the Aikido trajectory into a ROS JointTrajectory.
+  // TODO: Move this into a helper function.
```
mkoval: We should move this into a helper function. Otherwise, it will be extremely difficult to unit test this function.

src/control/ros/RosTrajectoryExecutor.cpp
```diff
@@ -0,0 +1,290 @@
...
+  util::StepSequence timeSequence{mTimestep, true, 0., _traj->getDuration()};
+  const auto numDofs = mSkeleton->getNumDofs();
+  const auto numWaypoints = timeSequence.getMaxSteps();
+  const auto numDerivatives = std::min<int>(_traj->getNumDerivatives(), 2);
```
mkoval: I am not sure if this is correct - it seems like it should be valid for `numDerivatives` to be `1`, so we can execute a piecewise linear trajectory.

src/control/ros/RosTrajectoryExecutor.cpp
```diff
@@ -0,0 +1,290 @@
...
+  goal.trajectory.points.reserve(numWaypoints);
+  for (const auto timeFromStart : timeSequence)
+  {
+    const auto timeAbsolute = _traj->getStartTime() + timeFromStart;
+    trajectory_msgs::JointTrajectoryPoint waypoint;
+
+    Eigen::VectorXd tangentVector;
+    auto state = space->createState();
+    _traj->evaluate(timeAbsolute, state);
+    space->logMap(state, tangentVector);
```
mkoval: Using `logMap` here only makes sense on `Rn` and `SO2` joints. We may want to add an explicit check to that effect. This is the compliment to the single-DOF restriction we have in #147.

src/control/ros/RosTrajectoryExecutor.cpp
```diff
@@ -0,0 +1,290 @@
...
+  // This function assumes that mMutex is locked.
+
+  using actionlib::TerminalState;
+  using Result = control_msgs::FollowJointTrajectoryResult;
+
+#if 0
+  ROS_INFO("Transition: CommState = %s TerminalState = %s",
+    _handle.getCommState().toString().c_str(),
+    _handle.getTerminalState().toString().c_str());
+#endif
```
mkoval: Delete dead code.

src/control/ros/RosTrajectoryExecutor.cpp
```diff
@@ -0,0 +1,290 @@
...
+
+    if (isSuccessful)
+    {
+      mPromise.set_value();
+    }
+    else
+    {
+      mPromise.set_exception(
+        std::make_exception_ptr(
+          std::runtime_error(message.str())));
```
mkoval: We should use a custom `std::runtime_error` subclass here. Otherwise, it will be difficult to distinguish between "expected" (e.g. trajectory was aborted) and "unexpected" failures (e.g. an invalid parameter).

src/control/ros/RosTrajectoryExecutor.cpp
```diff
@@ -0,0 +1,290 @@
...
+  mCallbackQueue.callAvailable();
+
+  if (!::ros::ok() && mInProgress)
+  {
+    mPromise.set_exception(
+      std::make_exception_ptr(std::runtime_error("Detected ROS shutdown.")));
+    mInProgress = false;
+  }
+
+  // TODO: Check mGoalTimeTolerance here as a fail-safe.
```
mkoval: I don't know what this means and don't think it is necessary.

include/aikido/control/ros/RosTrajectoryExecutor.hpp
```diff
@@ -0,0 +1,79 @@
...
+
+  virtual ~RosTrajectoryExecutor();
+
+  std::future<void> execute(trajectory::TrajectoryPtr _traj) override;
+
+  std::future<void> execute(
+    trajectory::TrajectoryPtr _traj, const ::ros::Time& _startTime);
+
+  /// Simulates mTraj. To be executed on a separate thread.
+  void spin();
```
mkoval: We should refactor this function into a `spinOnce`-style callback that we can call from the `Executor` implemented in #151. That is potentially important because it lets us consolidate all of our various callbacks, e.g. multiple `RosTrajectoryExecutor`s and `JointStateSubscriber`s, into a single thread.

mkoval: Nevermind! It turns out that `spin()` is already implemented that way. 😅

Maybe we should rename this to something more obvious? Maybe `step()` or `tick()`?

include/aikido/control/ros/RosTrajectoryExecutor.hpp
jslee02: Not necessary. DART switched to the `DART_DEPRECATED` since version 6.0.1.

src/control/ros/RosTrajectoryExecutor.cpp
gilwoolee: Are we supporting anything other than `Rn` and `SO2` joints? I can add a type-check before the whole conversion happens.

mkoval: I don't think so. At least, I can't think of any other joints that admit a *unique* parameterization as a single real number. Anything else is difficult to shoehorn into the `JointTrajectory` message.

src/control/ros/RosTrajectoryExecutor.cpp
mkoval: Good point. 👍

src/control/ros/RosTrajectoryExecutor.cpp
```diff
@@ -0,0 +1,290 @@
...
+    {
+      std::stringstream msg;
+      msg << "Trajectory is missing DegreeOfFreedom '" << dof->getName()
+          << "' for Skeleton '" << dof->getSkeleton()->getName() << "'.";
+      throw std::invalid_argument(msg.str());
+    }
+  }
+
+  // Setup the goal properties.
+  // TODO: Also set the goal properties.
```
mkoval: We should _eventually_ add support for populating the `goal_tolerance`, `goal_time_tolerance`, and `path_tolerance` fields in the `control_msgs/FollowJointTrajectory` message.

include/aikido/control/ros/RosTrajectoryExecutor.hpp
```diff
@@ -0,0 +1,79 @@
...
+
+namespace aikido {
+namespace control {
+namespace ros {
+
+class RosTrajectoryExecutor : public aikido::control::TrajectoryExecutor
+{
+public:
+  RosTrajectoryExecutor(
+    ::dart::dynamics::MetaSkeletonPtr skeleton,
```
mkoval: Great point! It seems like we could replace `mSkeleton` with a list of DOFs.

include/aikido/control/ros/RosTrajectoryExecutor.hpp
mkoval: @gilwoolee Can you switch from `DEPRECATED` to `DART_GENERATED` and remove this `#undef`.

src/control/ros/RosTrajectoryExecutor.cpp
```diff
@@ -0,0 +1,290 @@
...
+
+      const auto terminalMessage = terminalState.getText();
+      if (!terminalMessage.empty())
+        message << " (" << terminalMessage << ")";
+
+      isSuccessful = false;
+    }
+    else
+    {
+      message << "Execution failed";
```
mkoval: Believe it or not, this is actually correct! 😈 

There are two classes of errors in this function:

1. The state of the [`actionlib` state machine](http://wiki.ros.org/actionlib/DetailedDescription) is reflected by `getCommState()`. If the goal handle finishes _for any reason_ - even if communication is lost with the action server - the state machine transitions to `CommState::DONE`.
2. Once in `CommState::DONE`, `getTerminalState()` indicates why the goal terminated. This can be `TerminalState::SUCCEEDED` or one of several error conditions.
3. If the goal finished in `TerminalState::SUCCEEDED`, then `getResult()` contains the result message. In the case of `FollowJointTrajectoryResult`, the message contains an additional `error_code`.

To summarize: (1) and (2) are **`actionlib` errors** and (3) is a **`FollowJointTrajectory` error**. It's possible for the `actionlib` call to succeed, but for executing the trajectory to fail (in fact: this is the most common case of failure we observe in practice).

The `else` clause here populates `message` with "Execution failed" to differentiate `actionlib` errors from controller errors. Note that `isSuccessful` is **not** set to `false`, so the `message` will only be used if the `result->error_code != Result::SUCCESSFUL` condition fails below.

Does that make sense? Maybe add a comment or two to explain this? 😅

include/aikido/control/ros/RosTrajectoryExecutor.hpp
gilwoolee: This probably comes from my lack of knowledge in ROS. My question was, can't a `RosTrajectoryExecutor` be very flexible and take trajectories covering different set of dofs? Will `serverName` refer to something specific like arm-controller-server, etc.?

src/control/ros/RosTrajectoryExecutor.cpp
gilwoolee: Thanks!

gilwoolee:
@mkoval I addressed most of your comments, which resulted in two major changes:
  - a new conversion function from `Trajectory` to `trajectory_msgs::JointTrajectory` in `Conversions.hpp`
  - a new `Exception` class that inherits from `std::runtime_error`. I am actually a bit not sure what exactly how this class will be used and what methods we need here. Your comment was:
> We should use a custom std::runtime_error subclass here. Otherwise, it will be difficult to distinguish between "expected" (e.g. trajectory was aborted) and "unexpected" failures (e.g. an invalid parameter).

Could you add to this? Should I make an enum type to classify the errors into `expected` and `unexpected`? 

Unit tests are added for the new conversion method. I still need to write unit tests for `RosTrajectoryExecutor`, but since I've already made some big changes, it'd be great if you could comment on them while I work on the tests.

@jslee02 could you take a look too? Thanks a lot!

src/control/ros/Conversions.cpp
```diff
@@ -0,0 +1,438 @@
...
+    trajectory->addSegment(segmentCoefficients, segmentDuration, currState);
+
+    // Advance to the next segment.
+    currPosition = nextPosition;
+    currVelocity = nextVelocity;
+    currAcceleration = nextAcceleration;
+    currTimeFromStart = nextTimeFromStart;
+  }
+
+  return std::move(trajectory);
```
jslee02: `std::move( )` is unnecessary. Related [SO answer](http://stackoverflow.com/a/11817902).

src/control/ros/Conversions.cpp
```diff
@@ -0,0 +1,438 @@
...
+    auto rnJoint = std::dynamic_pointer_cast<RnJoint>(space->getSubspace(i));
+    auto so2Joint = std::dynamic_pointer_cast<SO2Joint>(space->getSubspace(i));
+    if (!rnJoint && !so2Joint)
+    {
+      throw std::invalid_argument(
+        "MetaSkeletonStateSpace must contain only RnJonts and SO2Joints.");
+    }
+
+    // For RnJoint, supports only 1D.
+    if (rnJoint && rnJoint->getDimension() > 1)
```
jslee02: I think we need to explicitly check whether the dimension is not one because it can be zero.

src/control/ros/RosTrajectoryExecutor.cpp
```diff
@@ -0,0 +1,249 @@
...
+{
+  // Do nothing.
+  // TODO: Should we wait for the current trajectory to finish executing?
+}
+
+//=============================================================================
+std::future<void> RosTrajectoryExecutor::execute(
+  trajectory::TrajectoryPtr traj)
+{
+  static const ::ros::Time invalidTime;
```
jslee02: Passing invalid time is intentional? I think we could simply pass the current time using `ros::Time::now()`.

include/aikido/control/ros/Conversions.hpp
```diff
@@ -0,0 +1,38 @@
...
+  const std::shared_ptr<
+    aikido::statespace::dart::MetaSkeletonStateSpace>& space,
+  const trajectory_msgs::JointTrajectory& jointTrajectory);
+
+/// Converts Aikido Trajectory to ROS JointTrajectory.
+/// Supports only 1D RnJoints and SO2Joints.
+/// \_param[in] trajectory Aikido trajectory to be converted.
+/// \_param[in] indexMap Mapping between trajectory's joints and ros joints.
+/// \_param[in] timestep Timestep between two consecutive waypoints.
+const trajectory_msgs::JointTrajectory convertTrajectoryToRosTrajectory(
```
jslee02: I would like to suggest patterning the two convert function names. `convertJointTrajectory()` doesn't convey the original and target type in the function name unlike to the latter function name, and `convertTrajectoryToRosTrajectory()` uses only `Trajectory` to refer `JointTrajectory` unlike to the former function name.

We could put both of original and target type name to the function name like `convertJointTrajectoryToRosJointTrajectory()` but this might be too long. So my preference is including only the goal type to the function name so that we can call the correct conversion functions depending on the target type. We don't need to worry about the original type since functions can overload for the argument types, which is the original type, but not for the return type.

**TL;DR**: I suggest renaming the two function names to something like:
```cpp
convertToSplineJointTrajectory()
convertToRosJointTrajectory()

// or
toSplineJointTrajectory()
toRosJointTrajectory()
```

include/aikido/control/ros/RosTrajectoryExecutionException.hpp
```diff
@@ -0,0 +1,33 @@
...
+
+/// This class wraps various exceptions that may arise during trajectory
+/// execution over ROS.
+class RosTrajectoryExecutionException: public std::runtime_error
+{
+public:
+
+  RosTrajectoryExecutionException(
+    const std::string& what,
+    actionlib::TerminalState terminalState);
```
mkoval: `TerminalState` is an `enum`, so there is no point in avoiding a copy here.

include/aikido/control/ros/RosTrajectoryExecutor.hpp
```diff
@@ -0,0 +1,95 @@
...
+    statespace::dart::MetaSkeletonStateSpacePtr space,
+    ::ros::NodeHandle node,
+    const std::string& serverName,
+    double timestep,
+    double goalTimeTolerance,
+    const std::map<std::string, size_t>& jointIndexMap,
+    std::chrono::milliseconds connectionTimeout
+      = std::chrono::milliseconds{1000},
+    std::chrono::milliseconds connectionPollingPeriod
+      = std::chrono::milliseconds{20}
```
mkoval: I am generally supportive of this, but I am not sure how to implement it. We store these `duration`s in member variables, so this would require `template`ing the entire class on `DurationA` and `DurationB`. It's possible we could do some type erasure shenaniganz to avoid this, but I don't think it's worth the effort at this point.

include/aikido/control/ros/RosTrajectoryExecutor.hpp
```diff
@@ -0,0 +1,95 @@
...
+  void spin();
+
+private:
+  using TrajectoryActionClient
+    = actionlib::ActionClient<control_msgs::FollowJointTrajectoryAction>;
+  using GoalHandle = TrajectoryActionClient::GoalHandle;
+
+  bool waitForServer();
+
+  void transitionCallback(GoalHandle handle);
```
mkoval: Nope. This function is passed as a callback to `ActionClient::sendGoal`, which expects this signature.

src/control/ros/Conversions.cpp
```diff
@@ -0,0 +1,438 @@
...
+  reorder(indexMap, tangentVector, &jointTangentVector);
+
+  waypoint.time_from_start = ::ros::Duration(timeFromStart);
+  waypoint.positions.assign(jointTangentVector.data(),
+    jointTangentVector.data() + jointTangentVector.size());
+
+  assert(0 <= numDerivatives && numDerivatives <= 2);
+  const std::array<std::vector<double>*, 2> derivatives{
+    &waypoint.velocities, &waypoint.accelerations};
+  for (int iderivative = 1; iderivative <= numDerivatives; ++iderivative)
```
mkoval: This is the convention I typically use for naming loop variables, so I suspect you're going to have a lot more comments like this on the other `ros_control` pull requests. 😅 

I generally like this convention, but I am fine with @gilwoolee renaming the variables if you insist... 😞

src/control/ros/Conversions.cpp
```diff
@@ -0,0 +1,438 @@
...
+      jointTangentVector.data() + jointTangentVector.size());
+  }
+}
+
+//=============================================================================
+// The rows of inVector is reordered in outVector.
+void reorder(std::map<size_t, size_t> indexMap,
+  const Eigen::VectorXd& inVector, Eigen::VectorXd* outVector)
+{
+  *outVector = Eigen::VectorXd::Zero(inVector.size());
```
mkoval: This is a helper function in an anonymous namespace, so I suggest just adding an `assert`.

src/control/ros/RosTrajectoryExecutor.cpp
mkoval: Passing an invalid time **is** intentional. Passing `ros::Time::now()` computes the timestamp in this process, whereas passing `ros::Time()` causes the controller to compute the timestamp when it receives the message. 

This difference can be significant if the client and server are running on different computers.

include/aikido/control/ros/RosTrajectoryExecutor.hpp
jslee02: We actually don't need to templatize the whole class but only this constructor. The member variable can be any specific duration type. Here is an example what I did: [member variable](https://github.com/personalrobotics/aikido/blob/master/include/aikido/util/ExecutorThread.hpp#L57), [templated constructor](https://github.com/personalrobotics/aikido/blob/master/include/aikido/util/detail/ExecutorThread-impl.hpp#L7-L11).

PR #150, opened 2017-04-04
ros_control/RosJointStateClient

gilwoolee:
Need to review, document, and write unit tests. 

(There seems to be some issue with build- I can only compile when I comment out tests in the root CMakeLists.)


include/aikido/control/ros/RosJointStateClient.hpp
```diff
@@ -0,0 +1,55 @@
...
+  RosJointStateClient(
+    dart::dynamics::SkeletonPtr _skeleton,
+    ::ros::NodeHandle _nodeHandle,
+    const std::string& _topicName,
+    size_t capacity);
+
+  void spin();
+
+  Eigen::VectorXd getLatestPosition(
+    const dart::dynamics::MetaSkeleton& _metaSkeleton) const;
```
mkoval: We should add a second functions as well:
```c++
Eigen::VectorXd getPositionAtTime(const MetaSkeleton&, const ros::Time&, bool)
```
that interpolates position at the specified time - optionally blocking for new data. This would eliminate a *huge* class of race conditions that we see in OpenRAVE.

However, this is purely an addition. So we can put this off for a follow-up pull request.

src/control/ros/RosJointStateClient.cpp
```diff
@@ -0,0 +1,102 @@
...
+  for (size_t i = 0; i < _jointState.name.size(); ++i)
+  {
+    const auto result = mBuffer.emplace(_jointState.name[i],
+      boost::circular_buffer<JointStateRecord>{mCapacity});
+    auto& buffer = result.first->second;
+
+    JointStateRecord record;
+    record.mStamp = _jointState.header.stamp;
+    record.mPosition = _jointState.position[i];
+    buffer.push_back(record);
```
mkoval: Check that the data in `buffer` is sequential, i.e. we did not receive `JointState` messages out of order. It's probably safe to ignore the message (and print a warning!) if this occurs.

brianhou:
From talking with @gilwoolee it seems that writing tests for this would involve setting up some more testing infrastructure and looking into `rostest`. Since I don't have a ton of time to look into this and it seems to be blocking some of our progress, I just added some documentation. Can we merge this?

include/aikido/control/ros/RosJointStateClient.hpp
```diff
@@ -0,0 +1,68 @@
...
+  /// Call all callbacks in mCallbackQueue.
+  void spin();
+
+  /// Returns the last position of each joint.
+  /// \param _metaSkeleton Skeleton to read DOFs from.
+  Eigen::VectorXd getLatestPosition(
+    const dart::dynamics::MetaSkeleton& _metaSkeleton) const;
+
+  // TODO: implement
+  // getPositionAtTime(const MetaSkeleton&, const ros::Time&, bool)
```
gilwoolee: Why do we need this?

src/control/ros/CMakeLists.txt
```diff
@@ -22,12 +35,14 @@ target_link_libraries("${PROJECT_NAME}_control_ros"
...
   ${trajectory_msgs_LIBRARIES}
 )
 
 add_component(${PROJECT_NAME} control_ros)
 add_component_targets(${PROJECT_NAME} control_ros "${PROJECT_NAME}_control_ros")
 add_component_dependencies(${PROJECT_NAME} control_ros control statespace trajectory)
 
-coveralls_add_sources(${sources})
-
+coveralls_add_sources(${sources})
```
gilwoolee: Nit: add back newline

src/control/ros/RosJointStateClient.cpp
```diff
@@ -0,0 +1,106 @@
...
+    auto& buffer = result.first->second;
+
+    JointStateRecord record;
+    record.mStamp = _jointState.header.stamp;
+    record.mPosition = _jointState.position[i];
+
+    // TODO: check that the data in buffer is sequential, i.e. that we did not
+    // receive JointState messages out of order. It's probably safe to ignore
+    // the message (and print a warning!) if this occurs.
+    buffer.push_back(record);
```
gilwoolee: @brianhou Can we implement this? I think we just have to check the last element's `stamp` with `record.mStamp`.

src/control/ros/RosJointStateClient.cpp
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  return position;
+}
+
+//=============================================================================
+void RosJointStateClient::jointStateCallback(
+  const sensor_msgs::JointState& _jointState)
+{
+  // This method assumes that mSkeleton->getMutex() and mMutex are locked.
```
gilwoolee: Why do we not lock here again? Isn't it safe to lock a locked mutex again?

include/aikido/control/ros/RosJointStateClient.hpp
```diff
@@ -0,0 +1,68 @@
...
+namespace control {
+namespace ros {
+
+class RosJointStateClient
+{
+public:
+  /// Constructor.
+  /// \param _skeleton Skeleton to read JointState updates for.
+  /// \param _nodeHandle ROS node.
+  /// \param _topicName Name of topic to subscribe to for JointState updates.
```
mkoval: Nit: Missing docstring for `capacity`.

include/aikido/control/ros/RosJointStateClient.hpp
```diff
@@ -0,0 +1,68 @@
...
+#include <dart/dynamics/dynamics.hpp>
+#include <ros/ros.h>
+#include <ros/callback_queue.h>
+#include <sensor_msgs/JointState.h>
+
+namespace aikido {
+namespace control {
+namespace ros {
+
+class RosJointStateClient
```
mkoval: Missing docstring.

include/aikido/control/ros/RosJointStateClient.hpp
```diff
@@ -0,0 +1,68 @@
...
+  /// \param _skeleton Skeleton to read JointState updates for.
+  /// \param _nodeHandle ROS node.
+  /// \param _topicName Name of topic to subscribe to for JointState updates.
+  RosJointStateClient(
+    dart::dynamics::SkeletonPtr _skeleton,
+    ::ros::NodeHandle _nodeHandle,
+    const std::string& _topicName,
+    size_t capacity);
+
+  /// Call all callbacks in mCallbackQueue.
```
mkoval: This docstring is not useful. 😉

include/aikido/control/ros/RosJointStateClient.hpp
```diff
@@ -0,0 +1,68 @@
...
+  void spin();
+
+  /// Returns the last position of each joint.
+  /// \param _metaSkeleton Skeleton to read DOFs from.
```
mkoval: - Docstring doesn't specify "of each joint **in `_metaSkeleton`**.
- Nit: "last" -> "most recent"
- Missing `\return` docstring.

include/aikido/control/ros/RosJointStateClient.hpp
mkoval: This is useful for two reasons
- If multiple nodes are publishing `JointState` messages on the same topic, then the elements of `getLatestPosition()` may have been populated at different times. This can lead to some strange artifacts, e.g. where the end-effector pose oscillates if you execute a synchronized arm + base trajectory.
- If you want to synchronize the configuration of the robot to an external sensor (e.g. a camera or RGBD sensor), there may be considerable latency between the sensor capturing data and the data reaching this process. This is problematic when implementing algorithms like a [self filter](http://wiki.ros.org/robot_self_filter#Robot_Self_Filtering).

src/control/ros/RosJointStateClient.cpp
```diff
@@ -0,0 +1,106 @@
...
+//=============================================================================
+RosJointStateClient::RosJointStateClient(
+      dart::dynamics::SkeletonPtr _skeleton,
+      ::ros::NodeHandle _nodeHandle,
+      const std::string& _topicName,
+      size_t capacity)
+  : mSkeleton{std::move(_skeleton)}
+  , mBuffer{}
+  , mCapacity{capacity}
+  , mCallbackQueue{} // Must be before mNodeHandle for order of destruction.
```
mkoval: Nit: I believe this comment should say **after**, which is both correct _and_ what is currently implemented. 😉

src/control/ros/RosJointStateClient.cpp
mkoval: We don't lock here because `spin` acquires the mutex once before calling `callAvailable`. This eliminates the considerable overhead of locking the mutex multiple times if there are a large number of messages queued.

It also means that `spin()` is atomic - the return value of `getLatestPosition()` will _only_ change when `spin()` returns. This is not a huge benefit, but (IMO) the fewer concurrent changes that are happening the easier things are to debug.

> Isn't it safe to lock a locked mutex again?

Only recursive mutexes, which we are **not** using here. In my opinion (which is concisely summarized [here](https://cppwisdom.quora.com/Recursive-mutexes-considered-harmful)), recursive mutexes are generally a bad idea because they mask subtle race conditions.

As [the engineer who implemented the POSIX recursive mutex puts it](https://groups.google.com/forum/?hl=en#!topic/comp.programming.threads/tcrTKnfP8HI%5B51-75%5D):

>Really, really. Recursive mutexes are really bad and they really do hide serious bugs.
> [...]
> But if that's all that's necessary, why does POSIX have recursive mutexes? Because of a dare.

😈

brianhou:
I think I've addressed the most recent issues raised by @mkoval and implemented the out-of-order message ignoring.

However, I'm actually having the same build issue as Gilwoo. I'm not familiar with why this might be happening -- @jslee02 would you mind looking at this?

I needed to make two changes to allow aikido to compile:

- `tests/CMakeLists.txt` causes an error with `add_subdirectory(gtest)`. Commenting out that line (line 2) causes no errors. The message says

```
  add_library cannot create target "gtest" because another target with the
  same name already exists.  The existing target is a shared library created
  in source directory "/usr/src/gtest".  See documentation for policy CMP0002
  for more details.
```

- `CMakeLists.txt` causes an error with `add_custom_target(tests DEPENDS ${all_tests})` and `add_ custom_target(run_tests COMMAND "${CMAKE_CTEST_COMMAND})")`. The messages say

```
  add_custom_target cannot create target "tests" because another target with
  the same name already exists.  The existing target is a custom target
  created in source directory
  "/home/bhou1/Development/aikido-ws/src/aikido/src/control/ros".  See
  documentation for policy CMP0002 for more details.

  add_custom_target cannot create target "run_tests" because another target
  with the same name already exists.  The existing target is a custom target
  created in source directory
  "/home/bhou1/Development/aikido-ws/src/aikido/src/control/ros".  See
  documentation for policy CMP0002 for more details.
```

Looking in `src/control/ros/CMakeLists.txt` doesn't show any obvious suspects, but commenting out those lines (lines 177-178) also allows it to build.

The master branch has all of those lines and builds just fine on my machine.

jslee02:
That is because some of ROS packages's target includes `gtest` and `tests`, and they conflicts with the `gtest` built internally in Aikido. [This](https://github.com/personalrobotics/aikido/pull/149/commits/5f0f1aea579d4386335a946f4caa55c5552eac9f) would a solution.

brianhou:
I see, thanks for clearing that up! In that case, it seems that we should merge #149 before this one?

jslee02:
> it seems that we should merge #149 before this one?

I believe so, and it seems already in good shape to merge to me. :smiley:

include/aikido/control/ros/RosJointStateClient.hpp
```diff
@@ -0,0 +1,72 @@
...
+    const dart::dynamics::MetaSkeleton& _metaSkeleton) const;
+
+  // TODO: implement
+  // getPositionAtTime(const MetaSkeleton&, const ros::Time&, bool)
+  // that interpolates position at the specified time, optionally blocking for
+  // new data.
+private:
+  struct JointStateRecord
+  {
+    inline bool isValid() const { return mStamp.isValid(); }
```
jslee02: Nit: `inline` is meaningless when the function is defined in a header file.

mkoval: From my understanding, this is not true. Do you have a citation for that? :neckbeard: 

The `inline` keyword is a _suggestion_ that the compiler should inline the function that it can choose to honor or ignore, depending upon what it's own optimizer thinks. It's true that GCC and Clang both aggressively inline functions that are available in the current translation unit (e.g. those that are defined in headers), but the `inline` keyword may still have an effect.

PR #151, opened 2017-04-04
ros_control/Executor

gilwoolee:
Needs documentation, review, and unit-tests.


src/util/ExecutorThread.cpp
```diff
@@ -0,0 +1,60 @@
...
+  while (mIsRunning.load())
+  {
+    try
+    {
+      mCallback();
+    }
+    catch (const std::exception& e)
+    {
+      std::cerr << "Exception thrown by callback: " << e.what() << std::endl;
+      mIsRunning.store(false);
```
mkoval: Nit: add a `break` here so we don't `sleep_until` after encountering an error.

src/util/ExecutorThread.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+  while (mIsRunning.load())
+  {
+    try
+    {
+      mCallback();
+    }
+    catch (const std::exception& e)
+    {
+      std::cerr << "Exception thrown by callback: " << e.what() << std::endl;
```
mkoval: Nit: We should find another way to handle this error, so we don't print directly to `std::cerr`. Unfortunately, I don't have an immediately better solution since Aikido doesn't use any particular logging framework.

src/util/ExecutorThread.cpp
jslee02: I don't have much experience in multithreaded programming so this might be a silly question: Shouldn't we throw `e` here instead so that the caller can handle the exception?

jslee02:
Added documentation and tests

src/util/ExecutorThread.cpp
gilwoolee: Yeah. I think we'll have to keep it as  a TODO for now.

gilwoolee:
Looks good! Thanks @jslee02 @mkoval @ClintLiddick for all your previous and current work on this. I'll merge it once Travis passes. :D

include/aikido/util/ExecutorMultiplexer.hpp
```diff
@@ -0,0 +1,50 @@
...
+#define AIKIDO_UTIL_EXECUTORMULTIPLEXER_HPP_
+
+#include <functional>
+#include <mutex>
+#include <vector>
+
+namespace aikido {
+namespace util {
+
+/// ExecutorMultiplexer stores callbacks and calls them in order of they added.
```
mkoval: The fact that callbacks are called in the order they are added isn't important. My original intent for implementing this class was to allow one `ExecutorThread` to call multiple executors. The purpose of `ExecutorThread` is to periodically call "spin" functions, like those in `RosJointStateClient` and `RosTrajectoryExecutor`.

I'd like to capture this "big picture" idea here. Here's my attempt at a new docstring:

> Combine multiple executors (i.e. no argument callbacks) into one executor.
>
> This helper class allows one `ExecutorThread to call multiple executors by sequentially calling the callbacks added to this class.

It is not critical, but I would slightly prefer to remove all of the comments that specify the order the callbacks will be called in. The fact that they are called in order is an implementation detail that is subject to change - and may very well need to change if we add a `removeCallback` helper function to this class.

include/aikido/util/ExecutorThread.hpp
```diff
@@ -0,0 +1,66 @@
...
+
+namespace aikido {
+namespace util {
+
+/// ExecutorThread is a wrapper of std::thread that calls a callback
+/// periodically.
+///
+/// \code
+/// ExecutorThread exec(
+///     []() { std::cout << "running\n";}, std::chrono::nanoseconds(1));
```
mkoval: This is a scary example because printing `running` will likely take longer than 1 ns. Can you change the example to something more realistic, e.g. 10 ms?

include/aikido/util/ExecutorThread.hpp
```diff
@@ -0,0 +1,66 @@
...
+///
+/// // thread is running
+///
+/// exec.stop();
```
mkoval: I suggest removing `stop()` from this example code, since `ExecutorThread` is intended to be used primarily as an RAII class. In fact, I'd like to mention the fact that the destructor stops the thread in this comment.

tests/util/test_Executor.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+//==============================================================================
+TEST(ExecutorThread, ExceptionThrownByCallback)
+{
+  ExecutorThread exec(
+      []() { throw std::exception(); }, std::chrono::milliseconds(1));
+
+  std::this_thread::sleep_for(std::chrono::seconds(3));
+
+  EXPECT_TRUE(!exec.isRunning());
```
mkoval: This is technically a race condition, but I don't see a better way of implementing it. I suppose the 3 second sleep is essentially a timeout for the test.

jslee02:
Thanks for the comments @mkoval . Now I have a better understanding of the classes' purpose.

Let's merge this once Travis CI becomes happy.

PR #153, opened 2017-04-04
Code format: util

jslee02:
**Changes**:
* Updated the style configuration to be close to the C++ style in [`STYLE.md`](https://github.com/personalrobotics/aikido/blob/master/STYLE.md#c-style) as much as possible.
* Applied it to `util`'s header(todo) and source files as the first stage
* Setup CI to check the format

**Reference**:
* [Clang-Format Style Options guide](http://releases.llvm.org/3.8.0/tools/clang/docs/ClangFormatStyleOptions.html)

**TODOs**:
* [x] breaking-constructorInitializers-before-comma
```cpp
// Good!
: mSeq(seq)
, mN(-1)
, mFinalIter(false)

// Bad
: mSeq(seq),
  mN(-1),
  mFinalIter(false)
```
* [x] Install clang-format-3.8 in pr-cleanroom


CMakeLists.txt
```diff
@@ -79,9 +83,14 @@ function(format_add_sources)
         " '${source_abs}'. This should never happen. Did you recently delete"
         " this file or modify 'CMAKE_CURRENT_LIST_DIR'")
     endif()
+    message(STATUS "FILES: ${source_abs}")
```
mkoval: Is this a debug print?

CMakeLists.txt
```diff
@@ -276,7 +285,7 @@ endif()
 #
 find_program(
   CLANG_FORMAT_EXECUTABLE
-  NAMES clang-format clang-format-3.9 clang-format-3.8
+  NAMES clang-format-3.8
```
mkoval: - Why did you remove support for other versions of `clang-format`?
- Could we use `clang-format-${LLVM_VERSION}` to make this slightly more portable?
- Even better, does Clang install a `Config.cmake` file that includes the path to this executable?

include/aikido/util/CatkinResourceRetriever.hpp
```diff
@@ -1,9 +1,9 @@
 #ifndef AIKIDO_UTIL_CATKINRESOURCERETRIEVER_HPP_
 #define AIKIDO_UTIL_CATKINRESOURCERETRIEVER_HPP_
 #include <string>
-#include <unordered_map>
 #include <vector>
 #include <dart/common/ResourceRetriever.hpp>
+#include <unordered_map>
```
mkoval: Nit: I generally like grouping `#include`s into three blocks: (1) standard library, (2) dependencies, (3) includes relative to this project. I find that this helps identify what binary dependencies (e.g. in this case: DART) a particular file has.

Is this something that we can easily do with `clang-format`? Thoughts?

include/aikido/util/KinBodyParser.hpp
```diff
@@ -15,8 +15,10 @@ namespace util {
 /// \code
 /// kinbody - attributes: name
 ///   body - attributes: name
-///     geom - attributes: name, type* (none, box, sphere, trimesh, cylinder), render
-///       data* (or collision) - file* scale [for trimesh] (see below for the detail)
+///     geom - attributes: name, type* (none, box, sphere, trimesh, cylinder),
+///     render
+///       data* (or collision) - file* scale [for trimesh] (see below for the
+///       detail)
```
mkoval: Nasty line wrapping.

include/aikido/util/StepSequence.hpp
```diff
@@ -21,8 +19,11 @@ class StepSequence
   /// point. If includeEndpoints is true then the final point in the sequence
   /// will be the end point, even if it is at less than stepSize from the second
   /// to last point.
-  StepSequence(const double _stepSize, const bool _includeEndpoints = true,
-               const double _startPoint = 0.0, const double _endPoint = 1.0);
+  StepSequence(
+      const double _stepSize,
+      const bool _includeEndpoints = true,
+      const double _startPoint = 0.0,
+      const double _endPoint = 1.0);
```
mkoval: It's meaningless to add a `const` qualifier to parameters that are passed by value in a function declaration. Is this something we can enforce with `clang-format`?

src/util/PseudoInverse.cpp
```diff
@@ -1,17 +1,17 @@
...
-#include <iostream>
 
 namespace aikido {
 namespace util {
 
 Eigen::MatrixXd pseudoinverse(const Eigen::MatrixXd& mat)
 {
 
-  double eps = 1e-6; 
+  double eps = 1e-6;
```
mkoval: Unrelated to this PR: This should be an argument, perhaps with a default value.

src/util/RNG.cpp
```diff
@@ -23,7 +23,7 @@ std::vector<std::unique_ptr<util::RNG>> splitEngine(
   std::vector<util::RNG::result_type> improvedSeeds(_numOutputs);
   seqSeeds.generate(std::begin(improvedSeeds), std::end(improvedSeeds));
 
-  // Create the 
+  // Create the
```
mkoval: Unrelated to this pull request: half-finished comment.

src/util/VanDerCorput.cpp
```diff
@@ -116,10 +131,8 @@ VanDerCorput::const_iterator VanDerCorput::end() const
 }
 
 //=============================================================================
-VanDerCorput::const_iterator::const_iterator(const VanDerCorput *seq)
-    : mSeq(seq)
-    , mN(-1)
-    , mFinalIter(false)
+VanDerCorput::const_iterator::const_iterator(const VanDerCorput* seq)
+  : mSeq(seq), mN(-1), mFinalIter(false)
```
mkoval: I _really_ don't like putting an initialization list on one line. I don't see the difference between this initialization list:
```c++
: mSeq(seq), mN(-1), mFinalIter(false)
```
and this list of assignments:
```c++
mSeq = seq; mN = -1; mFinalIter = false;
```
We wouldn't allow the latter, so why allow the former? 😉 

---

On a slight tangent, I prefer:
```c++
: mSeq(seq)
, mN(-1)
, mFinalIter(false)
```
over
```c++
: mSeq(seq),
  mN(-1),
  mFinalIter(false)
```
because it's easier to re-order lines. I think @psigen strongly prefers the latter, though. 😞

CMakeLists.txt
jslee02: Oh, yeah. This should be removed.

CMakeLists.txt
jslee02: > Why did you remove support for other versions of clang-format?

Well, I observed that they format the code slightly different (even between 3.8 and 3.9) with the same configuration. So I'd like to we use one specific version, then figure out how can we handle this later. I chose ClangFormat 3.8 because it is the latest version supported by trusty.



> Could we use clang-format-${LLVM_VERSION} to make this slightly more portable?

That's an good idea. But still, we should stay with ClangFormat 3.8 until we figure out how to get the same formatted code across different versions. Let me create an issue to track this.

> Even better, does Clang install a Config.cmake file that includes the path to this executable?

[Hm, I don't think so](http://packages.ubuntu.com/trusty-updates/amd64/clang-format-3.8/filelist).

include/aikido/util/CatkinResourceRetriever.hpp
jslee02: I also prefer that way. ClangFormat supports this feature, and I thought this is addressed by [this]. Let me double check.

include/aikido/util/KinBodyParser.hpp
jslee02: Okay, I need to manually wrap it or turn off formatting for this comment block.

include/aikido/util/StepSequence.hpp
jslee02: > Is this something we can enforce with clang-format?

I don't think so. `const` qualifier was added by us in a different PR. Let me remove it anyway.

src/util/PseudoInverse.cpp
jslee02: :+1: Let me change this so since it's just a simple change.

CMakeLists.txt
mkoval: That makes sense. I believe LLVM has a policy of biannual releases with *no guarantee* of backwards compatibility. So it's not surprising that `clang-format` changed significantly between versions. Can you just add a comment to this effect?

src/util/RNG.cpp
jslee02: It would be good if someone added this line finish it. :smile: Or, I'll try to do my best to guess what was it.

src/util/VanDerCorput.cpp
jslee02: > I really don't like putting an initialization list on one line. I don't see the difference between this initialization list ... this list of assignments...

Maybe the setting is to allow single line if it's possible. I personally prefer this way, but not that strongly. Let me check if we could enforce to list them per line.

> On a slight tangent, I prefer ... because it's easier to re-order lines. I think @psigen strongly prefers the latter, though.

Sorry @mkoval . Here is another one prefers the latter even with the understanding of the convenience of re-ordering. `, mN(-1)` looks not that good in aesthetic. Hm, I don't know how we decide the code style when there are disagreements in preferences each other. :cry:

include/aikido/util/CatkinResourceRetriever.hpp
jslee02: I updated the related option. We can group and prioritize headers using `IncludeCategories` option with regex for each group. @mkoval could you take a look the regex again. I'm sure you know regex much better than me.

src/util/VanDerCorput.cpp
jslee02: ClangFormat (at least version 3.8) doesn't support enforcing initialization list to take each line per item. The only related option is `ConstructorInitializerAllOnOneLineOrOnePerLine`, but it just enforces it only when it doesn't fit on a line.

Regarding where to break line in initialization line, there is an option `BreakConstructorInitializersBeforeComma` exactly what you want. I didn't turn it on though. :smiling_imp: I could get used to breaking-constructor-initializers-before-comma style if I have to, but I'd like to vote on not to do.

jslee02:
At the original post, I added a list of debatable style choices we need to decide.

Feel free to add your opinions anyone interested in.

src/util/VanDerCorput.cpp
psigen: I actually somewhat prefer the former, mostly because the trailing comma issue really bugs me.  I am somewhat ambivalent here as long as we are consistent.

psigen: Also, this is exactly why I prefer to spend time hand formatting the code to pass the **check** rather than spending too much time trying to make the **format** operation idempotent: there are going to be many issues like this where we want something that the formatter won't do, and the other issue with clang 3.9's formatter just churning the code for no reason.

jslee02: Okay, let's turn on `BreakConstructorInitializersBeforeComma` option to enforce putting commas at the beginnings of every initializer.



> Also, this is exactly why I prefer to spend time hand formatting the code to pass the check  rather than spending too much time trying to make the format operation idempotent: there are going to be many issues like this where we want something that the formatter won't do, as well as the other issue with clang 3.9's formatter just churning the code for no reason.

I basically agree with you. Alternatively, we could try other formatter that allows us much more fine tuning, but it would also be time-consuming work to find the best configuration for us. But I also think we don't want to repeat the same work of hand formatting for every PR from various contributors and we could be inconsistent.

Let's give a try auto formatting with clang-format and see if we really need to find other solution. :smile:

PR #154, opened 2017-04-05
Cleanup target include directories and libraries

jslee02:
This PR:
* switches to use `target_include_directories` instead of `include_directories` to convey the dependency headers to the dependent libraries
* changes `utils-urdf` to optional component


src/perception/CMakeLists.txt
```diff
@@ -1,6 +1,12 @@
 #==============================================================================
 # Dependencies
 #
+if(NOT DART_utils-urdf_FOUND)
+  message(STATUS "Looking for DART COMPONENT: dart-utils-urdf - NOT found, "
+      "to use perception, please install dart-utils-urdf")
+  return()
+endif()
```
mkoval: Nit: Why is `COMPONENT` capitalized here?

jslee02: I just did reminding `find_package(DART REQUIRED COMPONENTS ...)`, but no reason to capitalize it. Will change to lowercase.

jslee02:
> It looks like we lost the target_compile_options you added in #114. Was that intentional?

It's intentional. I removed it for targets that have dependencies of other Aikido targets that already includes the compile options not to duplicate the options. Adding the same options multiple times wouldn't hurt anything, but it (not duplicating) would be clearer I think.

mkoval:
> It's intentional. I removed it for targets that have dependencies of other Aikido targets that already includes the compile options not to duplicate the options. Adding the same options multiple times wouldn't hurt anything, but it (not duplicating) would be clearer I think.

I disagree. If the target requires those flags, then it should specify them. Otherwise we'll get into trouble if we ever change the dependency relationship between Aikido components. This is hairy enough to debug already that I'd prefer to not add more implicit constraints. 😬 

@psigen Thoughts?

jslee02:
> I disagree. If the target requires those flags, then it should specify them. Otherwise we'll get into trouble if we ever change the dependency relationship between Aikido components. This is hairy enough to debug already that I'd prefer to not add more implicit constraints.

It makes sense. I don't have a strong preference here so am fine with explicitly listing necessary flags and dependencies. To be clear, we would like to add all the necessary external dependencies to each Aikido targets (using `target_link_libraries()`), right?

jslee02:
Addressed @mkoval 's comment I believe.

PR #156, opened 2017-04-05
adding simulation world env to ApriltagsModule

pengjujin:
Changing ApriltagsModule to take a simulation world env pointer instead of a list of skeletons.


include/aikido/perception/AprilTagsModule.hpp
```diff
@@ -43,8 +43,9 @@ class AprilTagsModule : public PerceptionModule
 	virtual ~AprilTagsModule() = default;
 
 	// Documentation inherited
-	bool detectObjects(std::vector<dart::dynamics::SkeletonPtr>& skeleton_list, ros::Duration timeout = ros::Duration(0.0), ros::Time timestamp=ros::Time(0.0)) override; 
-
+	bool detectObjects(dart::simulation::WorldPtr env, 
+					   ros::Duration timeout = ros::Duration(0.0), 
+					   ros::Time timestamp=ros::Time(0.0)) override;
```
mkoval: Nit: Replace tabs with spaces.

include/aikido/perception/PerceptionModule.hpp
```diff
@@ -20,7 +20,7 @@ class PerceptionModule{
     /// \param[in] timeout The duration up to which to wait for the transform. Returns false if none of the markers get correctly transformed
     /// \param[in] timestamp Only detections more recent than this timestamp will be accepted. A timestamp of 0 greedily takes the first available message, and is the default behaviour. 
     /// \return bool Returns \c false if no detection observed, or if none of the detections has a more recent timestamp than the parameter. Returns \c true otherwise.
-    virtual bool detectObjects(std::vector<dart::dynamics::SkeletonPtr>& skeleton_list, ros::Duration timeout, ros::Time timestamp) = 0;
+    virtual bool detectObjects(dart::simulation::WorldPtr env, ros::Duration timeout, ros::Time timestamp) = 0;
```
mkoval: Nit: Take the `WorldPtr` by `const & to avoid an unnecessary copy.

src/perception/AprilTagsModule.cpp
```diff
@@ -33,15 +33,21 @@ AprilTagsModule::AprilTagsModule( ros::NodeHandle node, std::string markerTopic,
...
 	bool any_valid = false;
 	tf::TransformListener listener(mNode);
 	//Looks at all detected tags, looks up config file 
-	//Appends new skeletons to skeleton list
+	//Adding new skeletons to the world env
 	visualization_msgs::MarkerArrayConstPtr marker_message
  			= ros::topic::waitForMessage<visualization_msgs::MarkerArray>(mMarkerTopic,mNode,timeout);
 
+ 	// Making sure the Apriltag Message is non empty
+ 	if(marker_message == NULL){
```
mkoval: Nit: `NULL` -> `nullptr`.

src/perception/YamlAprilTagsDatabase.cpp
```diff
@@ -3,6 +3,8 @@
 #include <aikido/perception/YamlAprilTagsDatabase.hpp>
 #include <yaml-cpp/exceptions.h> 
 
+using namespace std;
```
mkoval: Do not use `using namespace`.

include/aikido/perception/PerceptionModule.hpp
jslee02: This might not be directly related to this PR, but it seems we have two styles for taking `std::shared_ptr`: (1) const reference and (2) copy and move (don't know if there is a better expression for this). Style (2) is used in some constructors and creators. (e.g., [this](https://github.com/personalrobotics/aikido/blob/master/src/distance/Weighted.cpp#L8-L10) and [this](https://github.com/personalrobotics/aikido/blob/master/include/aikido/distance/detail/defaults-impl.hpp#L27-L29)) In Style (2), we pass `std::shared_ptr` by value and move it to the destination.

For those cases (copying `std::shared_ptr` to the destination), it would be the same at the point of performance view ((1) copies it when assigning to the destination, (2) copies it when passing to the argument). But for the above function, (1) is the better way since we can avoid an unnecessary copy. So I would like to suggest consistently using Style (1).

Also, we could consider [providing an additional function that takes rvalue reference if necessary](http://stackoverflow.com/a/20909578).

jslee02:
> There are a number of formatting issues in this pull request. Could you consult the style guide and/or talk to @jslee02 to make the necessary changes?

Alternatively, we could resolve all the style issues as a wholesale using ClangFormat. [This](https://github.com/personalrobotics/aikido/pull/153) is the start!

include/aikido/perception/AprilTagsModule.hpp
jslee02: Will be fixed by #168

PR #157, opened 2017-04-05
Improve readability of .travis.yml

jslee02:


jslee02:
> But I don't like the extra spaces at the start of list items. This doesn't match the style I've seen in most YAML files.

Hm, I thought adding indentation is the convention: [here](http://www.yaml.org/spec/1.2/spec.html), [here](https://en.wikipedia.org/wiki/YAML). If there is the strict convention there, I would like to follow it!

psigen:
@jslee02 is correct that most YAML linters and the documentation suggests that indenting before the `-` in lists is the convention.

However, annoyingly, `.travis.yml` autogeneration (i.e. running something like `travis setup releases` from their CLI) seems to follow the convention that these items are _not_ indented.  I think this is why many `.travis.yml` files end up following this convention.  Both are readable by most parsers.

psigen:
Actually, it looks like even the YAML spec is ambiguous about this, as it has non-indented examples as well.
http://www.yaml.org/spec/1.2/spec.html#id2802802
http://www.yaml.org/spec/1.2/spec.html#id2772312

PR #158, opened 2017-04-05
Fix test_KinematicSimulationTrajectoryExecutor

### LINKED ISSUE #126: test_KinematicSimulationTrajectoryExecutor sometimes fail
It seems the success of `test_KinematicSimulationTrajectoryExecutor` is not deterministic.

jslee02:
`KinematicSimulationTrajectoryExecutorTest.execute_TrajectoryIsAlreadyRunning_Throws` test occasionally fails because the result is non-deterministic. This PR fixes the test to take into account the case as well.

Resolves #126

### DIFF
--- a/tests/control/test_KinematicSimulationTrajectoryExecutor.cpp
+++ b/tests/control/test_KinematicSimulationTrajectoryExecutor.cpp
@@ -155,13 +155,22 @@ TEST_F(KinematicSimulationTrajectoryExecutorTest, execute_TrajectoryIsAlreadyRun
 
   auto future = executor.execute(mTraj);
   
-  // Executor cannot not immediately execute the next one.
-  EXPECT_THROW(executor.execute(mTraj), std::runtime_error);
+  // Executor possibly not able to immediately execute the next one.
+  try
+  {
+    executor.execute(mTraj);
+  }
+  catch (const std::runtime_error& e)
+  {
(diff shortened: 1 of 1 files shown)


gilwoolee:
Oh haha this makes total sense!!! Thanks for looking into this @jslee02 !

tests/control/test_KinematicSimulationTrajectoryExecutor.cpp
```diff
@@ -155,14 +155,23 @@ TEST_F(KinematicSimulationTrajectoryExecutorTest, execute_TrajectoryIsAlreadyRun
...
+  try
+  {
+    executor.execute(mTraj);
+  }
+  catch (const std::runtime_error& e)
+  {
+    EXPECT_EQ(std::string(e.what()), "Another trajectory in execution.");
+
+    // It could not have finished the trajectory execution yet.
+    EXPECT_NE(mSkeleton->getDof(0)->getPosition(), 1.0);
```
gilwoolee: Does this always hold? What if the other thread finished the execution of line 156 by the time we got here?

jslee02: > Does this always hold?

It does not. :sob: I thought detached threads are immediately killed when the main thread throw an exception, which is totally not true. If we add `std::this_thread::sleep_for(std::chrono::seconds(5));` at line 166, this test fails.

Let me remove this line.

PR #159, opened 2017-04-06
Template Rn for dimension to exploit performance of fixed-size Eigen objects

### LINKED ISSUE #152: Consider templating Rn for dimension
I have an impression on `Rn` that it is a state space of dynamic dimension since it interacts with dynamic-size values (i.e., `Eigen::VectorXd`), which is not that case. Eigen generally provides better performance for fixed-size objects over dynamic-size objects.

So I'd like to suggest:
* deprecating `Rn` (if we care about the backward compatibility)
* templatizing `Rn` for the dimension
* renaming it to `R`
* type-aliasing for frequently used dimensions as
```cpp
template <int Dimension>
class R : public virtual StateSpace
{ // ...
};

using R1 = R<1>;
using R2 = R<2>;
using R3 = R<3>;
using R6 = R<6>;
```

jslee02:
**TODOs**:
* [x] Explicitly instantiates for frequently used dimensions
* [x] Use extern template to force the compiler not to instantiate for the explicitly instantiated types
* [x] Decide name: `template <int N> class Rn` vs `template <int N> class R` (I prefer `R<N>`)
* [x] Support dynamic size

Resolves #152


psigen:
I prefer the `R<N>` syntax. I think it makes semantic sense and then instantiations can look like `R2` or `R3`.

include/aikido/constraint/detail/JointStateSpaceHelpers-impl.hpp
```diff
@@ -56,7 +56,11 @@ inline Eigen::VectorXd getPositionUpperLimits(
 
 //=============================================================================
 using JointStateSpaceTypeList = util::type_list<
-  statespace::dart::RnJoint,
+  statespace::dart::R0Joint,
+  statespace::dart::R1Joint,
+  statespace::dart::R2Joint,
+  statespace::dart::R3Joint,
+  statespace::dart::R6Joint,
```
mkoval: Could we add `R<Eigen::Dynamic>` here now? 😁

include/aikido/constraint/uniform/RnBoxConstraint.hpp
```diff
@@ -66,20 +69,28 @@ class RnBoxConstraint
...
+  Vectord mLowerLimits;
+  Vectord mUpperLimits;
 };
 
+using R0BoxConstraint = RnBoxConstraint<0>;
+using R1BoxConstraint = RnBoxConstraint<1>;
+using R2BoxConstraint = RnBoxConstraint<2>;
+using R3BoxConstraint = RnBoxConstraint<3>;
+using R6BoxConstraint = RnBoxConstraint<6>;
+using RxBoxConstraint = RnBoxConstraint<Eigen::Dynamic>;
```
mkoval: Nit: Rename to `RnBoxConstraint`?

include/aikido/constraint/uniform/RnConstantSampler.hpp
```diff
@@ -26,14 +28,23 @@ class RnConstantSampler : public constraint::Sampleable
...
+  std::shared_ptr<statespace::R<N>> mSpace;
+  Vectord mValue;
 };
 
+using R0ConstantSampler = RnConstantSampler<0>;
+using R1ConstantSampler = RnConstantSampler<1>;
+using R2ConstantSampler = RnConstantSampler<2>;
+using R3ConstantSampler = RnConstantSampler<3>;
+using R6ConstantSampler = RnConstantSampler<6>;
+using RxConstantSampler = RnConstantSampler<Eigen::Dynamic>;
```
mkoval: Nit: Rename to `RnConstantSampler`?

include/aikido/constraint/uniform/detail/RnBoxConstraint-impl.hpp
```diff
@@ -0,0 +1,302 @@
...
+{
+  return mSpace;
+}
+
+//=============================================================================
+template <int N>
+bool RnBoxConstraintSampleGenerator<N>::sample(
+  statespace::StateSpace::State* _state)
+{
+  Eigen::VectorXd value(mDistributions.size());
```
mkoval: I think this can be `Eigen::Vector<double, N, 1>`.

include/aikido/distance/RnEuclidean.hpp
```diff
@@ -27,9 +27,19 @@ class RnEuclidean : public DistanceMetric
...
 };
-}
-}
+
+using R0Euclidean = RnEuclidean<0>;
+using R1Euclidean = RnEuclidean<1>;
+using R2Euclidean = RnEuclidean<2>;
+using R3Euclidean = RnEuclidean<3>;
+using R6Euclidean = RnEuclidean<6>;
+using RxEuclidean = RnEuclidean<Eigen::Dynamic>;
```
mkoval: Nit: Rename to `RnEuclidean`?

include/aikido/statespace/Rn.hpp
```diff
@@ -9,33 +9,46 @@ namespace statespace {
...
 
-  /// Constructs a \c _dimension dimensional real vector space.
-  ///
-  /// \param _dimension dimension of the space
-  explicit Rn(int _dimension);
+  /// Constructs a real vector space. If N is Eigen::Dynamic (i.e., -1), this
+  /// state space becomes a dynamic size real vector space; if N is a
+  /// non-negative, then this state space becomes a \c N dimensional fixed size
+  /// real vector space.
+  R();
```
mkoval: This should throw an exception if `N` is `Eigen::Dynamic`, because there is no way to resize it.

include/aikido/statespace/detail/Rn-impl.hpp
```diff
@@ -1,57 +1,275 @@
...
   {
     return this->getStateSpace()->setValue(this->getState(), _value);
   }
 };
 
+//=============================================================================
+template <int N>
+R<N>::R()
+{
+  static_assert(N > -2,
```
mkoval: - `N >= 0 || N == Eigen::Dynamic` would be cleaner, since it doesn't assume `Eigen::Dynamic = -1`.
- Can we add a `static_assert(N != Eigen::Dynamic)` to avoid the edge case below?

include/aikido/constraint/uniform/RnConstantSampler.hpp
jslee02: Which means, we also rename the template class `RnConstantSampler` to `RConstantSampler`, right?

mkoval: Yes, I think that makes sense if we renamed `Rn` to `R`.

include/aikido/statespace/detail/Rn-impl.hpp
jslee02: > N >= 0 || N == Eigen::Dynamic would be cleaner, since it doesn't assume Eigen::Dynamic = -1.

:+1: 



> Can we add a static_assert(N != Eigen::Dynamic) to avoid the edge case below?

I instead changed to throw exceptions for that case because that wouldn't be compiled whenever `N == Eigen::Dynamic`.

include/aikido/constraint/detail/JointStateSpaceHelpers-impl.hpp
jslee02: All the `RnX` classes support `Eigen::Dynamic`, but not for `RnJoint`. This is because all the dimensions of joints are determinant in compile time. More importantly, dynamic size dimension is not supported by none of DART's joint classes.

psigen:
Wait a second, doesn't it make sense to leave the template as Rn? I read it
as `Rn<where n=1>`

jslee02:
`Rn` is reserved for `R<Eigen::Dynamic>` so that it would be the same as before.

So we now have:
* `Rn = R<Eigen::Dynamic>`
* `R0 = R<0>`
* `R2 = R<1>`
* ...

I believe the above is better looking than below:
* `Rnn = Rn<Eigen::Dynamic> // what?!` 
* `Rn0 = Rn<0>`
* `Rn2 = Rn<1>`
* ...

psigen:
I agree with `R<n>` but it looks much weirder with `RConstantSampler<2>`
where it is unclear what word the template parameter is associated with.

jslee02:
I wish C++ support this syntax `R<n>ConstantSampler` (jk, that would be disaster :fearful:).

I believe mostly we might be using type aliasing for the templates like `R2ConstantSampler` so it would be fine with either template names. However, my concern with putting `n` to template names is `using Rnn = Rn<Eigen::Dynamic>`.

Anyways, I don't have a strong preference here as long as it makes sense. :smile:

mkoval:
I slightly prefer `R` for the reason @jslee02 described - the aliases `Rnn`, `RXd` and `Rxd` are all _very_ awkward for the `Eigen::Dynamic` alias. I could be convinced otherwise if @psigen is sufficiently stubborn. :neckbeard:

psigen:
Guys, I am [not](https://github.com/personalrobotics/aikido/pull/159#issuecomment-292544215) [opposed](https://github.com/personalrobotics/aikido/pull/159#issuecomment-293468420) to `R` for the StateSpace, I am opposed to it for types like `RBoxConstraint` where it seems like:
```
using R2BoxConstraint = RBoxConstraint<2>
```
is not clearer than:
```
using R2BoxConstraint = RnBoxConstraint<2>
```
because the template parameter is nowhere near the thing it is attached to.

One option could be to always just put the state space at the end like `BoxConstraintR<2>`.

jslee02:
I thought we all implicitly assume we want to keep the consistency of `R[n]` and `R[n]X` classes.

The alternatives could be options but `RnBoxConstraint<N>` is not consistent with `R<N>`, and `BoxConstraintR<N>` is not consistent with the current naming convention (at least until now) of `[SpaceName]OtherName`.

I think we should choose which one we should weight more between consistency and readability. :sweat:

mkoval:
> I thought we all implicitly assume that we want to keep the consistency of `R[n]` and `R[n]X` classes.

I strongly agree. **All** of the other constraint classes are named like `<StateSpace type><optional descriptor><constraint type>`. Breaking this convention just for `R<n>` will cause me (and I suspect many others) endless frustration.

If it comes down to `R<n>` vs. consistency, I'd rather go with `Rn<n>` with consistent names.

psigen:
I guess I'll just leave it up to you guys, but note that for strong consistency, the naming conventions end up being:
```
RJoint<N>  -> R2Joint
SOJoint<N> -> SO2Joint
SEJoint<N> -> SE3Joint
TJoint<N> -> T2Joint
```

And if you are in a situation where the constraint takes some template parameter
```
RSomeConstraint<N, M> -> R3SomeMConstraint
```

If you think those are acceptable, then by all means go for it.  Personally I think this leads to making it hard to tell how the template parameter is going to affect the substituted name, given that it inserts a number in the middle of the original name in this convention.  But I also don't really have a problem with inconsistent naming between the state spaces and templates that use them, e.g.
```
R<N> -> R2
RnBoxConstraint<N> -> R2BoxConstraint
```

Anyway, it's just a style opinion, so you guys should do whatever you conclude is best.

jslee02:
I think `RSomeConstraint<N, M> -> R3SomeMConstraint` is acceptable.


> Personally I think this leads to making it hard to tell how the template parameter is going to affect the substituted name, given that it inserts a number in the middle of the original name in this convention.

I agree with you, but I also don't know what would be the better alternative under the current convention. I think using different placeholder could be an option as you commented above unless we will support resizable statespace (in this case, we might prefer `X` for the aliasing name rather than the placeholder name), which I believe so:
```cpp
using R2 = Rx<2>
using Rn = Rx<Eigen::Dynamic>

using R2BoxConstraint = RxBoxConstraint<2>
using RnBoxConstraint = RxBoxConstraint<Eigen::Dynamic>

using R2BoxSome3Constraint = RxBoxSomeMConstraint<2, 3>
using RnBoxSome3Constraint = RxBoxSomeMConstraint<Eigen::Dynamic, 3>
```
However, we might have additional style issue if we start to put placeholder to template class name: upper letter vs. lower letter.

> Anyway, it's just a style opinion, so you guys should do whatever you conclude is best.

I agree. I would like to vote to not using placeholder in template class name. :sweat:

jslee02:
Alright, let's merge this as it is since at least we could use it as before (`Rn[X]` classes are the same), and revisit if necessary.

PR #160, opened 2017-04-06
Move test_resources into tests/resources

### LINKED ISSUE #27: Move `test_resources` into `tests/resources`?
It seems like it would make sense to put the contents of  `test_resources` into the existing test directory i.e. `tests/resources`.

There is already `tests/gtest` in there, so it seems reasonable to put other test dependencies in the sample place and avoid another top-level directory.

jslee02:
Resolves #27


mkoval:
@jslee02 Feel free to merge at will.

PR #163, opened 2017-04-10
Update installation instruction of CMake based build

jslee02:
to search ROS dependencies


README.md
```diff
@@ -31,7 +31,7 @@ Once the dependencies are installed, you can build AIKIDO using [CMake]:
 ```shell
 $ mkdir build
 $ cd build
-$ cmake ..
+$ cmake -DCMAKE_PREFIX_PATH=/opt/ros/indigo ..
```
mkoval: I'd rather add a line to the instructions that source `/opt/ros/indigo/setup.bash`, which is the typical way that people use Catkin workspaces. The setup file sets `CMAKE_PREFIX_PATH` as well as potentially a bunch of other variables in the environment hooks. This _may_ work, but it makes me nervous that the build may break in an unpredictable way if it assumes those environment hooks have run.

@psigen Thoughts? Is this an empty fear?

psigen: I agree with this assessment, running the workspace setup script sets many environment variables that may be critical to successful builds, including default CMake build modes, for example.

Note that this _cannot_ just be assumed to be `/opt/ros/indigo/setup.bash` because people might have different versions of dependent libraries in their ROS workspace, and this assumption would put them in a state where the aikido build is linked against libraries they are not expecting.

In any case, it seems like either the ROS workspace's full `setup.bash` should to be sourced ahead of time, or none of the ROS variables should be set.  I think this change should be removed, and an optional line added which is something like `source ROS workspace e.g. /opt/ros/indigo/setup.bash` as one of the first steps.

Is Aikido just always dependent on ROS now?  Why is this not being split into a new section like "Compiling for a ROS workspace"?

mkoval: > Is Aikido just always dependent on ROS now? Why is this not being split into a new section like "Compiling for a ROS workspace"?

No, it should still build without ROS. All of the ROS-dependent components are optional.

jslee02:
Updated according to the comments. Thanks @mkoval and @psigen !

README.md
```diff
@@ -31,10 +31,12 @@ Once the dependencies are installed, you can build AIKIDO using [CMake]:
 ```shell
 $ mkdir build
 $ cd build
+$ source /opt/ros/indigo/setup.bash
```
mkoval: I would change this to something like `source /path/to/my/workspace/setup.bash`. We can add a `# see below` comment and explain that you should source whatever workspace you are currently using. The default is `/opt/ros/indigo`, but @psigen pointed out that this is **not** correct if you've built a newer version of a dependency in a chained workspace.

This is not uncommon if you are stuck using an old version of ROS, but need to upgrade one package.

jslee02: The actual path would be `source /path/to/my/workspace/devel/setup.bash` (`devel` is added), right?

psigen: Yes, also we can move this comment to the very top: there's no reason to create the build directory first, and that way the remaining ROS-agnostic instructions are together.

mkoval: > ...devel is added), right?

No. `/opt/ros/indigo` is an install space by itself. There is no requirement that devel-spaces live in a directory called "devel" or install spaces live in a directory called "install" - those are just the default name of the directories used by `catkin_make` and `catkin build`.

jslee02:
Looks great to me! Merging.

PR #164, opened 2017-04-11
Add wrapper for OMPL path simplifier

aditya-vk:
This introduces a way to simplify and shorten paths obtained from the planners used in AIKIDO using OMPL methods.

AIKIDO Planner -> untimed trajectory -> simplifyOMPL -> simplified and shortened untimed trajectory.

-> The feature has been currently introduced same place as the planning methods although really a post-processing step. Is there a better place for this than here, currently?


include/aikido/planner/ompl/Planner.hpp
```diff
@@ -241,10 +244,45 @@ trajectory::InterpolatedPtr planOMPL(const ::ompl::base::PlannerPtr &_planner,
...
 
+
+/// Take in an aikido trajectory and simplify it using OMPL methods
+/// \param _statespace The StateSpace that the planner must plan within
+/// \param _interpolator An Interpolator defined on the StateSpace. This is used
+/// to interpolate between two points within the space.
+/// \param _dmetric A valid distance metric defined on the StateSpace
+/// \param _sampler A Sampleable that can sample states from the
+/// StateSpace. Warning: Many OMPL planners internally assume this sampler
+/// samples uniformly. Care should be taken when using a non-uniform sampler.
```
mkoval: Minor Nit: I don't know if this is actually true. As far as I know, most randomized algorithms require that the distribution has non-zero probability of sampling all configurations (i.e. have support over all of configuration space). There is nothing special about the uniform distribution. In fact, what it means to be "uniform" depends on your choice of measure.

@siddhss5 Is what I said above true? 😅

include/aikido/planner/ompl/Planner.hpp
```diff
@@ -241,10 +244,45 @@ trajectory::InterpolatedPtr planOMPL(const ::ompl::base::PlannerPtr &_planner,
...
+/// planning. This should include collision checking and any other constraints
+/// that must be satisfied for a state to be considered valid.
+/// \param _boundsConstraint A constraint used to determine whether states
+/// encountered during planning fall within any bounds specified on the
+/// StateSpace. In addition to the _validityConstraint, this must also be
+/// satsified for a state to be considered valid.
+/// \param _boundsProjector A Projectable that projects a state back within
+/// valid bounds defined on the StateSpace
+/// \param _maxPlanTime The maximum time to allow the planner to search for a
+/// solution
```
mkoval: Nit: Specify units (i.e. seconds).

include/aikido/planner/ompl/Planner.hpp
```diff
@@ -241,10 +244,45 @@ trajectory::InterpolatedPtr planOMPL(const ::ompl::base::PlannerPtr &_planner,
...
+/// \param _maxDistanceBtwValidityChecks The maximum distance (under dmetric) between
+/// validity checking two successive points on a tree extension
```
mkoval: Nit: Is "tree extension" meaningful for a shortcutting algorithm?

include/aikido/planner/ompl/Planner.hpp
```diff
@@ -241,10 +244,45 @@ trajectory::InterpolatedPtr planOMPL(const ::ompl::base::PlannerPtr &_planner,
...
+/// needs simplifying.
+trajectory::InterpolatedPtr simplifyOMPL(statespace::StateSpacePtr _stateSpace,
+                                           statespace::InterpolatorPtr _interpolator,
+                                           distance::DistanceMetricPtr _dmetric,
+                                           constraint::SampleablePtr _sampler,
+                                           constraint::TestablePtr _validityConstraint,
+                                           constraint::TestablePtr _boundsConstraint,
+                                           constraint::ProjectablePtr _boundsProjector, 
+                                           double _maxDistanceBtwValidityChecks,
+                                           trajectory::InterpolatedPtr _originalTraj);
```
mkoval: I'd prefer to return an `std::unique_ptr<Interpolated>` rather than a `shared_ptr`. It's easy to convert from a `unique_ptr` to a `shared_ptr`, but the converse is impossible.

src/planner/ompl/Planner.cpp
```diff
@@ -317,6 +317,68 @@ trajectory::InterpolatedPtr planCRRTConnect(
...
+for (size_t idx = 0; idx < _originalTraj->getNumWaypoints(); ++idx) 
+{
+      auto ompl_state = sspace->allocState(_originalTraj->getWaypoint(idx));
+      path->append(ompl_state);
+      sspace->freeState(ompl_state);
+}
+
+// Step 3: Use the OMPL methods to simplify the path
+::ompl::geometric::PathSimplifierPtr simplifier;
+simplifier.reset(new ::ompl::geometric::PathSimplifier(si));
```
mkoval: It is supfluous to use a `shared_ptr` here. Just declare a local variable with type `PathSimplifier`.

src/planner/ompl/Planner.cpp
```diff
@@ -317,6 +317,68 @@ trajectory::InterpolatedPtr planCRRTConnect(
...
+bool const shortened = simplifier->shortcutPath(*path, 1, 1, 1.0, 0.005);
```
mkoval: Make `maxSteps`, `maxEmptySteps`, `rangeRatio`, and `snapToVertex` parameters of this function.

Also, note that the current hard-coded settings are (in my opinion) nonsensical. Setting `maxSteps = 1` means that it will only try to shortcut two randomly selected points before returning. The default values in OMPL are a more reasonable choice, possibly with the modification that `rangeRatio = 1.0` and `snapToVertex = 0.0`.

Another option is to leave `maxSteps = 1`, but call `shortcutPath` in a loop until a time limit (or other termination condition) expires. This is the approach I used for `OMPLSimplifier` in `or_ompl`.

src/planner/ompl/Planner.cpp
```diff
@@ -317,6 +317,68 @@ trajectory::InterpolatedPtr planCRRTConnect(
...
+            static_cast<GeometricStateSpace::StateType *>(
+                path->getState(idx));
+        // Arbitrary timing
+        returnTraj->addWaypoint(idx, st->mState);
+  }
+  return returnTraj;  
 }
+else
+{
+  return _originalTraj;
```
mkoval: Returning `_originalPath` here is dangerous because it aliases the input argument. This could very easily cause a bug because the user expects `_originalTraj` to be unmodified.

This is a bit of a moot point because:

1. It's not possible to return `_originalTraj` once we change the return value to an `unique_ptr`.
2. It's safe to return `returnTraj` even if `shortcutPath` returned `false`.

src/planner/ompl/Planner.cpp
```diff
@@ -317,6 +317,68 @@ trajectory::InterpolatedPtr planCRRTConnect(
...
+
+// Step 2: Convert the AIKIDO trajectory to Geometric Path 
+::ompl::geometric::PathGeometric *path = new ::ompl::geometric::PathGeometric(si);
+
+for (size_t idx = 0; idx < _originalTraj->getNumWaypoints(); ++idx) 
+{
+      auto ompl_state = sspace->allocState(_originalTraj->getWaypoint(idx));
+      path->append(ompl_state);
+      sspace->freeState(ompl_state);
+}
```
mkoval: This code is duplicated from `planOMPL`. Please move it to a helper function.

src/planner/ompl/Planner.cpp
```diff
@@ -317,6 +317,68 @@ trajectory::InterpolatedPtr planCRRTConnect(
...
+        std::move(_stateSpace), std::move(_interpolator));
+
+  for (size_t idx = 0; idx < path->getStateCount(); ++idx) 
+  {
+        const auto *st =
+            static_cast<GeometricStateSpace::StateType *>(
+                path->getState(idx));
+        // Arbitrary timing
+        returnTraj->addWaypoint(idx, st->mState);
+  }
```
mkoval: This code should also be moved to a helper function.

include/aikido/planner/ompl/Planner.hpp
aditya-vk: Removed this parameter actually. Forgot to take it out of the comment as well. But will add a similar parameter for the timeout as suggested in your other comments.

src/planner/ompl/Planner.cpp
aditya-vk: `maxEmptySteps` seems to be an intuitive and natural reason for terminating the shortcutting procedure. So shall keep a count of empty steps through iterations and also use a timeout to bound time of shortcutting from top.

aditya-vk:
Note: Only the termination condition addressed. Other changes shall be addressed in the next commit!

include/aikido/planner/ompl/Planner.hpp
siddhss5: Yes! The only requirement is that the sampler densely sample configuration space.

src/planner/ompl/Planner.cpp
```diff
@@ -374,75 +376,78 @@ trajectory::InterpolatedPtr planCRRTConnect(
...
+  ::ompl::geometric::PathGeometric *path = new ::ompl::geometric::PathGeometric(si);
 
-// Step 2: Convert the AIKIDO trajectory to Geometric Path 
-::ompl::geometric::PathGeometric *path = new ::ompl::geometric::PathGeometric(si);
+  for (size_t idx = 0; idx < _originalTraj->getNumWaypoints(); ++idx) 
+  {
+        auto ompl_state = sspace->allocState(_originalTraj->getWaypoint(idx));
+        path->append(ompl_state);
+        sspace->freeState(ompl_state);
+  }
```
mkoval: This code should also be moved to a helper function.

include/aikido/planner/ompl/Planner.hpp
```diff
@@ -296,17 +298,23 @@ trajectory::InterpolatedPtr planOMPL(
...
+                                                        double _maxDistanceBtwValidityChecks,
+                                                        double _timeout, size_t _maxEmptySteps,
+                                                        trajectory::InterpolatedPtr _originalTraj);
+
+
+namespace{
+    std::unique_ptr<trajectory::Interpolated> aikido2ompl(::ompl::geometric::PathGeometric *path, 
+                                                    statespace::StateSpacePtr _stateSpace,
+                                                    statespace::InterpolatorPtr _interpolator);
+} // namespace
```
mkoval: - Anonymous `namespace`s should [almost](http://stackoverflow.com/a/358823) never be used in header files. This is a fine function that we could promote to a new `planning/ompl/Trajectory.hpp` file.
- Rename this function to match our other conversion functions, e.g. added in #147. I suggest`toInterpolatedTrajectory` for this function and `toOmplPathGeometric` for the inverse. @jslee02 Objections?
- Take the `PathGeometric` argument by `const &` instead of non-`const` pointer, unless there is a strong reason to do otherwise.

src/planner/ompl/Planner.cpp
```diff
@@ -374,75 +376,78 @@ trajectory::InterpolatedPtr planCRRTConnect(
...
 
-// Set the parameters for termination of simplification process
-::ompl::time::point const time_before = ::ompl::time::now();
-::ompl::time::point time_current;
-::ompl::time::duration const time_limit = ::ompl::time::seconds(_timeout);
-double empty_steps = 0;
+  // Set the parameters for termination of simplification process
+  ::ompl::time::point const time_before = ::ompl::time::now();
+  ::ompl::time::point time_current;
+  ::ompl::time::duration const time_limit = ::ompl::time::seconds(_timeout);
```
mkoval: Nit: I would slightly prefer to use `std::chrono` here, instead of the `ompl::time` wrapper for it.

src/planner/ompl/Planner.cpp
```diff
@@ -374,75 +376,78 @@ trajectory::InterpolatedPtr planCRRTConnect(
...
-      const auto *st =
-          static_cast<GeometricStateSpace::StateType *>(
-              path->getState(idx));
-      // Arbitrary timing
-      returnTraj->addWaypoint(idx, st->mState);
-}
-return returnTraj;  
+  std::pair <std::unique_ptr<trajectory::Interpolated>, bool> returnPair;
+  returnPair = std::make_pair(std::move(returnTraj),shorten_success); 
+  return returnPair;
```
mkoval: Nit: Why not just `return std::make_pair(std::move(returnTraj), shorten_success)`?

src/planner/ompl/Planner.cpp
```diff
@@ -372,6 +374,81 @@ trajectory::InterpolatedPtr planCRRTConnect(
...
+        _stateSpace, _interpolator, std::move(_dmetric), std::move(_sampler),
+        std::move(_validityConstraint), std::move(_boundsConstraint),
+        std::move(_boundsProjector), _maxDistanceBtwValidityChecks);
+
+  // Get the state space
+  auto sspace = ompl_static_pointer_cast<GeometricStateSpace>(
+        si->getStateSpace());
+
+  // Step 2: Convert the AIKIDO trajectory to Geometric Path 
+  ::ompl::geometric::PathGeometric *path = new ::ompl::geometric::PathGeometric(si);
```
mkoval: This leaks memory. I don't think `path` needs to be heap allocated, so you could replace this with just:
```c++
::ompl::geometric::PathGeometric path { si };
```

src/planner/ompl/Planner.cpp
```diff
@@ -372,6 +374,81 @@ trajectory::InterpolatedPtr planCRRTConnect(
...
+  returnPair = std::make_pair(std::move(returnTraj),shorten_success); 
+  return returnPair;
+
 }
-}
-}
+//=============================================================================
+} // ns aikido
+} // ns planner
+} // ns ompl
```
mkoval: Add a unit tests for `simplifyOMPL`. The obvious tests are: (1) each error condition, (2) shortcut a two-waypoint trajectory for a large number of iterations and verify that it doesn't change, and (3) shortcut a three waypoint trajectory for a large number of iterations and verify that the trajectory gets shorter.

It would also be nice to add tests for the `GeometricPath` conversion functions, but I feel too guilty to ask about that since I am the one who asked you to implement them. 😬

src/planner/ompl/Planner.cpp
aditya-vk: Haha yes :sweat_smile:

include/aikido/planner/ompl/Planner.hpp
aditya-vk: Oops. Did not realize I left this here. Was just playing around with the helper function. Didn't want to commit that yet. Shall make the changes suggested :)

src/planner/ompl/Planner.cpp
aditya-vk: Oh that's fine! I shall work on it!

aditya-vk:
ToDo on this PR: 
1. Add unit tests for simplifyOMPL
2. Move repeating blocks of code to helper functions
3. Add tests for helper functions

aditya-vk:
ToDo on this PR:
1. Unit Tests for `simplifyOMPL`
2. Tests for helper functions `toInterpolatedTrajectory` and `toOMPLTrajectory`

include/aikido/planner/ompl/Planner.hpp
aditya-vk: I left the helper functions within the file for now. Do you suggest I move them to a new `planner/ompl/Trajectory.hpp` file?

aditya-vk:
For the unit tests to be verified, this needs a fix at #178 
@jslee02 is looking at the issue

jslee02:
> For the unit tests to be verified, this needs a fix at #178
> @jslee02 is looking at the issue

I believe #178 addresses the issue. Please merge master branch into this to enable the unit testing  **once** #178 is merged into master branch.

src/planner/ompl/Planner.cpp
```diff
@@ -372,6 +374,100 @@ trajectory::InterpolatedPtr planCRRTConnect(
...
+  std::chrono::system_clock::time_point time_current;
+  std::chrono::duration<double> const time_limit = std::chrono::duration<double>(_timeout); 
+  double empty_steps = 0;
+  
+  do 
+  {
+
+    bool const shortened = simplifier.shortcutPath(path, 1, _maxEmptySteps, 1.0, 0.0);
+    empty_steps = (empty_steps + !shortened)*!shortened; // Increment only if shorten fails consecutively
+    time_current = ::ompl::time::now();
```
jslee02: This line causes a [compilation error](https://travis-ci.org/personalrobotics/aikido/builds/223049697#L2624). It should be replaced to `time_current = std::chrono::system_clock::now();`.

src/planner/ompl/Planner.cpp
```diff
@@ -372,6 +374,100 @@ trajectory::InterpolatedPtr planCRRTConnect(
...
+        // Arbitrary timing
+        returnInterpolated->addWaypoint(idx, st->mState);
+  }
+  return returnInterpolated;
 }
+//=============================================================================
+
+} // ns aikido
+} // ns planner
+} // ns ompl
```
jslee02: Nit: The order of namespace is inverse. It should be:
```cpp
} // namespace ompl
} // namespace planner
} // namespace aikido
```

src/planner/ompl/Planner.cpp
aditya-vk: :sweat_smile: 
Fixed it!

jslee02:
It seems all the unit tests pass but style check. Could you format the code by [building `format` target](https://github.com/personalrobotics/aikido/blob/master/STYLE.md#autoformatting-using-clangformat)?

src/planner/ompl/Planner.cpp
```diff
@@ -372,6 +374,114 @@ trajectory::InterpolatedPtr planCRRTConnect(
...
+      = std::chrono::duration<double>(_timeout);
+  double empty_steps = 0;
+
+  do
+  {
+
+    bool const shortened
+        = simplifier.shortcutPath(path, 1, _maxEmptySteps, 1.0, 0.0);
+    empty_steps = (empty_steps + !shortened)
+                  * !shortened; // Increment only if shorten fails consecutively
```
mkoval: Nit: This could be replaced with simply `empty_steps += !!shortened;`.

src/planner/ompl/Planner.cpp
```diff
@@ -372,6 +374,114 @@ trajectory::InterpolatedPtr planCRRTConnect(
...
+// use
+
+::ompl::geometric::PathGeometric toOMPLTrajectory(
+    trajectory::InterpolatedPtr _interpolatedTraj,
+    ::ompl::base::SpaceInformationPtr _si)
+{
+  auto sspace
+      = ompl_static_pointer_cast<GeometricStateSpace>(_si->getStateSpace());
+
+  ::ompl::geometric::PathGeometric returnPath{_si};
```
mkoval: Nit: `std::move(_si)`.

src/planner/ompl/Planner.cpp
```diff
@@ -372,6 +374,114 @@ trajectory::InterpolatedPtr planCRRTConnect(
...
+  std::pair<std::unique_ptr<trajectory::Interpolated>, bool> returnPair;
+  return std::make_pair(std::move(returnTraj), shorten_success);
 }
+//=============================================================================
+
+// Following are helper functions. Might want to move them elsewhere for generic
+// use
+
+::ompl::geometric::PathGeometric toOMPLTrajectory(
+    trajectory::InterpolatedPtr _interpolatedTraj,
```
mkoval: Nit: Take `_interpolatedTraj` by `const &`.

src/planner/ompl/Planner.cpp
```diff
@@ -372,6 +374,114 @@ trajectory::InterpolatedPtr planCRRTConnect(
...
+  {
+    auto ompl_state = sspace->allocState(_interpolatedTraj->getWaypoint(idx));
+    returnPath.append(ompl_state);
+    sspace->freeState(ompl_state);
+  }
+  return returnPath;
 }
+
+std::unique_ptr<trajectory::Interpolated> toInterpolatedTrajectory(
+    ::ompl::geometric::PathGeometric* _path,
```
mkoval: Take `_path` by `const &` instead of by pointer.

tests/planner/ompl/test_OMPLSimplifier.cpp
```diff
@@ -0,0 +1,509 @@
...
+//   auto si = getSpaceInformation(
+//       std::move(stateSpace), std::move(interpolator), std::move(dmetric),
+//       std::move(sampler), std::move(collConstraint),
+//       std::move(boundsConstraint), std::move(boundsProjection), 0.1);
+
+//   auto mvalidator = boost::dynamic_pointer_cast<aikido::planner::ompl::MotionValidator>(
+//       si->getMotionValidator());
+//   EXPECT_FALSE(mvalidator == nullptr);
+// }
+
```
mkoval: Remove this dead code.

tests/planner/ompl/test_OMPLSimplifier.cpp
```diff
@@ -0,0 +1,509 @@
...
+#include <aikido/planner/ompl/CRRT.hpp>
+#include <aikido/planner/ompl/CRRTConnect.hpp>
+#include <ompl/geometric/planners/rrt/RRTConnect.h>
+
+using StateSpace = aikido::statespace::dart::MetaSkeletonStateSpace;
+using aikido::planner::ompl::getSpaceInformation;
+using aikido::planner::ompl::CRRT;
+using aikido::planner::ompl::CRRTConnect;
+
+TEST_F(SimplifierTest, EndPointCheck)
```
mkoval: Rename to `StartAndEndPointsAreUnchanged` (or something similar).

tests/planner/ompl/test_OMPLSimplifier.cpp
```diff
@@ -0,0 +1,509 @@
...
+#include <aikido/planner/ompl/MotionValidator.hpp>
+#include <aikido/planner/ompl/CRRT.hpp>
+#include <aikido/planner/ompl/CRRTConnect.hpp>
+#include <ompl/geometric/planners/rrt/RRTConnect.h>
+
+using StateSpace = aikido::statespace::dart::MetaSkeletonStateSpace;
+using aikido::planner::ompl::getSpaceInformation;
+using aikido::planner::ompl::CRRT;
+using aikido::planner::ompl::CRRTConnect;
+
```
mkoval: Add a test that runs the simplifier on a three-waypoint trajectory for _many_ iterations and verifies that the trajectory gets shorter.

src/planner/ompl/Planner.cpp
aditya-vk: But then it would not be consecutive steps right? I was planning to terminate it based on `maxEmptySteps` only if the shortcutting fails consecutively for those many attempts.

tests/planner/ompl/test_OMPLSimplifier.cpp
```diff
@@ -0,0 +1,102 @@
...
+    simplifiedTraj->evaluate(i+1, s0);
+    r0 = s0.getSubStateHandle<R3>(0);
+    auto nextState = r0.getValue();
+
+    simplifiedTrajDistance += pow(nextState[0] - currentState[0], 2) + pow(nextState[1] - currentState[1], 2) + pow(nextState[2] - currentState[2], 2);
+  }
+  simplifiedTrajDistance = simplifiedTrajDistance / (simplifiedTraj->getDuration());
+
+  EXPECT_TRUE(shorten_success == ( simplifiedTrajDistance < trajDistance ));
```
aditya-vk: @mkoval I have a feeling you won't like the way I did this..does this look fine in making sure the trajectory shortening is in agreement with boolean returned? The distance calculation could be better, any suggestions?

tests/planner/ompl/test_OMPLSimplifier.cpp
aditya-vk: Do you think the current test would suffice, given the start and goal poses are as set?

src/planner/ompl/Planner.cpp
aditya-vk: Added tests for trajectory conversion functions. Converted an interpolated trajectory back and forth and ensured it had not changed. That should suffice, yes?

src/planner/ompl/Planner.cpp
aditya-vk: Done! But what is the advantage?

aditya-vk:
Formatted with `clang` but still Travis fails to build without errors :sweat:

src/planner/ompl/Planner.cpp
mkoval: The current version copies `_si` the `smart_ptr` into the argument, then allows the original `smart_ptr` instance to go out of scope. This requires incrementing, then decrementing the reference counter inside the `smart_ptr`, which is a relatively expensive atomic operation because `smart_ptr`s are thread safe.

`std::move`ing the `smart_ptr` into the argument leaves the reference counter unchanged by coalescing the increment and decrement into one operation.

src/planner/ompl/Planner.cpp
mkoval: I see, you're right. This is clever.

...unfortunately, clever code is generally hard to debug! 😬 I don't feel too strongly about this, but I would prefer if we simply wrote out a longer `if` statement:
```c++
if (shortened)
  empty_steps = 0;
else
  empty_steps += 1;
```

src/planner/ompl/Planner.cpp
```diff
@@ -372,6 +374,114 @@ trajectory::InterpolatedPtr planCRRTConnect(
...
+    distance::DistanceMetricPtr _dmetric,
+    constraint::SampleablePtr _sampler,
+    constraint::TestablePtr _validityConstraint,
+    constraint::TestablePtr _boundsConstraint,
+    constraint::ProjectablePtr _boundsProjector,
+    double _maxDistanceBtwValidityChecks,
+    double _timeout,
+    size_t _maxEmptySteps,
+    trajectory::InterpolatedPtr _originalTraj)
+{
```
mkoval: Nit:
- Check that `_maxEmptyChecks` is non-negative.
- Check that `_timeout` is non-negative.

tests/planner/ompl/test_TrajectoryConversions.cpp
```diff
@@ -0,0 +1,100 @@
...
+      std::move(_collConstraint),
+      std::move(_boundsConstraint),
+      std::move(_boundsProjection),
+      0.1);
+
+  auto omplTraj = aikido::planner::ompl::toOMPLTrajectory(
+      traj, si);
+
+  auto interpolatedTraj = aikido::planner::ompl::toInterpolatedTrajectory(
+    omplTraj, stateSpace, interpolator);
```
mkoval: 👎 Please write separate tests for `toOMPLTrajectory` and `toInterpolatedTrajectory`.

It's important that unit tests be as granular as possible, so it is obvious _where_ the bug resides if a tests fails. We should generally aspire to (but may not always achieve) making a bug obvious from just from the name of the test that failed.

It is generally pretty painless to split tests apart as long as you put the setup code in helper functions.

tests/planner/ompl/test_OMPLSimplifier.cpp
```diff
@@ -0,0 +1,102 @@
...
+  {
+    traj->evaluate(i, stateCurrent);
+    auto rCurrent = stateCurrent.getSubStateHandle<R3>(0);
+    auto currentState = rCurrent.getValue();
+    
+    traj->evaluate(i+1, stateNext);
+    auto rNext = stateNext.getSubStateHandle<R3>(0);
+    auto nextState = rNext.getValue();
+
+    trajDistance += sqrt(pow(nextState[0] - currentState[0], 2) + pow(nextState[1] - currentState[1], 2) + pow(nextState[2] - currentState[2], 2));
```
mkoval: We should use `_dmetric` to compute distance here, since that's what the simplifier is trying to minimize.

tests/planner/ompl/test_OMPLSimplifier.cpp
```diff
@@ -0,0 +1,102 @@
...
+    simplifiedTraj->evaluate(i, stateCurrent);
+    auto rCurrent = stateCurrent.getSubStateHandle<R3>(0);
+    auto currentState = rCurrent.getValue();
+
+    simplifiedTraj->evaluate(i+1, stateNext);
+    auto rNext = stateNext.getSubStateHandle<R3>(0);
+    auto nextState = rNext.getValue();
+
+    simplifiedTrajDistance += sqrt(pow(nextState[0] - currentState[0], 2) + pow(nextState[1] - currentState[1], 2) + pow(nextState[2] - currentState[2], 2));
+  }
```
mkoval: Put this code in a helper function, since it is duplicated twice. We can either leave that function in this file in an anonymous `namespace` _or_ put it in Aikido proper. If you choose the latter, please also write tests for that function. 😁  😈 

As a result, we should add a second test that verifies that the return value is `false` if we pass the function a two-waypoint trajectory.

tests/planner/ompl/test_OMPLSimplifier.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+  EXPECT_TRUE(shorten_success == ( simplifiedTrajDistance < trajDistance ));
```
mkoval: I believe this should _always_ succeed. If you make a three waypoint trajectory with no obstacles in the environment, than any two randomly sampled points along that trajectory can be connected by a straight line.

tests/planner/ompl/test_OMPLSimplifier.cpp
```diff
@@ -0,0 +1,102 @@
...
+  auto r0 = s0.getSubStateHandle<R3>(0);
+  EXPECT_TRUE(r0.getValue().isApprox(startPose));
+
+  // Check the last waypoint
+  simplifiedTraj->evaluate(simplifiedTraj->getDuration(), s0);
+  r0 = s0.getSubStateHandle<R3>(0);
+  EXPECT_TRUE(r0.getValue().isApprox(goalPose));
+
+// Test that the boolean returned is in agreement with shortening process
+//=============================================================================
```
mkoval: Split this into two separate tests, see below.

tests/planner/ompl/test_OMPLSimplifier.cpp
```diff
@@ -0,0 +1,102 @@
...
+using StateSpace = aikido::statespace::dart::MetaSkeletonStateSpace;
+using aikido::planner::ompl::getSpaceInformation;
+using aikido::planner::ompl::CRRT;
+using aikido::planner::ompl::CRRTConnect;
+
+
+TEST_F(SimplifierTest, EndPointCheckShortenVerify)
+{
+  Eigen::Vector3d startPose(-5, -5, 0);
+  Eigen::Vector3d goalPose(5, 5, 0);
```
mkoval: Nit: These are positions, not poses.

tests/planner/ompl/test_OMPLSimplifier.cpp
aditya-vk: Just to be sure. Like say if the two points selected always happened to be on a straight line, then the shortening would not have happened.

tests/planner/ompl/test_OMPLSimplifier.cpp
aditya-vk: Chose the former :P
Yep, added the two waypoint trajectory simplification check.

src/planner/ompl/Planner.cpp
```diff
@@ -372,6 +374,124 @@ trajectory::InterpolatedPtr planCRRTConnect(
...
+  if (_timeout < 0)
+  {
+    throw std::invalid_argument("Timeout must be >= 0");
+  }
+
+  if (_maxEmptySteps < 0)
+  {
+    throw std::invalid_argument("Max empty steps must be >= 0");
+  }
```
jslee02: Please remove this check. This check is pointless since `_maxEmptySteps` is always positive value as its type is unsigned.

tests/planner/ompl/test_OMPLSimplifier.cpp
```diff
@@ -0,0 +1,197 @@
...
+    auto stateCurrent = stateSpace->createState();
+    auto stateNext = stateSpace->createState();
+
+    double trajDistance = 0.0;
+    
+    for(size_t i = 0; i < traj->getDuration(); ++i)
+    {
+      traj->evaluate(i, stateCurrent);
+      auto rCurrent = stateCurrent.getSubStateHandle<R3>(0);
+      auto currentState = rCurrent.getValue();
```
jslee02: `rCurrent` and `currentState` are not used. This will cause the unused-variable warnings.

tests/planner/ompl/test_OMPLSimplifier.cpp
```diff
@@ -0,0 +1,197 @@
...
+      
+      traj->evaluate(i+1, stateNext);
+      auto rNext = stateNext.getSubStateHandle<R3>(0);
+      auto nextState = rNext.getValue();
```
jslee02: `rNext` and `nextState` are not used. This will cause the unused-variable warnings.

tests/planner/ompl/test_OMPLSimplifier.cpp
```diff
@@ -0,0 +1,197 @@
...
+        Eigen::Vector3d(0.1, 0.1, 0.1));
+
+  auto simplifiedPair = aikido::planner::ompl::simplifyOMPL(
+      stateSpace, interpolator, std::move(dmetric), std::move(sampler), 
+      std::move(collConstraint), std::move(boundsConstraint), std::move(boundsProjection),
+      0.1, 5.0, 10, traj);
+
+  // Simplification results
+  auto simplifiedTraj = std::move(simplifiedPair.first);
+  bool shorten_success = simplifiedPair.second;
```
jslee02: I guess you might want to check the success by `EXPECT_TRUE(shorten_success);`.

tests/planner/ompl/test_OMPLSimplifier.cpp
```diff
@@ -0,0 +1,189 @@
...
+  {
+    auto stateCurrent = stateSpace->createState();
+    auto stateNext = stateSpace->createState();
+
+    double trajDistance = 0.0;
+    
+    for(size_t i = 0; i < traj->getDuration(); ++i)
+    {
+      traj->evaluate(i, stateCurrent);   
+      traj->evaluate(i+1, stateNext);
```
mkoval: This evaluates the trajectory past its end. The simplest fix is to change the loop condition to`getNumWaypoints() - 1`.  If you do that, then you need to handle the special case where `getNumWaypoints() == 0` to prevent overflow from `-1` to `UNIT_MAX`.

Another option is to initialize `i = 1` and to change the indicies in the loop to `i - 1` and `i`.

tests/planner/ompl/test_OMPLSimplifier.cpp
```diff
@@ -0,0 +1,189 @@
...
+  }
+
+  aikido::trajectory::InterpolatedPtr planTrajForTest(
+      aikido::statespace::dart::MetaSkeletonStateSpacePtr _stateSpace,
+      aikido::statespace::InterpolatorPtr& _interpolator,
+      aikido::distance::DistanceMetricPtr& _dmetric,
+      aikido::constraint::SampleablePtr& _sampler,
+      aikido::constraint::TestablePtr& _collConstraint,
+      aikido::constraint::TestablePtr& _boundsConstraint,
+      aikido::constraint::ProjectablePtr& _boundsProjection,
```
mkoval: Why are all of these passed by non-const reference? I think it would be fine to pass them by value.

tests/planner/ompl/test_OMPLSimplifier.cpp
```diff
@@ -0,0 +1,189 @@
...
+      Eigen::Vector3d _startPose, Eigen::Vector3d _goalPose
```
mkoval: It [is dangerous](https://eigen.tuxfamily.org/dox/group__TopicPassingByValue.html) to pass `Eigen` types by value because it can introduce alignment issues. Please pass these by `const &`.

tests/planner/ompl/test_OMPLSimplifier.cpp
```diff
@@ -0,0 +1,189 @@
...
+
+  // Original Trajectory
+  auto traj = planTrajForTest(stateSpace, interpolator, dmetric, sampler, collConstraint, 
+      boundsConstraint, boundsProjection, startPose, goalPose);
+
+  // Setup for simplification
+  aikido::distance::DistanceMetricPtr dmetric = aikido::distance::createDistanceMetric(stateSpace);
+  aikido::constraint::SampleablePtr sampler = aikido::constraint::createSampleableBounds(stateSpace, make_rng()); 
+  aikido::constraint::ProjectablePtr boundsProjection = aikido::constraint::createProjectableBounds(stateSpace);
+  aikido::constraint::TestablePtr boundsConstraint = aikido::constraint::createTestableBounds(stateSpace);
```
mkoval: Nit: I suggest making these member variables that are initialized in `SetUp()`. That will eliminate a lot of duplicate code from these tests.

tests/planner/ompl/test_OMPLSimplifier.cpp
```diff
@@ -0,0 +1,189 @@
...
+  aikido::constraint::TestablePtr collConstraint = std::make_shared<MockTranslationalRobotConstraint>(
+        stateSpace, Eigen::Vector3d(-0.1, -0.1, -0.1),
+        Eigen::Vector3d(0.1, 0.1, 0.1));
```
mkoval: Nit: I don't think we need a real constraint for this. It would require less setup to use the `Satisfied` constraint.

tests/planner/ompl/test_OMPLSimplifier.cpp
```diff
@@ -0,0 +1,189 @@
...
+  // Check the first waypoint
+  auto s0 = stateSpace->createState();
+  simplifiedTraj->evaluate(0, s0);
+  auto r0 = s0.getSubStateHandle<R3>(0);
+  EXPECT_TRUE(r0.getValue().isApprox(startPose));
+
+  // Check the last waypoint
+  simplifiedTraj->evaluate(simplifiedTraj->getDuration(), s0);
+  r0 = s0.getSubStateHandle<R3>(0);
+  EXPECT_TRUE(r0.getValue().isApprox(goalPose));
```
mkoval: Replace this with the `EXPECT_EIGEN_EQUAL` macro defined in `eigen_tests.hpp`. It has the same effect, but will print a beautiful error message if there is a mismatch. 😍

The same comment applies to the other `EXPECT_TRUE(... isApprox ... )` calls in these tests.

tests/planner/ompl/test_OMPLSimplifier.cpp
```diff
@@ -0,0 +1,189 @@
...
+
+  // Simplification results
+  auto simplifiedTraj = std::move(simplifiedPair.first);
+  bool shorten_success = simplifiedPair.second;
+
+  aikido::distance::DistanceMetricPtr Ddmetric = aikido::distance::createDistanceMetric(stateSpace);
+  double trajDistance = computeTrajLength<aikido::trajectory::InterpolatedPtr>(traj, stateSpace, Ddmetric);
+  double simplifiedTrajDistance = computeTrajLength<std::unique_ptr<aikido::trajectory::Interpolated>>(simplifiedTraj, stateSpace, Ddmetric);
+
+  EXPECT_TRUE(shorten_success == (simplifiedTrajDistance < trajDistance));
```
mkoval: You raise an excellent point that shortcutting still could fail if we sample two points on the same segment.

Even so, I  would prefer to increase the number of iterations and `EXPECT_TRUE(shorten_success)`. One way of implementing that is to call `simplifyOMPL` in a loop up to `N` times with a generous `_timeLimit` and `_numAttempts = 1`. Adjust `N` until the tests passes with _very high_ probability.

Note that it doesn't take many iterations to effectively guarantee that `shorten_success` will be `true`. If the middle waypoint is halfway along the trajectory, every shortcutting attempt has a 50% probability of reducing the length of the trajectory. The probability that `N` sequential iterations will fail is `(0.5) ^ N`.

tests/planner/ompl/test_TrajectoryConversions.cpp
```diff
@@ -0,0 +1,230 @@
...
+
+    auto goalState = _stateSpace->createState();
+    auto subState2 = _stateSpace->getSubStateHandle<R3>(goalState, 0);
+    subState2.setValue(_goalPose);
+
+    // Plan
+    auto traj = aikido::planner::ompl::planOMPL<ompl::geometric::RRTConnect>(
+        startState, goalState, _stateSpace, _interpolator, std::move(_dmetric),
+        std::move(_sampler), std::move(_collConstraint),
+        std::move(_boundsConstraint), std::move(_boundsProjection), 5.0, 0.1);
```
mkoval: 👎 It is overkill to plan a path using an RRT for testing purposes. It is also undesirable because the result is not deterministic. Please construct a test trajectory by hand.

tests/planner/ompl/test_TrajectoryConversions.cpp
```diff
@@ -0,0 +1,230 @@
...
+      boundsConstraint, boundsProjection, startPose, goalPose);
+
+  // Setup
+  aikido::distance::DistanceMetricPtr _dmetric = aikido::distance::createDistanceMetric(stateSpace);
+  aikido::constraint::SampleablePtr _sampler = aikido::constraint::createSampleableBounds(stateSpace, make_rng()); 
+  aikido::constraint::ProjectablePtr _boundsProjection = aikido::constraint::createProjectableBounds(stateSpace);
+  aikido::constraint::TestablePtr _boundsConstraint = aikido::constraint::createTestableBounds(stateSpace);
+  aikido::constraint::TestablePtr _collConstraint = std::make_shared<MockTranslationalRobotConstraint>(
+        stateSpace, Eigen::Vector3d(-0.1, -0.1, -0.1),
+        Eigen::Vector3d(0.1, 0.1, 0.1));
```
mkoval: Same comment as above about putting these in `setUp()` to minimize code duplication.

jslee02:
> Still missing a check that _maxEmptySteps >= 0.

@aditya-vk added it, but I asked to remove it because it's always positive as the type is `size_t`. Is there any possibility `size_t` can be negative?

mkoval:
@jslee02 good point, I missed that until @aditya-vk pointed it out.

tests/planner/ompl/test_OMPLSimplifier.cpp
```diff
@@ -0,0 +1,189 @@
...
+using aikido::planner::ompl::getSpaceInformation;
+using aikido::planner::ompl::CRRT;
+using aikido::planner::ompl::CRRTConnect;
+
+namespace {
+
+  template <typename T>
+  double computeTrajLength(T& traj,
+      aikido::statespace::dart::MetaSkeletonStateSpacePtr stateSpace,
+      aikido::distance::DistanceMetricPtr& dmetric)
```
aditya-vk: `planOMPL` returns an `std::shared_ptr` while `simplifyOMPL` returns an `std::unique_ptr` `Interpolated` trajectory. Templated it to account for both types. since I will scratch planOMPL completely and make trajectories by hand I guess I can get rid of the templating.

tests/planner/ompl/test_OMPLSimplifier.cpp
```diff
@@ -0,0 +1,189 @@
...
+  {
+    auto stateCurrent = stateSpace->createState();
+    auto stateNext = stateSpace->createState();
+
+    double trajDistance = 0.0;
+    
+    for(size_t i = 0; i < traj->getDuration(); ++i)
```
aditya-vk: Hmm. I was under the impression that we were always obtaining untimed trajectories out of the planner (implying unit timing) which need not be true. You are right. I shall make corrections.

tests/planner/ompl/test_OMPLSimplifier.cpp
aditya-vk: `getDuration()` starts counting from 0. So if there are 9 points in the trajectory we get `getDuration() = 8`. We would hence not be evaluating past the trajectory's end. But I will be moving to `getNumWayPoints()`. If `getNumWayPoints == 0` I will skip the evaluation completely. Maybe an `if-else` statement for handling the special case? 
```
if(getNumWayPoints())
     evaluation()
```

tests/planner/ompl/test_OMPLSimplifier.cpp
aditya-vk: Sure!
But the hope is that there is never a mismatch :smile:

tests/planner/ompl/OMPLTestHelpers.hpp
```diff
@@ -1,6 +1,8 @@
 #ifndef AIKIDO_TESTS_OMPL_OMPLTESTHELPERS_HPP_
 #define AIKIDO_TESTS_OMPL_OMPLTESTHELPERS_HPP_
 
+#include <math.h>
```
aditya-vk: Got it! Removed it for now though. Don't need it. Using `dmetric` for trajectory length calculation now as suggested previously.

tests/planner/ompl/test_OMPLSimplifier.cpp
aditya-vk: I won't need any of these anymore since I will no longer use planOMPL, the ones in `SetUp()` should suffice

src/planner/ompl/Planner.cpp
```diff
@@ -372,6 +374,119 @@ trajectory::InterpolatedPtr planCRRTConnect(
...
+    auto ompl_state = sspace->allocState(_interpolatedTraj->getWaypoint(idx));
+    returnPath.append(ompl_state);
+    sspace->freeState(ompl_state);
+  }
+  return returnPath;
 }
+
+std::unique_ptr<trajectory::Interpolated> toInterpolatedTrajectory(
+    const ::ompl::geometric::PathGeometric& _path,
+    statespace::StateSpacePtr _stateSpace,
```
aditya-vk: Would it not be more expensive to call a function to get `stateSpace` when we already have access to it? I am not really sure though..just curious to know

tests/planner/ompl/test_OMPLSimplifier.cpp
aditya-vk: Left it templated. Even if I construct a trajectory by hand I am forced to make it a shared pointer since simplifyOMPL takes as input aikido::trajectory::Interpolated which is a shared pointer while it outputs a unique pointer. I would like to use this function to compute the distance for both types.

tests/planner/ompl/test_OMPLSimplifier.cpp
aditya-vk: Removed all of this from the cleaner code since trajectory is now being constructed by hand.

tests/planner/ompl/test_OMPLSimplifier.cpp
aditya-vk: since I took 3 waypoints, I gave `maxEmptySteps = 20` with `timeOut = 20`. The probability of failing now should be less than 0.1%.

tests/planner/ompl/test_TrajectoryConversions.cpp
aditya-vk: Such a good point. Removed planning completely. Thanks!!

tests/planner/ompl/test_TrajectoryConversions.cpp
aditya-vk: Removed.

tests/planner/ompl/test_OMPLSimplifier.cpp
aditya-vk: Made it fail on purpose. Is beautiful indeed! :heart_eyes:

tests/planner/ompl/test_OMPLSimplifier.cpp
aditya-vk: Is there ever a possibility of zero waypoints though?

src/planner/ompl/Planner.cpp
```diff
@@ -372,6 +374,119 @@ trajectory::InterpolatedPtr planCRRTConnect(
...
+
+// Following are helper functions. Might want to move them elsewhere for generic
+// use
+
+::ompl::geometric::PathGeometric toOMPLTrajectory(
+    const trajectory::InterpolatedPtr& _interpolatedTraj,
+    ::ompl::base::SpaceInformationPtr _si)
+{
+  auto sspace
+      = ompl_static_pointer_cast<GeometricStateSpace>(_si->getStateSpace());
```
aditya-vk: I am a little confused here. To create the `::ompl::geometric::PathGeometric` I will need `SpaceInformationPtr`. So I cannot do it in (1) unless:
a) I create it in (2) and send it as an argument to (1)
b) I also pass the `SpaceInformationPtr` as an argument to (1).

I did not clearly understand why we are refactoring this like this, why would we want to do it?

mkoval: I missed that we needed the `SpaceInformationPtr` to construct the `PathGeometric`. Since that is the case,I agree that we *should not* split this function up. 😅 

However, I'd still like to replace the `static_cast` with a `dynamic_cast` [at least in `Debug` mode].

src/planner/ompl/Planner.cpp
mkoval: Maybe. It's hard to tell:

- If the compiler inlines the `getStateSpace()` call, then it could potentially be as efficient as passing the value in as an argument. We'd end up with the same number of `shared_ptr` copies either way.
- In this case, `Interpolator` is a pure `virtual` base class, so the call will not be inlined and will incur the penalty of a `virtual `function call (i.e. a `vtable` lookup). There is definitely _some_ overhead for a `virtual` function call, but it's often *much* less than you'd expect with a modern optimizing compiler.
- If we cared a lot about performance, then we might get the best of both worlds by making the type of the `_interpolator` parameter a template argument. A good compiler could then [devirtualize](http://hubicka.blogspot.com/2014/01/devirtualization-in-c-part-1.html) the function, then inline it to achieve the same result as (1), i.e. zero overhead.

If we really wanted to know, we'd have to run a benchmark and, if it's performance is an issue, look at the assembly code the compiler generated and determine why it's slow.

But...the difference is negligible. This is *especially* true since we only call this function once per (very long) planning query. There are also a lot of other optimizations that would get us more bang-for-the-buck.

**TL;DR:** Premature optimization is the root of all evil. Eliminating an unnecessary parameter provides a cleaner API and eliminates an entire class of errors (i.e. mixing different state spaces).

tests/planner/ompl/test_OMPLSimplifier.cpp
mkoval: This function still does not need to be templated! 😈 You can take the argument by `const &` and dereference the `unique_ptr` and `smart_ptr` at the call site.

(The same is true for a lot of Aikdo functions, including `planOMPL`, but that's a battle for another day.)

tests/planner/ompl/test_OMPLSimplifier.cpp
mkoval: Not if the code is working correctly. But: (1) a function shouldn't SEGFAULT [which this could] if you provide it a valid input, (2) if an empty trajectory is not a valid input it should be documented and tested as such, and (3) a bug in the code we're testing could easily return an empty trajectory.

src/planner/ompl/Planner.cpp
mkoval: Nevermind, I see now that `_maxEmptySteps` is unsigned.

src/planner/ompl/Planner.cpp
mkoval: This comment has been addressed.

tests/planner/ompl/test_OMPLSimplifier.cpp
aditya-vk: :sweat_smile: 
Yes. Yes. Haha! Sorry :|

aditya-vk:
1. Removed templating.
2. Replaced static cast with dynamic cast and checked for nullptr

tests/planner/ompl/test_TrajectoryConversions.cpp
```diff
@@ -0,0 +1,194 @@
...
+      0.1);
+
+  auto omplTraj = aikido::planner::ompl::toOMPLTrajectory(
+      traj, si);
+  
+  // Match every point
+  auto s0 = stateSpace->createState();
+  for(size_t idx = 0; idx < traj->getNumWaypoints(); ++idx)
+  {
+    traj->evaluate(idx, s0);
```
mkoval: This should be `getWaypoint`, not `evaluate`.

tests/planner/ompl/test_TrajectoryConversions.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+  auto omplTraj = aikido::planner::ompl::toOMPLTrajectory(
+      traj, si);
+
+  auto interpolatedTraj = aikido::planner::ompl::toInterpolatedTrajectory(
+  omplTraj, interpolator);
+
+  // Match every point
+  auto s0 = stateSpace->createState();
+  for(size_t idx = 0; idx < interpolatedTraj->getDuration(); ++idx)
```
mkoval: This should be `getNumWaypoints`, not `getDuration`.

tests/planner/ompl/test_TrajectoryConversions.cpp
```diff
@@ -0,0 +1,194 @@
...
+  {
+    interpolatedTraj->evaluate(idx, s0);
```
mkoval: This should be `getWaypoint`, not `evaluate`.

tests/planner/ompl/test_TrajectoryConversions.cpp
```diff
@@ -0,0 +1,194 @@
...
+      std::move(boundsProjection),
+      0.1);
+
+  auto omplTraj = aikido::planner::ompl::toOMPLTrajectory(
+      traj, si);
+
+  auto interpolatedTraj = aikido::planner::ompl::toInterpolatedTrajectory(
+  omplTraj, interpolator);
+
+  // Match every point
```
mkoval: Verify that the two trajectories have the same number of waypoints with an `ASSERT_EQUAL`. Otherwise, the test may SEGFAULT in the loop below.

tests/planner/ompl/test_TrajectoryConversions.cpp
```diff
@@ -0,0 +1,194 @@
...
+      std::move(sampler),
+      std::move(collConstraint),
+      std::move(boundsConstraint),
+      std::move(boundsProjection),
+      0.1);
+
+  auto omplTraj = aikido::planner::ompl::toOMPLTrajectory(
+      traj, si);
+  
+  // Match every point
```
mkoval: Verify that the two trajectories have the same number of waypoints with an `ASSERT_EQUAL`. Otherwise, the test may SEGFAULT in the loop below.

aditya-vk: The first test was checking that both trajectories have same number of waypoints. If they don't have equal waypoints, the tests would fail there and never come here. So this should be okay, yes?

tests/planner/ompl/test_TrajectoryConversions.cpp
aditya-vk: As stated above.

tests/planner/ompl/test_TrajectoryConversions.cpp
aditya-vk: `evaluate` works only if unit timing, got it! Thanks!

tests/planner/ompl/test_TrajectoryConversions.cpp
mkoval: Nope:

1. All tests are independent. GTest will continue to run tests even after one tests fail. This is useful because it often uncovers multiple bugs at once in one Travis run.
2. Tests aren't guaranteed to be ordered in any particular way. Or even that all of them are run - you can run a subset of tests by passing commandline arguments to GTest.
3. The `EXPECT_` macros intentionally allow testcase to continue and mark it as failed once it is done running. Only `ASSERT_` macros cause the test to immediately exit.

In general, all tests should be independent and a test should never crash, only gracefully report failure. If a check is critical to avoid a SEGFAULT (e.g. a length check like this), you should use `ASSERT_` instead of `EXPECT_`.

aditya-vk: I understand. Thanks! Will make the change right away!

aditya-vk:
All suggestions addressed.  :smile:

PR #166, opened 2017-04-12
Adding an abstract BarrettHandPositionCommandExecutor class

gilwoolee:
This PR adds BarrettHandPositionCommandExecutor and renames several classes. This is necessary/convenient to have a non-simulation counterpart `ros::RosBarrettHandPositionCommandExecutor` which has the same `execute` method signature as the existing sim-mode-executor.

- adds an abstract class `BarrettHandPositionCommandExecutor` which contains `execute` method for four finger goal positions (3 primal, 1 spread).
- renames existing `Barrett*Executor` classes to `SimBarrett*Executor`
- removes `collideWith` param from their `execute` method to match the abstract class's `execute` signature
- adds`collideWith` to the contructors for `SimBarrett*Executor` classes 
- adds`setCollideWith` methods to  `SimBarrett*Executor` classes

One question for @mkoval : shouldn't we expose `step` (or `spin`) method for all Executors, including the higher level abstract classes? (applies to #149 as well)


src/control/SimBarrettFingerPositionCommandExecutor.cpp
```diff
@@ -153,7 +156,7 @@ void BarrettFingerPositionCommandExecutor::step(double _timeSincePreviousCall)
   if (mDistalOnly)
     return;
 
-  // Check proximal collision 
+  // Check proximal collision
   bool proximalCollision = mCollisionDetector->collide(
     mProximalCollisionGroup.get(), mCollideWith.get(),
     mCollisionOptions, &collisionResult);
```
jslee02: This is not added by this PR, but we don't need to pass the collision result parameter since we only check whether there is a collision or not. By passing `nullptr` (or not passing), we could save time to compute the exact contact information such as point, normal, and penetration depth.

include/aikido/control/SimBarrettFingerPositionCommandExecutor.hpp
```diff
@@ -12,63 +12,68 @@
...
-private: 
-  constexpr static double kMimicRatio = 0.333; 
+  /// Resets CollisionGroup to check collision with fingers.
+  /// \param _collideWith CollisionGroup to check collision with fingers.
+  /// \return false if fails to change collideWith (during execution).
+  bool setCollideWith(::dart::collision::CollisionGroupPtr collideWith);
+
+private:
+  constexpr static double kMimicRatio = 0.333;
   constexpr static double kProximalVelocity = 0.01;
```
jslee02: This is not added by this PR, but renaming `kProximalVelocity` to `kProximalSpeed` would make more sense since the use of this value seems the value doesn't encode the direction of motion.

include/aikido/control/SimBarrettFingerPositionCommandExecutor.hpp
```diff
@@ -12,63 +12,68 @@
...
   constexpr static double kDistalVelocity = kProximalVelocity*kMimicRatio;
```
jslee02: Ditto

src/control/SimBarrettFingerPositionCommandExecutor.cpp
```diff
@@ -39,6 +41,10 @@ BarrettFingerPositionCommandExecutor::BarrettFingerPositionCommandExecutor(
   if (!mCollisionDetector)
     throw std::invalid_argument("CollisionDetctor is null.");
 
+  if (!mCollideWith)
+    throw std::invalid_argument("CollideWith is null.");
```
jslee02: If we assume `mCollideWith` is always not null, then we might don't need to pass the collision detector because `mCollideWith` has its collision detector in it (we can use it instead). Moreover, the collision detector of `mCollideWith` and `_collisionDetector` should be the same to perform collision detection anyway so not passing additional collision detector would reduce the possibility of mistake.

FYI, the collision detection between two collision groups can be done as
```cpp
bool collision = mSpreadCollisionGroup->collide(mCollideWith.get(),
    mCollisionOptions, &collisionResult);
```

include/aikido/control/SimBarrettFingerPositionCommandExecutor.hpp
```diff
@@ -12,63 +12,68 @@
...
-  std::future<void> execute(double _goalPosition,
-    ::dart::collision::CollisionGroupPtr _collideWith);
+  std::future<void> execute(double _goalPosition);
 
-  /// Returns mimic ratio, i.e. how much the distal joint moves relative to 
+  /// Returns mimic ratio, i.e. how much the distal joint moves relative to
   /// the proximal joint. The joint movements follow
   /// this ratio only when both joints are moving.
   /// \return mimic ratio.
   static double getMimicRatio();
```
jslee02: Nit: We could make this `constexpr` in this PR or leave a note as a todo.

include/aikido/control/BarrettHandPositionCommandExecutor.hpp
```diff
@@ -1,86 +1,28 @@
...
 
 namespace aikido {
 namespace control {
 
-/// Position command executor for BarrettHand fingers. 
-/// Assumes that fingers are underactuated: proximal joint is actuated
-/// and distal joint moves with certain mimic ratio until collision.
-/// If 
+/// Abstract class for position commands for a Barrett Hand.
 class BarrettHandPositionCommandExecutor
```
mkoval: Rename to `PositionCommandExecutor`.

include/aikido/control/BarrettHandPositionCommandExecutor.hpp
```diff
@@ -1,86 +1,28 @@
...
-  ///        the fingers will move only upto the limit.
-  /// \param _collideWith CollisionGroup to check collision with fingers.
+  /// Execute hand to goalPosition
+  /// \param goalPositions End dof pose for proximal joints and spread.
   /// \return Future which becomes available when the execution completes.
-  std::future<void> execute(
-    Eigen::Matrix<double, 4, 1> _goalPositions, 
-    ::dart::collision::CollisionGroupPtr _collideWith);
+  virtual std::future<void> execute(
+    Eigen::Matrix<double, 4, 1> goalPositions) = 0;
```
mkoval: Actually, make this an `Eigen::VectorXd` since this is now a generic interface - potentially for hardware other than the BarrettHand. E.g. we _might_ decide to use a similar interface for the pan-tilt unit.

Also, [passing a fixed-size `Eigen::Matrix` by value is **wrong**](https://eigen.tuxfamily.org/dox/group__TopicPassingByValue.html) because the argument may have incorrect alignment. This is more than a nit. 😉

include/aikido/control/BarrettHandPositionCommandExecutor.hpp
```diff
@@ -1,86 +1,28 @@
...
-  /// mProximalGoalPositions, mSpreadGoalPositin, mSpread
-  std::mutex mMutex;
-
-  /// Flag for indicating execution of a command. 
-  bool mInExecution;
-
-  /// Values for executing a position and spread command. 
-  Eigen::Vector3d mProximalGoalPositions; 
-  double mSpreadGoalPosition;
+  virtual void step(double timeSincePreviousCall) = 0;
```
mkoval: Remove the `timeSincePreviousCall` argument. If this is necessary inside some executors, maintain this state as a member variable.

include/aikido/control/SimBarrettFingerPositionCommandExecutor.hpp
```diff
@@ -12,63 +12,68 @@
...
-/// It moves a finger to a desired point; it may stop early if 
+/// This executor mimics the behavior of BarretFinger.
+/// It moves a finger to a desired point; it may stop early if
 /// joint limit is reached or collision is detected.
 /// Only the proximal joint is actuated; the distal joint moves with mimic ratio.
 /// When collision is detected on the distal link, the finger stops.
 /// When collision is detected on the proximal link, the distal link moves
 /// until it reaches joint limit or until distal collision is detected.
-class BarrettFingerPositionCommandExecutor
+class SimBarrettFingerPositionCommandExecutor
```
mkoval: `Simulated`, not `Sim`! 😉

include/aikido/control/SimBarrettFingerPositionCommandExecutor.hpp
```diff
@@ -12,63 +12,68 @@
...
-  ///        See dart/collison/Option.h for more information 
-  BarrettFingerPositionCommandExecutor(
+  ///        maxNumContacts = 1.)
+  ///        See dart/collison/Option.h for more information
+  SimBarrettFingerPositionCommandExecutor(
     ::dart::dynamics::ChainPtr _finger, int _proximal, int _distal,
     ::dart::collision::CollisionDetectorPtr _collisionDetector,
+    ::dart::collision::CollisionGroupPtr _collideWith,
     ::dart::collision::CollisionOption _collisionOptions
       = ::dart::collision::CollisionOption(false, 1));
```
mkoval: I suggest making the `_collideWith` argument optional. If it's not specified, default to an empty `CollisionGroup`.

include/aikido/control/BarrettHandPositionCommandExecutor.hpp
jslee02: :+1: if this is generic for arbitrary dofs. I assumed this class is generic only for Barrett Hand, which I believe the dimension of configuration is four.

gilwoolee:
@jslee02 As far as I know, I don't think it's necessary to create an abstract class for fingers. The abstraction at hand level is necessary so that both simulation-mode and real-mode can be handled by the same API. The real-mode does not need finger-executors; we just need to call something like `RosBarrettHandPositionCommandExecutor::execute(goalPosition)` that sends commands using `actionlib`. On a separate thread (e.g. in `libherb`) we'll have a `MultiplexExecutor` that calls `step()` method of `RosBarrettHandPositionCommandExecutor`( `BarrettHandKinematicSimulationExecutor`) to check the execution status (or step forward).

gilwoolee:
After discussion with @mkoval I had to make another major change: change `KinematicSimulationTrajectoryExecutor::step` to just step once, and remove all multi-threading component. The multi-threaded, step-forwarding part is expected to be handled by an external thread calling `step()` regularly. This make `KinematicSimulationTrajectoryExecutor` consistent with all other executors.

@mkoval @jslee02 please take a look. Thanks!

src/control/BarrettHandKinematicSimulationPositionCommandExecutor.cpp
```diff
@@ -27,50 +30,61 @@ BarrettHandPositionCommandExecutor::BarrettHandPositionCommandExecutor(
...
+    message << "GoalPositions must have 4 elements, but ["
+      << goalPositions.size() << "] given.";
+    throw std::invalid_argument(message.str());
+  }
+
   mPromise.reset(new std::promise<void>());
-  mProximalGoalPositions = _positions.topRows(3); 
-  mSpreadGoalPosition = _positions(3);
-  mCollideWith = _collideWith;
+  mProximalGoalPositions = goalPositions.topRows(3);
```
jslee02: Nit: `topRows()` is a method for matrices. We could the method for vectors as `head<3>()`

src/control/BarrettHandKinematicSimulationPositionCommandExecutor.cpp
```diff
@@ -27,50 +30,61 @@ BarrettHandPositionCommandExecutor::BarrettHandPositionCommandExecutor(
...
+  mSpreadGoalPosition = goalPositions(3);
```
jslee02: Nit: Brackets `[ ]` are preferable for vector indexing.

include/aikido/control/BarrettFingerKinematicSimulationSpreadCommandExecutor.hpp
```diff
@@ -0,0 +1,96 @@
...
+  /// \return future which becomes available when the execution completes.
+  std::future<void> execute(double goalPosition);
+
+  /// Moves the joint of the finger by fixed speed*timeSincePreviousCall
+  /// until execute's goalPosition by spread dof or collision is detected.
+  /// If multiple threads are accessing this function or the skeleton associated
+  /// with this executor, it is necessary to lock the skeleton before
+  /// calling this method.
+  /// \param[in] timeSincePreviousCall Time interval to take.
+  void step(const std::chrono::milliseconds& timeSincePreviousCall);
```
gilwoolee: This call would always be called by `BarrettHandKinematicSimulationPositionCommandExecutor::step()` method, so I am not sure whether a default is necessary. We can set it to something I guess... :)

jslee02: If that's the case, I'm fine with not setting a default value here.

include/aikido/control/BarrettFingerKinematicSimulationPositionCommandExecutor.hpp
```diff
@@ -8,67 +8,70 @@
...
+  /// \param[in] finger Finger to be controlled by this Executor.
+  /// \param[in] proximal Index of proximal dof
+  /// \param[in] distal Index of distal dof
+  /// \param[in] collideWith CollisionGroup to check collision with fingers.
+  /// \param[in] collisionOptions Default is (enableContact=false, binaryCheck=true,
+  ///        maxNumContacts = 1.)
+  ///        See dart/collison/Option.h for more information
+  BarrettFingerKinematicSimulationPositionCommandExecutor(
+    ::dart::dynamics::ChainPtr finger, size_t proximal, size_t distal,
+    ::dart::collision::CollisionGroupPtr collideWith,
```
brianhou: My understanding is that the arguments should be `finger`, `proximal`, `distal` (which are all required arguments), `collisionDetector` and `collideWith` (which are both optional and default to `nullptr`), and `collisionOptions` (which remains optional and defaults to `false` and `1`).

It seems like the three cases are:
1. **If `collisionDetector` and `collideWith` are not specified**, set `collisionDetector ` to an `FCLCollisionDetector` and `collideWith` to a new empty `CollisionGroup` with that collision detector.
2. **If `collisionDetector` is specified and `collideWith` is not**, set `collideWith` to a new empty `CollisionGroup` with `collisionDetector` as the collision detector
3. **If `collisionDetector` and `collideWith` are specified**, check that the collision detector for `collideWith` matches `collisionDetector`. if not, set `collideWith` to a new empty `CollisionGroup` with `collisionDetector` as the collision detector

@jslee02 Is my understanding correct and if so, does e16880e9fca951199c50b7639a8ba46f7dbaa318 seem reasonable? In particular, I'm unsure what it means for two collision detectors to not be the same. Do they have to be literally the same object or just the same type of collision detector? It also seems surprising to me to automatically set `collideWith` to an empty collision group if there were originally collision objects in there, so I think I might be misunderstanding something. :confused:

jslee02: The three cases sound good. I would like to add one additional case that `collisionDetector` is `nullptr` and `collideWith` is specified. In this case, we set `mCollisionDetector` to `collideWith->getCollisionDetector()`.

> In particular, I'm unsure what it means for two collision detectors to not be the same. Do they have to be literally the same object or just the same type of collision detector?

They have to be the same _type_, but don't necessarily need to be the same object (but it would be good to be the same object because it could share collision geometries -- this is an implementation detail). The main role of a collision detector is to enforce us to create collision groups of the same collision detection engines (e.g., fcl, bullet, ode). This is necessary because we cannot collide two collision groups created from different collision engines (unless a collision detection engine supports that feature, which I haven't heard about it).

The last bullet point case in my previous comment is that `collideWith`'s collision detector (who created `collideWith`) is not the same type with `collisionDetector`. In this case, we cannot collide HERB hand with `collideWith` because of the reason described above. So we might want to either (1) create a new collision group from the shape frames in `collideWith` through `collisionDetector` or (2) create a collision group of HERB hand through the collision detector type of `collideWith` by `collideWith->getCollisionDetector()->createCollisionGroup(...)`.

Sorry, probably my description was unclear. :sweat:

brianhou: Got it, thank you for the clarification! I've rewritten the logic to add the fourth case. :smiley:

include/aikido/control/BarrettHandKinematicSimulationPositionCommandExecutor.hpp
```diff
@@ -0,0 +1,95 @@
...
+  ///        should have (spread, proximal, distal) joints.
+  ///        Third finger should have 2 joints (proximal, distal).
+  /// \param[in] spreadCommandExecutor Executors to control
+  ///        spreads of the fingers.
+  /// \param[in] collideWith CollisionGroup to check collision with fingers.
+  BarrettHandKinematicSimulationPositionCommandExecutor(
+    const std::array<
+      BarrettFingerKinematicSimulationPositionCommandExecutorPtr, 3>& positionCommandExecutors,
+    BarrettFingerKinematicSimulationSpreadCommandExecutorPtr spreadCommandExecutor,
+    ::dart::collision::CollisionGroupPtr collideWith);
```
brianhou: I wound up basically copying and pasting the optional nullptr logic from `BarrettFingerKinematicSimulationPositionCommandExecutor`. Since the two classes are completely independent I didn't see an obvious way to avoid that... :cry:

src/control/BarrettFingerKinematicSimulationPositionCommandExecutor.cpp
```diff
@@ -40,24 +40,35 @@ ::BarrettFingerKinematicSimulationPositionCommandExecutor(
...
-  // Default mCollideWith to empty collision group. If a collision group is
-  // given and its collision detector does not match mCollisionDetector, set the
-  // collision group to an empty collision group.
-  if (!mCollideWith ||
-      mCollisionDetector != mCollideWith->getCollisionDetector())
+  if (mCollisionDetector && mCollideWith)
   {
-    if (mCollideWith)
+    // If a collision group is given and its collision detector does not match
+    // mCollisionDetector, set the collision group to an empty collision group.
```
jslee02: We could add `ShapeFrame`s of `collideWith` to the new collision group as
```cpp
mCollideWith = mCollisionDetector->createCollisionGroup();
for (auto i = 0u; i < collideWith->getNumShapeFrames(); ++i)
  mCollideWith->addShapeFrame(collideWith->getShapeFrame(i));
```
so that still can check collision as expected.

include/aikido/control/BarrettFingerKinematicSimulationSpreadCommandExecutor.hpp
```diff
@@ -0,0 +1,96 @@
...
+public:
+  /// Constructor.
+  /// \param[in] fingers 2 fingers to be controlled by this Executor.
+  /// \param[in] spread Index of spread dof
+  /// \param[in] collideWith CollisionGroup to check collision with fingers.
+  /// \param[in] collisionOptions Default is (enableContact=false, binaryCheck=true,
+  ///        maxNumContacts = 1.)
+  BarrettFingerKinematicSimulationSpreadCommandExecutor(
+    std::array<::dart::dynamics::ChainPtr, 2> fingers, size_t spread,
+    ::dart::collision::CollisionGroupPtr collideWith,
```
mkoval: Nit: Make `collideWith` optional.

include/aikido/control/BarrettHandKinematicSimulationPositionCommandExecutor.hpp
```diff
@@ -0,0 +1,100 @@
...
+  /// \param[in] collideWith CollisionGroup to check collision with fingers.
+  ///        If nullptr, default to empty CollisionGroup
+  BarrettHandKinematicSimulationPositionCommandExecutor(
+    const std::array<
+      BarrettFingerKinematicSimulationPositionCommandExecutorPtr, 3>& positionCommandExecutors,
+    BarrettFingerKinematicSimulationSpreadCommandExecutorPtr spreadCommandExecutor,
+    ::dart::collision::CollisionDetectorPtr collisionDetector = nullptr,
+    ::dart::collision::CollisionGroupPtr collideWith = nullptr);
+
+  /// Set relevant variables for moving fingers.
```
mkoval: Nit: Confusing docstring.

include/aikido/control/BarrettHandKinematicSimulationPositionCommandExecutor.hpp
```diff
@@ -0,0 +1,100 @@
...
+  bool setCollideWith(::dart::collision::CollisionGroupPtr collideWith);
+
+private:
+
+  constexpr static int kNumPositionExecutor = 3;
+  constexpr static int kNumSpreadExecutor = 1;
+  constexpr static auto kWaitPeriod = std::chrono::milliseconds(1);
+
+  /// Executor for proximal and distal joints.
+  std::array<BarrettFingerKinematicSimulationPositionCommandExecutorPtr, 3> mPositionCommandExecutors;
```
mkoval: Nit: Use `kNumPositionExecutor` instead of `3` for the template parameter.

include/aikido/control/BarrettHandKinematicSimulationPositionCommandExecutor.hpp
```diff
@@ -0,0 +1,100 @@
...
+  /// \param[in] spreadCommandExecutor Executors to control
+  ///        spreads of the fingers.
+  /// \param[in] collisionDetector CollisionDetector to check collision with fingers.
+  ///        If nullptr, default to FCLCollisionDetector.
+  /// \param[in] collideWith CollisionGroup to check collision with fingers.
+  ///        If nullptr, default to empty CollisionGroup
+  BarrettHandKinematicSimulationPositionCommandExecutor(
+    const std::array<
+      BarrettFingerKinematicSimulationPositionCommandExecutorPtr, 3>& positionCommandExecutors,
+    BarrettFingerKinematicSimulationSpreadCommandExecutorPtr spreadCommandExecutor,
```
mkoval: I suggest hiding the instantiation of `positionCommandExecutors` and `spreadCommandExecutor` inside the constructor - they're more of an implementation detail than part of the public API.

include/aikido/control/PositionCommandExecutor.hpp
```diff
@@ -0,0 +1,30 @@
+#ifndef AIKIDO_CONTROL_POSITIONCOMANDEXECUTOR_HPP_
+#define AIKIDO_CONTROL_POSITIONCOMANDEXECUTOR_HPP_
+#include <Eigen/Dense>
+#include <future>
+
+namespace aikido {
+namespace control {
+
+/// Abstract class for executing position commands.
```
mkoval: Nit: Elaborate that "a position command requests several DegreesOfFreedom to move to a specified configuration and provides feedback when the move is complete."

include/aikido/control/PositionCommandExecutor.hpp
```diff
@@ -0,0 +1,30 @@
...
+class PositionCommandExecutor
+{
+public:
+  virtual ~PositionCommandExecutor() = default;
+
+  /// Execute hand to goalPosition
+  /// \param goalPositions Goal positions
+  /// \return future which becomes available when the execution completes.
```
mkoval: Nit: Note that the command succeeds when movement stops, not when the `DegreeOfFreedom`s reach the specified `goalPositions`. Depending upon the implementation, the command may terminate for a variety of reasons, e.g. because of collision.

src/control/BarrettFingerKinematicSimulationPositionCommandExecutor.cpp
```diff
@@ -72,47 +96,47 @@ std::future<void> BarrettFingerPositionCommandExecutor::execute(
...
 {
   mPromise->set_value();
   mInExecution = false;
-  mCollideWith.reset();
 }
+
 //=============================================================================
-void BarrettFingerPositionCommandExecutor::step(double _timeSincePreviousCall)
+void BarrettFingerKinematicSimulationPositionCommandExecutor::step(
+ const std::chrono::milliseconds& timeSincePreviousCall)
```
mkoval: I suggest removing this argument entirely and doing the bookkeeping in a member variable. If we do that, this class can also implement the `PositionCommandExecutor` interface.

That is not a hard requirement, but it is a Nice To Have..

src/control/BarrettFingerKinematicSimulationPositionCommandExecutor.cpp
```diff
@@ -185,13 +209,26 @@ void BarrettFingerPositionCommandExecutor::step(double _timeSincePreviousCall)
...
+//=============================================================================
+bool BarrettFingerKinematicSimulationPositionCommandExecutor::setCollideWith(
+  ::dart::collision::CollisionGroupPtr collideWith)
+{
+  std::lock_guard<std::mutex> lockSpin(mMutex);
+
+  if (mInExecution)
+    return false;
+
+  mCollideWith = collideWith;
```
mkoval: Nit: `std::move`.

src/control/BarrettFingerKinematicSimulationPositionCommandExecutor.cpp
```diff
@@ -185,13 +209,26 @@ void BarrettFingerPositionCommandExecutor::step(double _timeSincePreviousCall)
...
-}
 
 }
+
+//=============================================================================
+bool BarrettFingerKinematicSimulationPositionCommandExecutor::setCollideWith(
+  ::dart::collision::CollisionGroupPtr collideWith)
+{
+  std::lock_guard<std::mutex> lockSpin(mMutex);
+
```
mkoval: Nit: Check that `collideWith->getCollisionDetector()` is correct. Alternatively, update `mCollisionDetector` with that value.

src/control/BarrettFingerKinematicSimulationSpreadCommandExecutor.cpp
```diff
@@ -104,53 +101,53 @@ std::future<void> BarrettFingerSpreadCommandExecutor::execute(
...
     mInExecution = true;
 
     return mPromise->get_future();
   }
 }
 
 //=============================================================================
-void BarrettFingerSpreadCommandExecutor::step(double _timeSincePreviousCall)
+void BarrettFingerKinematicSimulationSpreadCommandExecutor::step(
+  const std::chrono::milliseconds& timeSincePreviousCall)
```
mkoval: Same comment as above. I suggest removing this parameter entirely, so this class complies with the `PositionCommandExecutor` interface.

src/control/BarrettFingerKinematicSimulationSpreadCommandExecutor.cpp
```diff
@@ -104,53 +101,53 @@ std::future<void> BarrettFingerSpreadCommandExecutor::execute(
...
   std::lock_guard<std::mutex> lock(mMutex);
 
   if (!mInExecution)
     return;
-  
+
   // Current spread. Check that all spreads have same values.
   double spread = mSpreadDofs[0]->getPosition();
-  for(int i=1; i < kNumFingers; ++i)
+  for(size_t i = 1; i < kNumFingers; ++i)
```
mkoval: This looks like an off-by-one error. What am I missing? 🤔

src/control/BarrettFingerKinematicSimulationSpreadCommandExecutor.cpp
```diff
@@ -159,18 +156,32 @@ void BarrettFingerSpreadCommandExecutor::step(double _timeSincePreviousCall)
...
+//=============================================================================
+bool BarrettFingerKinematicSimulationSpreadCommandExecutor::setCollideWith(
+  ::dart::collision::CollisionGroupPtr collideWith)
+{
+  std::lock_guard<std::mutex> lockSpin(mMutex);
+
+  if (mInExecution)
+    return false;
+
+  mCollideWith = collideWith;
```
mkoval: Nit: `std::move`.

src/control/BarrettFingerKinematicSimulationSpreadCommandExecutor.cpp
```diff
@@ -159,18 +156,32 @@ void BarrettFingerSpreadCommandExecutor::step(double _timeSincePreviousCall)
...
     spreadDof->setPosition(newSpread);
-
 }
 
+//=============================================================================
+bool BarrettFingerKinematicSimulationSpreadCommandExecutor::setCollideWith(
+  ::dart::collision::CollisionGroupPtr collideWith)
+{
+  std::lock_guard<std::mutex> lockSpin(mMutex);
+
```
mkoval: Verify (or make) `collideWIth->getCollisionDetector()` be consistent with `mCollisionDetector`.

include/aikido/control/BarrettFingerKinematicSimulationPositionCommandExecutor.hpp
```diff
@@ -8,67 +8,74 @@
...
+  /// \return false if fails to change collideWith (during execution).
+  bool setCollideWith(::dart::collision::CollisionGroupPtr collideWith);
 
-private: 
-  constexpr static double kMimicRatio = 0.333; 
-  constexpr static double kProximalVelocity = 0.01;
-  constexpr static double kDistalVelocity = kProximalVelocity*kMimicRatio;
+private:
+  constexpr static double kMimicRatio = 0.333;
+  constexpr static double kProximalSpeed = 1;
```
brianhou: This was because the hand was moving _really_ slowly when we were visualizing it in sim. I don't know if either is correct. How would we check that?

include/aikido/control/BarrettFingerKinematicSimulationPositionCommandExecutor.hpp
```diff
@@ -8,67 +8,74 @@
...
-    ::dart::collision::CollisionGroupPtr _collideWith);
+  std::future<void> execute(double goalPosition);
 
-  /// Returns mimic ratio, i.e. how much the distal joint moves relative to 
+  /// Returns mimic ratio, i.e. how much the distal joint moves relative to
   /// the proximal joint. The joint movements follow
   /// this ratio only when both joints are moving.
   /// \return mimic ratio.
-  static double getMimicRatio();
+  constexpr static double getMimicRatio() { return kMimicRatio; }
```
brianhou: It seems that the only time that `getMimicRatio` is used is in tests, but `kMimicRatio` is private so getting rid of this function wouldn't work.

include/aikido/control/BarrettFingerKinematicSimulationPositionCommandExecutor.hpp
mkoval: I suspect we arbitrarily chose a value for testing, then. 😢

1. Our controller [calls the `trapezoidalMove()` function](https://github.com/personalrobotics/barrett_ros_control/blob/master/src/tasks/BHMoveHandTask.cpp#L78) to move the hand:
2. That function [sets the `E` Puck property to the desired position, then the `MODE` property to `MODE_TRAPEZOIDAL`](https://github.com/BarrettTechnology/libbarrett/blob/8e97d0f6d3b36fc14c5520d583691146627a97ca/src/products/hand.cpp#L156) to execute the action.
3. Here is [a snippet of the description](http://support.barrett.com/wiki/Hand/282/ControlModes) of the behavior of the hand in `MODE_TRAPEZOIDAL` (the full description goes into much more detail):

    > [...] The BarrettHand will generate its own trapezoidal velocity profile to accelerate from its present position (using ACCEL) until it reaches max velocity (MV), then it will decelerate to reach the target endpoint position at zero velocity. [...]

So the correct way to get this value is to query the `MV` property on the four pucks in the BarrettHand, then use the transmission ratio to convert the motor velocity to a joint velocity. We should do this at some point (maybe file an issue?), but it's not urgent to get this right.

In the meantime, we can estimate a value by timing how long it takes the hand to close. This is all an approximation, anyway, since we are assuming the finger immediately accelerates to its maximum velocity.

From watching a YouTube video of HERB, I estimate around 1 s. Since the joint range is 140 degrees, this is around 2.4 rad/s. 🔬

mkoval: On second thought, I am not sure why this value is hardcoded at all. We should use the velocity limits that are stored in the DART `Skeleton`. Those are currently set to 2 rad/s in `herb_description`, which is already in the correct ballpark. 👍 

@brianhou Are you up for making that change? 😁

include/aikido/control/BarrettFingerKinematicSimulationPositionCommandExecutor.hpp
mkoval: Understood. Let's leave it as it is.

src/control/BarrettFingerKinematicSimulationPositionCommandExecutor.cpp
gilwoolee: So would we have `setTimeSimcePreviousCall()`? At each call to `step()` we are updating the dofs values using `timeSincePreviousCall`, which is different per every call.

src/control/BarrettFingerKinematicSimulationSpreadCommandExecutor.cpp
brianhou: This isn't an off-by-one error because it's comparing the 0th spread to all other spreads.

gilwoolee: I don't think it's off-by-one. It uses the 0th-finger's spread value to check that the rest of the fingers (the other finger) has the same spread value. We implemented this because the spread is symmetric.

src/control/BarrettFingerKinematicSimulationPositionCommandExecutor.cpp
mkoval: I was thinking that we'd have a `mTimeOfPreviousCall` member variable that gets set to `now()` at the bottom of each function call. Then, we'd use `now() - mTimeOfPreviousCall` as the time since the previous call.

src/control/BarrettFingerKinematicSimulationSpreadCommandExecutor.cpp
mkoval: Got it, I see that now. 👍

include/aikido/control/BarrettFingerKinematicSimulationPositionCommandExecutor.hpp
brianhou: Hmm, it seems like making this change would require `kProximalSpeed` and `kDistalSpeed` to not be static, right? I'm not sure why they're static now and whether that's necessary though.

brianhou:
@mkoval @jslee02 I think I've addressed all your suggestions (except perhaps the one about reading `kProximalSpeed` from `herb_description`). Let me know if this looks good to merge!

brianhou:
I realized I forgot to fix the tests, but I've fixed some of them now.
- `test_BarrettFinger` tests now compile, although they seem to indicate that `step` isn't actually updating anything. @gilwoolee based on our conversation earlier this afternoon, this might be expected behavior because there isn't a separate thread running? Is that correct?
- `test_BarrettHand` partially compiles, although that's because I've commented out the majority of it. Now that `robot` is a parameter, it seems that testing this will become more difficult. I believe that `aikido` currently doesn't have any dependency on `herb_description`, so we can't load in the urdf as a skeleton (which makes it impossible to construct a `BarrettHandCommandExecutor`). Any thoughts on how to fix that?

gilwoolee:
@brianhou To answer the questions: 
- Yes. It is correct. We need to convert those that take in `step(timeSincePreviousCall)` by measuring the time between `execute(...)` `step()` and see if the dof values after `step()` approximately matches the expected value (computed by `ellapsedTime *speed`. Some of the tests  test collision-related behaviors, so they can just run for completion (until `future` returns). We should be able to get the same dof values as before.
- I don't know how it exactly works, but @jslee02 knows about how to import resource or make some package dependency for test-only purposes. @jslee02 could you add to this?

mkoval:
The only real option I see is to include a BarrettHand model in the Aikido repository. We can put that in the `test/resources` directory, where we currently store files necessary to test `CatkinResourceRetriever`.

I suggest building `herb_description` and copying the standalone BH-280 URDF file out of the devel (or install) space. Then, edit the URDF file to: (1) replace all references to visual geometry with the collision geometry to reduce the number of binary mesh files that we need to copy and (2) modify the path to mesh files to resolve when we run unit tests.

@jslee02 had some trouble with (2) in DART, but was able to work around it using a custom `ResourceRetriever`. He may be able to shed some light about how to approach this.

jslee02:
Do we want to test `BarrettHandKinematicSimulationPositionCommandExecutor` that is dependent on `herb_description`? If so, I would like to suggest moving this class (and other BarrettHand related classes) to `libherb` since it's not a robot agnostic implementation.

brianhou:
We decided that the right thing to do is to include a Barrett hand model in the repository at some point. In the meantime, we'll merge this (to unblock other PRs) and open an issue to fix that.

PR #173, opened 2017-04-15
Add RosPositionCommandExecutor

Shushman:
This PR introduces a ROS executor for commands to the Barrett Hand through `aikido/control/ros/RosBarrettHandPositionCommandExecutor.hpp-cpp`.

A point to note is that there is a lot of code (and logic) reuse between this and `RosTrajectoryExecutor`. Any way to make this more succinct?


include/aikido/control/ros/RosBarrettHandPositionCommandExecutor.hpp
```diff
@@ -0,0 +1,81 @@
...
+  /// \param[in] connectionPollingPeriod Polling period for server connection.
+  template <typename DurationA, typename DurationB>
+  RosBarrettHandPositionCommandExecutor(
+    ::ros::NodeHandle node,
+    const std::string& serverName,
+    const DurationA& connectionTimeout = std::chrono::milliseconds{1000},
+    const DurationB& connectionPollingPeriod = std::chrono::milliseconds{20}
+  );
+
+  ~RosBarrettHandPositionCommandExecutor();
```
mkoval: Either make this `virtual` or mark the class as `final`. In this case, I'd prefer the former.

include/aikido/control/ros/RosBarrettHandPositionCommandExecutor.hpp
```diff
@@ -0,0 +1,81 @@
...
+  /// \param[in] serverName Name of the server to send traejctory to.
+  /// \param[in] connectionTimeout Timeout for server connection.
+  /// \param[in] connectionPollingPeriod Polling period for server connection.
+  template <typename DurationA, typename DurationB>
+  RosBarrettHandPositionCommandExecutor(
+    ::ros::NodeHandle node,
+    const std::string& serverName,
+    const DurationA& connectionTimeout = std::chrono::milliseconds{1000},
+    const DurationB& connectionPollingPeriod = std::chrono::milliseconds{20}
+  );
```
mkoval: Nit: I don't think there is any value in templateing `DurationA` and `DurationB` here, since we immediately store these values in `std::chrono::milliseconds` member variables. To make this work, we would have to template the entire class on these two types - which seems like it is distinctly _not_ worth it.

I believe @jslee02, @gilwoolee, and I discussed this in a previous PR. Tagging them to see if they have anything to add.

include/aikido/control/ros/RosBarrettHandPositionCommandExecutor.hpp
```diff
@@ -0,0 +1,81 @@
...
+
+  ~RosBarrettHandPositionCommandExecutor();
+
+  /// Sends positions to ROS server for execution.
+  /// \param[in] goalPositions Vector of 3 finger positions and 1 spread position
+  /// \param[in] jointNames Vector of 4 joint names for fingers
+  std::future<void> execute(Eigen::VectorXd& goalPositions,std::vector<std::string>& jointNames);
```
mkoval: - Take `goalPositions` by `const &`.
- Make `jointNames` a constructor argument, instead of making it an argument here. This is necessary to comply with the `PositionCommandExecutor` interface that @gilwoolee added in #166.

include/aikido/control/ros/RosBarrettHandPositionCommandExecutor.hpp
```diff
@@ -0,0 +1,81 @@
...
+#include <pr_control_msgs/SetPositionAction.h>
+#include <actionlib/client/action_client.h>
+#include <Eigen/Dense>
+
+namespace aikido{
+namespace control {
+namespace ros {
+
+/// This class sends commands for the hand position to the ROS server
+class RosBarrettHandPositionCommandExecutor
```
mkoval: - This should inherit from `PositionCommandExecutor ` that is added in #166.
- This class is **not** specific to the BarrettHand, so it needs to be renamed. I suggest `RosPositionCommandExecutor` to mimic `RosTrajectoryExecutor`.

include/aikido/control/ros/RosBarrettHandPositionCommandExecutor.hpp
```diff
@@ -0,0 +1,81 @@
...
+  std::future<void> execute(Eigen::VectorXd& goalPositions,std::vector<std::string>& jointNames);
+
+  /// Sends positions to ROS server for execution.
+  /// \param[in] goalPositions Vector of 3 finger positions and 1 spread position
+  /// \param[in] jointNames Vector of 4 joint names for fingers
+  /// \param[in] startTime Start time for the trajectory.
+  std::future<void> execute(
+    Eigen::VectorXd& goalPositions, 
+    std::vector<std::string>& jointNames,
+    const ::ros::Time& startTime);
```
mkoval: The `pr_control_msgs/SetPosition` action doesn't have a time parameter, so this argument is not necessary. We should remove this function and roll everything into the `execute(const Eigen::VectorXd&)` overload.

src/control/ros/Conversions.cpp
```diff
@@ -429,6 +429,36 @@ trajectory_msgs::JointTrajectory toRosJointTrajectory(
   return jointTrajectory;
 }
 
+//=============================================================================
+sensor_msgs::JointState toJointState(
+  const Eigen::VectorXd& goalPositions, const std::vector<std::string> jointNames)
```
mkoval: Nit: Take both arguments by `const &`.

src/control/ros/Conversions.cpp
```diff
@@ -429,6 +429,36 @@ trajectory_msgs::JointTrajectory toRosJointTrajectory(
...
+{
+  if (goalPositions.size() != jointNames.size())
+  {
+    std::stringstream message;
+    message << "The size of goalPositions must be the same as jointNames!" ;
+    throw std::invalid_argument(message.str());
```
mkoval: Nit: Include `goalPositions.size()` and `jointNames.size()` in the message.

src/control/ros/Conversions.cpp
```diff
@@ -429,6 +429,36 @@ trajectory_msgs::JointTrajectory toRosJointTrajectory(
...
+  const auto numJoints = jointNames.size();
+
+  jointState.name.reserve(numJoints);
+  jointState.position.reserve(numJoints);
+
+  for (size_t i = 0; i < numJoints ; ++i)
+  {
+    jointState.name.emplace_back(jointNames[i]);
+    jointState.position.emplace_back(goalPositions[i]);
+  }
```
mkoval: Nit: All of this code can be replaced by:
```c++
sensor_msgs::JointState jointState;
jointState.name = std::move(jointNames); // assuming jointNames is taken by value
jointState.position.assign(goalPositions.data(), goalPositions.data() + goalPositions.size());
return jointState;
```

src/control/ros/RosBarrettHandPositionCommandExecutor.cpp
```diff
@@ -0,0 +1,181 @@
...
+      const std::string& serverName,
+      const DurationA& connectionTimeout,
+      const DurationB& connectionPollingPeriod)
+  : mNode(std::move(node))
+  , mCallbackQueue{}
+  , mClient{mNode, serverName, &mCallbackQueue}
+  , mConnectionTimeout{
+    std::chrono::duration_cast<milliseconds>(connectionTimeout)}
+  , mConnectionPollingPeriod{
+    std::chrono::duration_cast<milliseconds>(connectionPollingPeriod)}
```
mkoval: See my comment above about this cast. I'd prefer to eliminate the template parameters entirely and force _the user_ to perform this conversion, if necessary.

If we do decide to keep the template parameters, then this implementation needs to be moved into a header file. By convention, it would go into a file named `src/control/ros/detail/RosBarrettHandPositionCommandExecutor-impl.cpp`.

src/control/ros/RosBarrettHandPositionCommandExecutor.cpp
```diff
@@ -0,0 +1,181 @@
...
+  return execute(goalPositions, jointNames, invalidTime);
+}
+
+//=============================================================================
+std::future<void> RosBarrettHandPositionCommandExecutor::execute(
+  Eigen::VectorXd& goalPositions, 
+  std::vector<std::string>& jointNames,
+  const ::ros::Time& startTime)
+{
+  if (goalPositions.size() != 4)
```
mkoval: Replace the hard-coded value `4` with `jointNames.size()`.

src/control/ros/RosBarrettHandPositionCommandExecutor.cpp
```diff
@@ -0,0 +1,181 @@
...
+      << goalPositions.size() << "] given.";
+    throw std::invalid_argument(message.str());
+  }
+
+  pr_control_msgs::SetPositionGoal goal;
+
+  // Convert goal positions and joint names to jointstate
+  goal.command = toJointState(goalPositions,jointNames);
+
+  goal.command.header.stamp = startTime;
```
mkoval: I don't _think_ `SetPositionGoal` has a `header`. Are you sure this compiles?

src/control/ros/RosBarrettHandPositionCommandExecutor.cpp
```diff
@@ -0,0 +1,181 @@
...
+      boost::bind(&RosBarrettHandPositionCommandExecutor::transitionCallback, this, _1));
+
+    return mPromise.get_future();
+  }
+
+}
+
+//=============================================================================
+// TODO : This is directly reused from RosTrajectoryExecutor.cpp
+// Anyway to reuse this? (assuming we want this here)
```
mkoval: I agree! Code duplication sucks. 😁 

It should be straightforward to move this to a helper function:
```c++
template <class ActionSpec, class TimeoutDuration, class PeriodDuration>
bool waitForActionServer(
  actionlib::ActionClient<ActionSpec>& actionClient,
  ros::CallbackQueue* callbackQueue,
  TimeoutDuration timeoutDuration = std::chrono::seconds{ 1 },
  PeriodDuration periodDuration = std::chronos::milliseconds{ 10 }
);
```

src/control/ros/RosBarrettHandPositionCommandExecutor.cpp
```diff
@@ -0,0 +1,181 @@
...
+      mPromise.set_value();
+
+    mInProgress = false;
+
+  }
+}
+
+//=============================================================================
+// TODO : This is also basically copied from RosTrajectoryExecutor
+void RosBarrettHandPositionCommandExecutor::step()
```
mkoval: I think that's fine in for now. We may want to write a base class for `actionnlib`-based executors in the future, but there really isn't all that much duplicated code once we move `waitForServer` to a helper function.

include/aikido/control/ros/RosBarrettHandPositionCommandExecutor.hpp
Shushman: Changed name and is now inheriting

include/aikido/control/ros/RosBarrettHandPositionCommandExecutor.hpp
Shushman: Removed template params and changed to `std::chrono::milliseconds`

include/aikido/control/ros/RosBarrettHandPositionCommandExecutor.hpp
Shushman: Made `virtual`

include/aikido/control/ros/RosBarrettHandPositionCommandExecutor.hpp
Shushman: Done both

include/aikido/control/ros/RosBarrettHandPositionCommandExecutor.hpp
Shushman: Removed - only a single `execute` method now

src/control/ros/RosBarrettHandPositionCommandExecutor.cpp
Shushman: Done

src/control/ros/RosBarrettHandPositionCommandExecutor.cpp
Shushman: Removed as we no longer need `startTime`. But in general the `command` which is `JointState` does have a header.

src/control/ros/RosBarrettHandPositionCommandExecutor.cpp
Shushman: I've moved this to `aikido/include/control/ros/util.hpp` and am calling this directly with templates in `execute`.

include/aikido/control/ros/Conversions.hpp
```diff
@@ -28,6 +29,14 @@ std::unique_ptr<aikido::trajectory::Spline> toSplineJointTrajectory(
 trajectory_msgs::JointTrajectory toRosJointTrajectory(
   const aikido::trajectory::TrajectoryPtr& trajectory, double timestep);
 
+/// Converts Eigen VectorXd and joint names to JointState
+/// \param[in] goalPositions The required positions for the fingers
```
mkoval: Nit: This is not specific to finger joints! 😉 

Also, we should add "position" to the name of this function, since we may very well need equivalent functions for velocities and/or efforts. Perhaps `positionsToJointState`?

include/aikido/control/ros/RosPositionCommandExecutor.hpp
```diff
@@ -0,0 +1,75 @@
...
+#include <pr_control_msgs/SetPositionAction.h>
+#include <actionlib/client/action_client.h>
+#include <aikido/control/PositionCommandExecutor.hpp>
+#include <Eigen/Dense>
+
+namespace aikido{
+namespace control {
+namespace ros {
+
+/// This class sends commands for the hand position to the ROS server
```
mkoval: Nit: This comment is not very helpful. At a minimum, we should specify that this class uses `actionlib` to command a `pr_control_msgs/SetPosition` action.

include/aikido/control/ros/RosPositionCommandExecutor.hpp
```diff
@@ -0,0 +1,75 @@
...
+class RosPositionCommandExecutor : public aikido::control::PositionCommandExecutor
+{
+public:
+  /// Constructor
+  /// \param[in] node ROS node handle for action client.
+  /// \param[in] serverName Name of the server to send traejctory to.
```
mkoval: Typo: "traejctory "

include/aikido/control/ros/RosPositionCommandExecutor.hpp
```diff
@@ -0,0 +1,75 @@
...
+  /// \param[in] connectionTimeout Timeout for server connection.
+  /// \param[in] connectionPollingPeriod Polling period for server connection.
+  RosPositionCommandExecutor(
+    ::ros::NodeHandle node,
+    const std::string& serverName,
+    const std::vector<std::string>& jointNames,
+    const std::chrono::milliseconds connectionTimeout = std::chrono::milliseconds{1000},
+    const std::chrono::milliseconds connectionPollingPeriod = std::chrono::milliseconds{20}
```
mkoval: Nit: Const is meaningless on a parameter passed by value in a function declaration.

include/aikido/control/ros/util.hpp
```diff
@@ -0,0 +1,44 @@
...
+  ::ros::CallbackQueue& callbackQueue,
+  TimeoutDuration timeoutDuration = std::chrono::milliseconds{ 1000 },
+  PeriodDuration periodDuration = std::chrono::milliseconds{ 10 }
+)
+{
+  using Clock = std::chrono::steady_clock;
+
+  const auto startTime = Clock::now();
+  const auto endTime = startTime + timeoutDuration;
+  auto currentTime = startTime + periodDuration;
```
mkoval: I _think_ this should be initialized to `startTime`.

include/aikido/control/ros/util.hpp
```diff
@@ -0,0 +1,44 @@
...
+namespace control {
+namespace ros {
+
+template <class ActionSpec, class TimeoutDuration, class PeriodDuration>
+bool waitForActionServer(
+  actionlib::ActionClient<ActionSpec>& actionClient,
+  ::ros::CallbackQueue& callbackQueue,
+  TimeoutDuration timeoutDuration = std::chrono::milliseconds{ 1000 },
+  PeriodDuration periodDuration = std::chrono::milliseconds{ 10 }
+)
```
mkoval: This function needs a docstring. Also, it should only be _declared_ in the `util.hpp` file. The implementation should be `#included` from a `detail/util-impl.hpp` file to match the convention we follow inn the rest of Aikido.

src/control/ros/Conversions.cpp
```diff
@@ -429,6 +429,29 @@ trajectory_msgs::JointTrajectory toRosJointTrajectory(
...
 
+//=============================================================================
+sensor_msgs::JointState toJointState(
+  const Eigen::VectorXd& goalPositions, const std::vector<std::string>& jointNames)
+{
+  if (goalPositions.size() != jointNames.size())
+  {
+    std::stringstream message;
+    message << "The size of goalPositions ("<<goalPositions.size()<<
+      ") must be the same as jointNames ("<<jointNames.size()<<"!" ;
```
mkoval: There are mismatched parens in this error message.

src/control/ros/Conversions.cpp
```diff
@@ -429,6 +429,29 @@ trajectory_msgs::JointTrajectory toRosJointTrajectory(
...
+    throw std::invalid_argument(message.str());
+  }
+
+  sensor_msgs::JointState jointState;
+
+  jointState.name = std::move(jointNames);
```
mkoval: Nit: It is not possible to `std::move` out of `jointNames` because it was passed by `const` reference. You should either remove the `std::move` or modify the function to take `jointNames` by value. I'd slightly prefer the latter.

src/control/ros/RosPositionCommandExecutor.cpp
```diff
@@ -0,0 +1,139 @@
...
+namespace control {
+namespace ros {
+
+using std::chrono::milliseconds;
+
+//=============================================================================
+RosPositionCommandExecutor::RosPositionCommandExecutor(
+      ::ros::NodeHandle node,
+      const std::string& serverName,
+      const std::vector<std::string>& jointNames,
```
mkoval: Nit: Take `jointNames` by value and `std::move` it below.

src/control/ros/RosPositionCommandExecutor.cpp
```diff
@@ -0,0 +1,139 @@
...
+  , mConnectionPollingPeriod{connectionPollingPeriod}
+  , mInProgress{false}
+{
+}
+
+//=============================================================================
+RosPositionCommandExecutor::~RosPositionCommandExecutor()
+{
+  // Do nothing.
+  // TODO: Should we wait for the current trajectory to finish executing?
```
mkoval: Nit: This comment was copied and pasted from `RosTrajectoryExecutor`. There are no trajectories in this file! 😁

include/aikido/control/ros/util.hpp
Shushman: I based it off this : 
https://github.com/personalrobotics/aikido/blob/master/src/control/ros/RosTrajectoryExecutor.cpp#L157
Can you help me figure it out? @gilwoolee any comments?

src/control/ros/RosPositionCommandExecutor.cpp
Shushman: I removed the `const` as well after looking up `std::move`

include/aikido/control/ros/detail/util-impl.hpp
```diff
@@ -0,0 +1,40 @@
...
+  ::ros::CallbackQueue& callbackQueue,
+  TimeoutDuration timeoutDuration = std::chrono::milliseconds{ 1000 },
+  PeriodDuration periodDuration = std::chrono::milliseconds{ 10 }
+)
+{
+  using Clock = std::chrono::steady_clock;
+
+  const auto startTime = Clock::now();
+  const auto endTime = startTime + timeoutDuration;
+  auto currentTime = startTime + periodDuration;
```
mkoval: This should not have `periodDuration` added onto it. Otherwise, the first iteration will have twice the expected period.

src/control/ros/Conversions.cpp
mkoval: This is now resolved.

src/control/ros/RosPositionCommandExecutor.cpp
mkoval: It looks like this is still not resolved.

include/aikido/control/ros/RosPositionCommandExecutor.hpp
mkoval: This is not resolved.

PR #175, opened 2017-04-16
Yaml extension for Eigen::Matrix and std::unordered_map

gilwoolee:
Adds YAML parser for Eigen datatype, originally implemented by @mkoval . Copied from [kenv](https://github.com/personalrobotics/kenv/blob/master/box2d_kenv/include/box2d_kenv/yaml_utils.h). Allows operations such as `node.as<Eigen::VectorXd>, node.as<Eigen::Isometry3d>` , etc..


include/aikido/util/yaml_utils.hpp
```diff
@@ -0,0 +1,202 @@
+#ifndef AIKIDO_UTIL_YAMLUTILS_HPP_
+#define AIKIDO_UTIL_YAMLUTILS_HPP_
+
+#include <functional>
+#include <mutex>
+#include <vector>
+#include <iostream>
+#include <boost/format.hpp>
+#include <boost/make_shared.hpp>
```
mkoval: Of these, I think only `Eigen/Dense`, `yaml-cpp/yaml.h`, and `boost/format.hpp` are actually used. We should remove the unused `#include` directives.

Actually, I'd also like to replace `boost::format` with `std:;stringstream` to eliminate that dependency as well.

include/aikido/util/yaml_utils.hpp
```diff
@@ -0,0 +1,202 @@
...
+
+template <class _Scalar,
+          int _Rows, int _Cols, int _Options, int _MaxRows, int _MaxCols>
+inline void deserialize(
+    YAML::Node const &node,
+    Eigen::Matrix<_Scalar, _Rows, _Cols, _Options, _MaxRows, _MaxCols> &matrix)
+{
+    typedef Eigen::Matrix<_Scalar, _Rows, _Cols, _Options, _MaxRows, _MaxCols> MatrixType;
+    typedef typename MatrixType::Index Index;
+    typedef typename MatrixType::Scalar Scalar;
```
mkoval: Nit: Prefer `using` aliases to `typedef`s.

include/aikido/util/yaml_utils.hpp
```diff
@@ -0,0 +1,202 @@
...
+  {
+    assert(false && "Unknown MatrixType.");
+  }
+};
+
+template <typename MatrixType>
+struct encode_impl<MatrixType, true> {
+  static Node encode(MatrixType const &matrix)
+  {
+    typedef typename MatrixType::Index Index;
```
mkoval: Nit: Prefer `using` aliases to `typedef`s.

include/aikido/util/yaml_utils.hpp
```diff
@@ -0,0 +1,202 @@
...
+    {
+        return convert<MatrixType>::decode(node, transform.matrix());
+    }
+};
+
+template <class T>
+inline void operator>>(Node const &node, T &value)
+{
+    value = node.as<T>();
+}
```
mkoval: I'd like to remove this. If I remember correctly, I added this as a workaround to support both `yaml-cpp` versions 0.3 (which used the bitshift operators) and 0.5 (which used the `as` function).

include/aikido/util/yaml_utils.hpp
```diff
@@ -0,0 +1,202 @@
...
+#include <Eigen/Dense>
+#include <yaml-cpp/yaml.h>
+
+
+namespace aikido {
+namespace util {
+
+template <class _Scalar,
+          int _Rows, int _Cols, int _Options, int _MaxRows, int _MaxCols>
+inline void deserialize(
```
mkoval: Missing docstring. I'm willing to let it slide for the `convert` that we specialize for `yaml-cpp` - but this is purely a helper function, so it should be properly documented. 😁

include/aikido/util/yaml_utils.hpp
```diff
@@ -0,0 +1,202 @@
...
+                        Eigen::Transform<_Scalar, Dim, Mode, _Options> &pose)
+{
+    deserialize(node, pose.matrix());
+}
+
+template <class T>
+inline bool has_child(YAML::Node const &node, T const &key)
+{
+	return node[key].IsDefined();
+}
```
mkoval: This function is not used, please delete it.

include/aikido/util/yaml_utils.hpp
```diff
@@ -0,0 +1,202 @@
...
+        }
+    } else {
+        throw runtime_error(str(
+            format("Unknown type of matrix '%s'.") % node.Tag()));
+    }
+}
+
+template <class _Scalar, int Dim, int Mode, int _Options>
+inline void deserialize(YAML::Node const &node,
+                        Eigen::Transform<_Scalar, Dim, Mode, _Options> &pose)
```
mkoval: Missing docstring. See above. 😉

include/aikido/util/yaml_utils.hpp
```diff
@@ -0,0 +1,202 @@
...
+
+} // namespace util
+} // namespace aikido
+
+
+namespace YAML {
+namespace detail {
+
+template <typename MatrixType, bool IsVectorAtCompileTime>
+struct encode_impl {
```
mkoval: This is an internal helper class, so it deserves a docstring. 😁

The reason this exists is so we can use template specialization to switch between serializing vectors and serializing matrices based on the `IsVectorAtCompileTime` flag. This is a "nice to have" that lets us generate `!Vector [1, 2, 3]` instead of the syntactic travesty like `!Matrix [[1], [2] ,[3]]` when serializing an Eigen vector.

include/aikido/util/yaml_utils.hpp
```diff
@@ -0,0 +1,202 @@
...
+          row.push_back(matrix(r, c));
+      }
+
+      node.push_back(row);
+    }
+    return node;
+  }
+};
+
+} //detail
```
mkoval: Nit: We _may_ want to move this helper code into the `detail` directory. @gilwoolee @jslee02 Thoughts?

jslee02:
> This file is a bit of a mess (as it was in muul). Does it make sense to split some of this into the detail directory? @gilwoolee @jslee02 does this make sense to you, and if so, which parts?

I believe none of these functions needs to be exposed as public API. So all the code should be in `detail` I think. Also, we might mostly want to enable to use `yaml` with Eigen type. For this, what I would do is like:
* move this whole file to `detail` directory
* create `yaml_cpp.cpp` that includes both `<yaml-cpp/yaml.h>` and this file
* encourage Aikido to use `yaml_cpp.hpp` all the time rather than `<yaml-cpp/yaml.h>`

More importantly, we already [have](https://github.com/personalrobotics/aikido/blob/master/include/aikido/perception/eigen_yaml.hpp) this file in `perception` component in a different name. :joy:

include/aikido/util/yaml_utils.hpp
```diff
@@ -0,0 +1,202 @@
...
+
+
+namespace aikido {
+namespace util {
+
+template <class _Scalar,
+          int _Rows, int _Cols, int _Options, int _MaxRows, int _MaxCols>
+inline void deserialize(
+    YAML::Node const &node,
+    Eigen::Matrix<_Scalar, _Rows, _Cols, _Options, _MaxRows, _MaxCols> &matrix)
```
jslee02: We might want to follow the suggestion of Eigen's "[Writing Functions Taking Eigen Types as Parameters](https://eigen.tuxfamily.org/dox/TopicFunctionTakingEigenTypes.html)" here.

jslee02:
TODOs:
* [ ] add `unordered_map` conversion
* [ ] add docstrings

include/aikido/util/yaml_utils.hpp
jslee02: I believe we don't use this function out of `template <...> struct convert`. Could I just move this into it as you said?

jslee02:
I believe I've addressed all @mkoval 's comments.

Question: It seems the `yaml-cpp`'s convention is to return `false` when failed to decode rather than throwing an exception, but the decoder for `Eigen::Matrix` in this PR throws exceptions for the case. So I would like to change it to follow the convention. @mkoval How do you think?

include/aikido/util/detail/yaml_extension.hpp
```diff
@@ -0,0 +1,246 @@
...
+        row.push_back(matrix(r, c));
+
+      node.push_back(row);
+    }
+
+    return node;
+  }
+};
+
+} // namespace (anonymous)
```
mkoval: Don't use anonymous namespaces in headers. I suggest putting this in a `detail` namespace instead.

include/aikido/util/detail/yaml_extension.hpp
```diff
@@ -0,0 +1,246 @@
...
+        matrix);
+  }
+
+  static bool decode(
+      const YAML::Node& node,
+      Eigen::Matrix<_Scalar, _Dim, _Mode, _Options>& matrix)
+  {
+    if (node.Tag() == "Vector" || node.Tag() == "!Vector"
+        || node.Tag() == "Matrix"
+        || node.Tag() == "!Matrix")
```
mkoval: I am _pretty_ sure that we only need the set with the `!` or the one without. If I remember correctly, I have both here because the behavior changed between `yaml-cpp` 0.3 and 0.5. Aikido only supports `yaml-cpp` 0.5, so that is a moot point.

Unfortunately, I don't remember which is used in 0.5. I'd like to do a quick test and eliminate the superfluous check.

include/aikido/util/detail/yaml_extension.hpp
```diff
@@ -0,0 +1,246 @@
...
+  }
+
+  static bool decode(const Node& node, UnorderedMap& map)
+  {
+    if (!node.IsMap())
+      return false;
+
+    map.clear();
+    for (const auto& it : node)
+      map[it.first.as<_Key>()] = it.second.as<_Tp>();
```
mkoval: What happens if `as<>()` fails? If we want to return `false` on failure, then we need to handle that case explicitly.

include/aikido/util/detail/yaml_extension.hpp
```diff
@@ -0,0 +1,246 @@
...
+
+    return node;
+  }
+
+  static bool decode(const Node& node, UnorderedMap& map)
+  {
+    if (!node.IsMap())
+      return false;
+
+    map.clear();
```
mkoval: Nit: Call `map.reserve(node.size())`.

psigen:
I do not think it is a good idea to introduce exceptions into the `yaml-cpp` API.  This is going to screw up the type of failure that users are expecting to need to handle in parsing code.

And doesn't the `yaml-cpp` `.as<>` wrapper raise a `TypedBadConversion` if the internal converter returns _false_ anyway?  That exception will include additional information like where the conversion error happened _and_ the type.

https://github.com/jbeder/yaml-cpp/blob/master/include/yaml-cpp/node/impl.h#L125

jslee02:
I agree what @mkoval and @psigen said. It seems we shouldn't introduce an exception for `struct convert` as `yaml-cpp` does and let `.as<>` raise a `TypedBadConver` exception when `decode` returns false.

Also, I like the idea not to use additional tag.

include/aikido/util/detail/yaml_extension.hpp
jslee02: I thought we use anonymous namespace for objects that are only used in the implementation, and use `detail` namespace for objects that are used somewhere out of the implementation but don't expect that is used by the user.

In most cases, the implementation is placed in a source file, but we have this `decode` in a header file because it's a template. I think we could consider putting this into a separate file with the name `yaml_extension-impl.hpp` instead.

include/aikido/util/detail/yaml_extension.hpp
jslee02: I'm not sure either, but it seems the spec says `!` should be used. So let's stick with that. :sweat_smile:

jslee02:
Hm, after I looking into more `yaml-cpp` code, I started to incline using exceptions because it throws exception anyway and  `.as<>` throws exceptions with insufficient information of the error. It only shows where (line and column) the exception occurred but not what the exact cause. It only says "bad conversion". 

As a consensus, I would like to use exceptions but `yaml-cpp`'s custom exception object, [`RepresentationException`](https://github.com/jbeder/yaml-cpp/blob/master/include/yaml-cpp/exceptions.h#L144) that allows us to pass the error message.

include/aikido/util/detail/yaml_extension.hpp
mkoval: I don't think we should _ever_ put an anonymous `namespace` in a header file. A header file can be `#include`ed into more than one translation unit, so this will result in unnecessary duplication of the objects (in the best case) or difficult-to-diagnose bugs (in the worst case, e.g. possibly if the namespace contains `static` member or global variables).

I see the `detail` namespace as the logical equivalent of an anonymous namespace in the header file, i.e. it is an implementation detail of other functions in that file.

mkoval:
I agree with @jslee02's conclusion - let's keep exceptions, but switch from `std::runtime_error` to `YAML::Exception`. Like @jslee02 said, `yaml-cpp` generates terrible error messages.

jslee02:
I believe this PR is ready to merge.

One remained an issue but not a blocker is removing the requirement of tag for vector and matrix. I'll create an issue for this to implement this in the future.

include/aikido/util/detail/yaml_extension.hpp
```diff
@@ -0,0 +1,248 @@
...
+};
+
+inline YAML::Mark getMark(const YAML::Node& node)
+{
+#ifdef YAMLCPP_NODE_HAS_MARK
+  return node.Mark();
+#else
+  DART_UNUSED(node);
+  return YAML::Mark::null_mark();
+#endif
```
mkoval: Why is this `#ifdef` necessary? Can you add a comment explaining why?

include/aikido/util/detail/yaml_extension.hpp
```diff
@@ -0,0 +1,248 @@
...
+//==============================================================================
+// Specialization for Eigen::Matrix<...>. This enables to use YAML::Node
+// with std::unordered_map as `Node(Eigen::Matrix3d::Identity())` and
+// `node.as<Eigen::Matrix3d>()`.
+template <typename _Scalar,
+          int _Rows,
+          int _Cols,
+          int _Options,
+          int _MaxRows,
+          int _MaxCols>
```
mkoval: @jslee02 Did you determine whether if or not we can replace this with `template <class _Derived> struct<Eigen::MatrixBase<_Derived>>`? I really liked that idea.

jslee02: Yeah, I haven't figured it out yet, but it doesn't work for struct. :disappointed: Let's put this over for now.

PR #177, opened 2017-04-17
changed frame for HERB3

pengjujin:
changed the frame used by rviz and took out  the unneeded print statements in perception

### DIFF
--- a/src/perception/AprilTagsModule.cpp
+++ b/src/perception/AprilTagsModule.cpp
@@ -116,7 +116,6 @@ bool AprilTagsModule::detectObjects(const dart::simulation::WorldPtr& env, ros::
 				
 			if(env_skeleton == nullptr){
 				// Getting the model for the new object
-				std::cout << "getting new thing" << std::endl;
 				is_new_skel = true;
 				dart::utils::DartLoader urdfLoader;
 				skel_to_update = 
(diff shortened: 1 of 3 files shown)


mkoval:
No! This is **not** the correct way to fix this. Please revert this change ASAP.

mkoval:
The correct way to make this change is to make the `frame_id` a variable. This is actually relativley painless to implement by adding a `frame_id` argument to the `InteractiveMarkerViewer` constructor that gets passed down to all of the other `*Marker` classes that it constructs (for `BodyNode`s, etc). 

More importantly, switching from `frame_id` to `base_footprint` is likely *not* the correct fix for whatever the problem is. We typically use `map` as the "fixed frame" for visualization, collision checking, etc. because it remains stationary while the robot moves (or, at least, _should_ remain stationary if HERB's localization system is working properly). If we change this frame to `base_footprint`, I suspect that the visualization will be quite broken as soon as HERB begins to move around the environment.

If the issue is simply that the visualization does not work when localization is not running, then we probably should start a `static_transform_publisher` that emulates the localization stack by publishing an identity transformation from `/map` to `/odom` and `/odom` to `/base_link`.

brianhou:
@pengjujin can you make these changes? Thanks!

PR #178, opened 2017-04-17
Fix ompl tests builds

jslee02:
`OMPL_FOUND` is only defined in the scope of `<aikido_dir>/src/planner/ompl` so `tests/planner/CMakeListx.txt` couldn't check if ompl is found. This PR instead check whether `aikido_planner_ompl` target is built to decide whether Aikido builds the ompl tests or not.

This problem was discovered by @aditya-vk .


aditya-vk:
I don't much about this but the edit does not still generate the executables for tests defined under `tests/planner/ompl` like it does for `tests/planner/parabolic`.

jslee02:
So far, ompl test were not being tests. Many of errors are fixed by 372517b.

PR #179, opened 2017-04-18
Revert "changed frame for HERB3"

brianhou:
Reverts personalrobotics/aikido#177


mkoval:
Thanks for reverting this so quickly! 👏

PR #180, opened 2017-04-18
Add satisfiable tolerance parameter to TSR constructor

dqyi11:


jslee02:
I think we could simply pass the tolerance to `TSR::isSatisfied(...)`. How do you think @mkoval ?

dqyi11:
@jslee02   isSatisfied() is inherited from a base class Testable. 
We will still have to add similar arguments in the constructors of adapter classes unless we change isSatisfiable() of Testable.

jslee02:
> We will still have to add similar arguments in the constructors of adapter classes unless we change isSatisfiable() of Testable.

:+1: I thought other `Testable` classes also need some tolerance but it seems not.

include/aikido/constraint/TSR.hpp
```diff
@@ -44,7 +44,8 @@ class TSR : public Sampleable,
       const Eigen::Isometry3d& _T0_w = Eigen::Isometry3d::Identity(),
       const Eigen::Matrix<double, 6, 2>& _Bw =
           Eigen::Matrix<double, 6, 2>::Zero(),
-      const Eigen::Isometry3d& _Tw_e = Eigen::Isometry3d::Identity());
+      const Eigen::Isometry3d& _Tw_e = Eigen::Isometry3d::Identity(),
+      const double& _satisfiableTolerance = 1e-6);
```
jslee02: Nit: Passing const reference for primitive type is pointless. Please change to `double`.

include/aikido/constraint/TSR.hpp
```diff
@@ -57,7 +58,8 @@ class TSR : public Sampleable,
   TSR(const Eigen::Isometry3d& _T0_w = Eigen::Isometry3d::Identity(),
       const Eigen::Matrix<double, 6, 2>& _Bw =
           Eigen::Matrix<double, 6, 2>::Zero(),
-      const Eigen::Isometry3d& _Tw_e = Eigen::Isometry3d::Identity());
+      const Eigen::Isometry3d& _Tw_e = Eigen::Isometry3d::Identity(),
+      const double& _satisfiableTolerance = 1e-6);
```
jslee02: ditto

include/aikido/constraint/TSR.hpp
```diff
@@ -124,9 +126,13 @@ class TSR : public Sampleable,
   /// This often represent an offset from "w" to the origin of the end-effector.
   Eigen::Isometry3d mTw_e;
 
+  /// Tolerance for checking satisfiability
+  double mSatisfiableTolerance;
+
 private:
   std::unique_ptr<util::RNG> mRng;
   std::shared_ptr<statespace::SE3> mStateSpace;
+
```
jslee02: Nit: Could we remove this empty line?

src/constraint/TSR.cpp
```diff
@@ -79,25 +79,28 @@ class TSRSampleGenerator : public SampleGenerator
...
 TSR::TSR(std::unique_ptr<util::RNG> _rng, const Eigen::Isometry3d& _T0_w,
-         const Eigen::Matrix<double, 6, 2>& _Bw, const Eigen::Isometry3d& _Tw_e)
+         const Eigen::Matrix<double, 6, 2>& _Bw, const Eigen::Isometry3d& _Tw_e,
+         const double& _satisfiableTolerance)
     : mT0_w(_T0_w)
     , mBw(_Bw)
     , mTw_e(_Tw_e)
     , mRng(std::move(_rng))
     , mStateSpace(std::make_shared<SE3>())
+    , mSatisfiableTolerance(_satisfiableTolerance)
```
jslee02: Initialize `mSatisfiableTolerance` before `mRng` and `mStateSpace` to resolve the [warning](https://travis-ci.org/personalrobotics/aikido/builds/223337806#L2840).

src/constraint/TSR.cpp
```diff
@@ -79,25 +79,28 @@ class TSRSampleGenerator : public SampleGenerator
...
 TSR::TSR(const Eigen::Isometry3d& _T0_w, const Eigen::Matrix<double, 6, 2>& _Bw,
-         const Eigen::Isometry3d& _Tw_e)
+         const Eigen::Isometry3d& _Tw_e, const double& _satisfiableTolerance)
     : mT0_w(_T0_w)
     , mBw(_Bw)
     , mTw_e(_Tw_e)
     , mRng(std::unique_ptr<util::RNG>(
           new util::RNGWrapper<std::default_random_engine>(0)))
     , mStateSpace(std::make_shared<SE3>())
+    , mSatisfiableTolerance(_satisfiableTolerance)
```
jslee02: ditto

src/constraint/TSR.cpp
```diff
@@ -109,6 +112,7 @@ TSR::TSR(const TSR& other)
     , mTw_e(other.mTw_e)
     , mRng(std::move(other.mRng->clone()))
     , mStateSpace(std::make_shared<SE3>())
+    , mSatisfiableTolerance(other.mSatisfiableTolerance)
```
jslee02: ditto

src/constraint/TSR.cpp
```diff
@@ -120,6 +124,7 @@ TSR::TSR(TSR&& other)
     , mTw_e(other.mTw_e)
     , mRng(std::move(other.mRng))
     , mStateSpace(std::make_shared<SE3>())
+    , mSatisfiableTolerance(other.mSatisfiableTolerance)
```
jslee02: ditto

dqyi11:
@jslee02  I addressed all your comments. Why does the order matter? 
Also I put it as a public variable, because I see other member variables are public. 
Shall I move it to private/protected and write get/set?

include/aikido/constraint/TSR.hpp
```diff
@@ -44,7 +44,8 @@ class TSR : public Sampleable,
       const Eigen::Isometry3d& _T0_w = Eigen::Isometry3d::Identity(),
       const Eigen::Matrix<double, 6, 2>& _Bw =
           Eigen::Matrix<double, 6, 2>::Zero(),
-      const Eigen::Isometry3d& _Tw_e = Eigen::Isometry3d::Identity());
+      const Eigen::Isometry3d& _Tw_e = Eigen::Isometry3d::Identity(),
+      const double _satisfiableTolerance = 1e-6);
```
jslee02: Nit: `const` is meaningless for primitive types please change this to `double _satisfiableTolerance = 1e-6`

include/aikido/constraint/TSR.hpp
```diff
@@ -57,7 +58,8 @@ class TSR : public Sampleable,
   TSR(const Eigen::Isometry3d& _T0_w = Eigen::Isometry3d::Identity(),
       const Eigen::Matrix<double, 6, 2>& _Bw =
           Eigen::Matrix<double, 6, 2>::Zero(),
-      const Eigen::Isometry3d& _Tw_e = Eigen::Isometry3d::Identity());
+      const Eigen::Isometry3d& _Tw_e = Eigen::Isometry3d::Identity(),
+      const double _satisfiableTolerance = 1e-6);
```
jslee02: ditto

include/aikido/constraint/TSR.hpp
```diff
@@ -112,6 +114,10 @@ class TSR : public Sampleable,
   bool project(const statespace::StateSpace::State* _s,
       statespace::StateSpace::State* _out) const override;
 
+  double getSatisfiableTolerance();
+   
+  void setSatisfiableTolerance(const double satisfiableTolerance);
```
jslee02: Nit: missing docstring

include/aikido/constraint/TSR.hpp
```diff
@@ -112,6 +114,10 @@ class TSR : public Sampleable,
   bool project(const statespace::StateSpace::State* _s,
       statespace::StateSpace::State* _out) const override;
 
+  double getSatisfiableTolerance();
```
jslee02: Nit: missing docstring

src/constraint/TSR.cpp
```diff
@@ -79,10 +79,12 @@ class TSRSampleGenerator : public SampleGenerator
 
 //=============================================================================
 TSR::TSR(std::unique_ptr<util::RNG> _rng, const Eigen::Isometry3d& _T0_w,
-         const Eigen::Matrix<double, 6, 2>& _Bw, const Eigen::Isometry3d& _Tw_e)
+         const Eigen::Matrix<double, 6, 2>& _Bw, const Eigen::Isometry3d& _Tw_e,
+         const double _satisfiableTolerance)
```
jslee02: ditto

src/constraint/TSR.cpp
```diff
@@ -91,10 +93,11 @@ TSR::TSR(std::unique_ptr<util::RNG> _rng, const Eigen::Isometry3d& _T0_w,
 
 //=============================================================================
 TSR::TSR(const Eigen::Isometry3d& _T0_w, const Eigen::Matrix<double, 6, 2>& _Bw,
-         const Eigen::Isometry3d& _Tw_e)
+         const Eigen::Isometry3d& _Tw_e, const double _satisfiableTolerance)
```
jslee02: ditto

include/aikido/constraint/TSR.hpp
gilwoolee: Need docstring for the new variable.

include/aikido/constraint/TSR.hpp
```diff
@@ -112,6 +114,14 @@ class TSR : public Sampleable,
   bool project(const statespace::StateSpace::State* _s,
       statespace::StateSpace::State* _out) const override;
 
+  /// Get the staisfiable tolerance for isSatisfiable.
+  /// \param[out] _out Satisfialble tolerance, double.
+  double getSatisfiableTolerance();
```
gilwoolee: Change this to `getTestableTolerance()`. Although, I wonder why these set and get methods are necessary. Why do we want to change the tolerance once a TSR is created?

include/aikido/constraint/TSR.hpp
```diff
@@ -112,6 +114,14 @@ class TSR : public Sampleable,
...
+  
+  /// Set the staisfiable tolerance for isSatisfiable.
+  /// \param _satisfiableTolerance Satisfiable tolerance to set.
```
gilwoolee: Typo. Though not sure why we need this method.

include/aikido/constraint/TSR.hpp
```diff
@@ -124,7 +134,10 @@ class TSR : public Sampleable,
   /// This often represent an offset from "w" to the origin of the end-effector.
   Eigen::Isometry3d mTw_e;
 
+
 private:
+  /// Tolerance for checking satisfiability
+  double mSatisfiableTolerance;
```
gilwoolee: Consider changing this to `Eigen::Vector6d`

dqyi11: Yes, I also think Eigen::Vector6d makes more sense.

dqyi11:
I have addressed @gilwoolee 's comments.
I have not changed it to Vector6d, because the current implementation takes the norm of the distance vector to compare with a double threshold. I will make the change if comparing two vectors is expected.
Set() and Get() are still kept. Will keep or remove after hearing back from @mkoval  and @jslee02 .

include/aikido/constraint/TSR.hpp
```diff
@@ -44,7 +44,8 @@ class TSR : public Sampleable,
       const Eigen::Isometry3d& _T0_w = Eigen::Isometry3d::Identity(),
```
mkoval: Add a docstring for `_testableTolerance `.

include/aikido/constraint/TSR.hpp
```diff
@@ -44,7 +44,8 @@ class TSR : public Sampleable,
...
       const Eigen::Matrix<double, 6, 2>& _Bw =
           Eigen::Matrix<double, 6, 2>::Zero(),
-      const Eigen::Isometry3d& _Tw_e = Eigen::Isometry3d::Identity());
+      const Eigen::Isometry3d& _Tw_e = Eigen::Isometry3d::Identity(),
+      const double _testableTolerance = 1e-6);
```
mkoval: Don't mark arguments that are passed by value as `const` in function declarations.

include/aikido/constraint/TSR.hpp
```diff
@@ -57,7 +58,8 @@ class TSR : public Sampleable,
   TSR(const Eigen::Isometry3d& _T0_w = Eigen::Isometry3d::Identity(),
       const Eigen::Matrix<double, 6, 2>& _Bw =
           Eigen::Matrix<double, 6, 2>::Zero(),
-      const Eigen::Isometry3d& _Tw_e = Eigen::Isometry3d::Identity());
+      const Eigen::Isometry3d& _Tw_e = Eigen::Isometry3d::Identity(),
+      const double _testableTolerance = 1e-6);
```
mkoval: Don't mark arguments that are passed by value as `const` in function declarations.

include/aikido/constraint/TSR.hpp
```diff
@@ -57,7 +58,8 @@ class TSR : public Sampleable,
   TSR(const Eigen::Isometry3d& _T0_w = Eigen::Isometry3d::Identity(),
```
mkoval: Add a docstring for `_testableTolerance `.

include/aikido/constraint/TSR.hpp
```diff
@@ -112,6 +114,14 @@ class TSR : public Sampleable,
   bool project(const statespace::StateSpace::State* _s,
       statespace::StateSpace::State* _out) const override;
 
+  /// Get the testable tolerance used in isSatisfiable.
+  /// \param[out] _out Testable tolerance, double.
+  double getTestableTolerance();
+  
+  /// Set the testable tolerance used in isSatisfiable.
+  /// \param _testableTolerance Testable tolerance to set.
+  void setTestableTolerance(const double _testableTolerance);
```
mkoval: Don't mark arguments that are passed by value as `const` in function declarations.

jslee02:
As a third option, we could provide either way by providing `class TSR::ErrorMethod` that has a virtual function that measures the error and concrete classes that implement the error measurements. Then use a reasonable one as the default.

include/aikido/constraint/TSR.hpp
dqyi11: Fixed

include/aikido/constraint/TSR.hpp
dqyi11: Fixed

include/aikido/constraint/TSR.hpp
dqyi11: added.

include/aikido/constraint/TSR.hpp
dqyi11: Fixed

include/aikido/constraint/TSR.hpp
dqyi11: Fixed

gilwoolee:
Based on what @mkoval said, `double` tolerance seems to be fine since it's only about _numerical precision_. @jslee02 's suggestion seems reasonable, but it may be an overcomplication if this is only for numerical precision.

I'm still not convinced that we need `setTolerance` and `getTolerance`.  Why do we want to change tolerance after initializing it? @mkoval @jslee02 thoughts?

jslee02:
> Why do we want to change tolerance after initializing it? @mkoval @jslee02 thoughts?

I don't have a compelling reason to have the getters/setters but also not to have. One use case I can think of now is doing planning changing the error tolerance when it's failed to see if the failure is because of it. I'm fine either leaving them or not.

PR #182, opened 2017-04-19
Removes locking from step()

gilwoolee:
`KinematicSimulationTrajectoryExecutor::step()` should no longer lock the skeleton and let the caller be responsible for locking if it's running under multi-threaded setting. This is consistent with other executors.

### DIFF
--- a/include/aikido/control/BarrettHandKinematicSimulationPositionCommandExecutor.hpp
+++ b/include/aikido/control/BarrettHandKinematicSimulationPositionCommandExecutor.hpp
@@ -47,7 +47,9 @@ public:
   /// \return Future which becomes available when the execution completes.
   std::future<void> execute(const Eigen::VectorXd& goalPositions) override;
 
-  // Documentation inherited.
+  /// If multiple threads are accessing this function or the skeleton associated
+  /// with this executor, it is necessary to lock the skeleton before
+  /// calling this method.
   void step() override;
 
   /// Resets CollisionGroup to check collision with fingers.
--- a/include/aikido/control/BarrettHandKinematicSimulationPositionCommandExecutor.hpp
+++ b/include/aikido/control/BarrettHandKinematicSimulationPositionCommandExecutor.hpp
@@ -47,6 +47,8 @@ public:
   /// \return Future which becomes available when the execution completes.
   std::future<void> execute(const Eigen::VectorXd& goalPositions) override;
 
+  /// \copydoc PositionCommandExecutor::step()
+  ///
   /// If multiple threads are accessing this function or the skeleton associated
   /// with this executor, it is necessary to lock the skeleton before
   /// calling this method.
--- a/include/aikido/control/KinematicSimulationTrajectoryExecutor.hpp
+++ b/include/aikido/control/KinematicSimulationTrajectoryExecutor.hpp
@@ -35,7 +35,9 @@ public:
   std::future<void> execute(
     trajectory::TrajectoryPtr traj) override;
 
-  // Documentation inherited.
+  /// If multiple threads are accessing this function or the skeleton associated
+  /// with this executor, it is necessary to lock the skeleton before
+  /// calling this method.
   void step() override;
 
 private:
--- a/include/aikido/control/KinematicSimulationTrajectoryExecutor.hpp
+++ b/include/aikido/control/KinematicSimulationTrajectoryExecutor.hpp
@@ -35,6 +35,8 @@ public:
   std::future<void> execute(
     trajectory::TrajectoryPtr traj) override;
 
+  /// \copydoc PositionCommandExecutor::step()
+  ///
   /// If multiple threads are accessing this function or the skeleton associated
   /// with this executor, it is necessary to lock the skeleton before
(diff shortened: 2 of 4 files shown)


include/aikido/control/BarrettHandKinematicSimulationPositionCommandExecutor.hpp
```diff
@@ -47,7 +47,9 @@ class BarrettHandKinematicSimulationPositionCommandExecutor
   /// \return Future which becomes available when the execution completes.
   std::future<void> execute(const Eigen::VectorXd& goalPositions) override;
 
-  // Documentation inherited.
+  /// If multiple threads are accessing this function or the skeleton associated
+  /// with this executor, it is necessary to lock the skeleton before
+  /// calling this method.
```
jslee02: Nit: I believe this comment is a note to be _added_ to the original docstring of `PositionCommandExecutor::step()`, but this will just overwrite the original docstring. It would be nice to copy the original docstring and add this node as:
```cpp
  /// \copydoc PositionCommandExecutor::step()
  ///
  /// If multiple threads are accessing this function or the skeleton associated
  /// with this executor, it is necessary to lock the skeleton before
  /// calling this method.
```

Also, it would be great to actually add a docstring to `PositionCommandExecutor::step()`. Currently, it's `// Step once`. If it's enough then make it doxygen comment by prepending one more `/`.

include/aikido/control/KinematicSimulationTrajectoryExecutor.hpp
```diff
@@ -35,7 +35,9 @@ class KinematicSimulationTrajectoryExecutor : public TrajectoryExecutor
   std::future<void> execute(
     trajectory::TrajectoryPtr traj) override;
 
-  // Documentation inherited.
+  /// If multiple threads are accessing this function or the skeleton associated
+  /// with this executor, it is necessary to lock the skeleton before
+  /// calling this method.
```
jslee02: ditto

src/control/KinematicSimulationTrajectoryExecutor.cpp
```diff
@@ -106,7 +106,7 @@ void KinematicSimulationTrajectoryExecutor::step()
   }
 
   auto timeSinceBeginning = system_clock::now() - mExecutionStartTime;
-  double t = duration_cast<seconds>(timeSinceBeginning).count();
+  auto t = duration<double>(timeSinceBeginning).count();
```
jslee02: Nit: I believe the unit of time here is seconds. It would be nice to either (1) change the variable name to encode the unit or (2) add trailing comment something like `// sec`.

Additionally, it would be much better to comment the unit of the time parameter to the docstring of `Trajectory::evaluate()`.

src/control/KinematicSimulationTrajectoryExecutor.cpp
```diff
@@ -129,6 +126,7 @@ void KinematicSimulationTrajectoryExecutor::step()
     mPromise->set_value();
     mInExecution = false;
   }
+
```
jslee02: Nit: Could we remove this empty line?

src/control/KinematicSimulationTrajectoryExecutor.cpp
mkoval: Why was this changed to `duration` instead of `duration_cast?` I believe `timeSinceBeginning` is a duration, so `duration_cast` was appropriate here. 🤔

src/control/KinematicSimulationTrajectoryExecutor.cpp
```diff
@@ -105,8 +105,9 @@ void KinematicSimulationTrajectoryExecutor::step()
     mInExecution = false;
   }
 
+  using seconds = std::chrono::duration<float>;
```
jslee02: Nit: I believe the type of time should be `double` as `Trajectory` takes it as `double`.

include/aikido/trajectory/Trajectory.hpp
```diff
@@ -50,8 +50,8 @@ class Trajectory
   /// function is implementation-defined if \c _t is not between
   /// \c getStartTime() and \c getEndTime().
   ///
-  /// \param _t time parameter
-  /// \param[out] _state output state of the trajectory at time \c _t
+  /// \param _t time parameter (in seconds)
+  /// \param[out] _state output state of the trajectory at time \c _t (seconds)
   virtual void evaluate(double _t,
     statespace::StateSpace::State *_state) const = 0;
```
mkoval: This isn't really "in seconds" -it's an arbitrary path parameter.

src/control/KinematicSimulationTrajectoryExecutor.cpp
```diff
@@ -105,8 +105,9 @@ void KinematicSimulationTrajectoryExecutor::step()
     mInExecution = false;
   }
 
+  using seconds = std::chrono::duration<double>;
```
mkoval: This is a potentially confusing alias, because `std::chrono::seconds` stores integers and this alias stores `double`s.

PR #184, opened 2017-04-21
Add StateSaver class.

brianhou:
RAII class suggested by @mkoval in personalrobotics/libherb#26.


include/aikido/statespace/dart/detail/MetaSkeletonStateSpaceSaver-impl.hpp
```diff
@@ -0,0 +1,18 @@
...
+
+MetaSkeletonStateSpaceSaver::MetaSkeletonStateSpaceSaver(MetaSkeletonStateSpacePtr _space)
+: mSpace(std::move(_space))
+, mPositions(mSpace->getMetaSkeleton()->getPositions())
+{
+}
+
+MetaSkeletonStateSpaceSaver::~MetaSkeletonStateSpaceSaver()
+{
+  mSpace->getMetaSkeleton()->setPositions(mPositions);
```
jslee02: Nit: It would be nice if we have a simple validity check whether the size of `mPositions` is still equal to the dofs of the `MetaSkeleton` (if not, we print a warning). I understand it wouldn't be able to catch other structural changes such as changing the joint type with the same dofs, but at least we could catch dimension changing cases.

PR #195, opened 2017-04-27
Test Aikido for release/debug modes and on macOS

jslee02:
This PR:
* builds and runs Aikido tests once per build (removed duplicate builds for coveralls)
* tests Aikido for both of release and debug mode in parallel
* tests Aikido on macOS in parallel.
* fixes ompl tests with OMPL 1.3.0
* suppresses warnings per Apple Clang 4.0.0
  * -Wpessimizing-move: when a local variable is moved on return and the return type is the same as the variable. Copy elision cannot take place with a move, but can take place if the variable is returned directly.
  * -Wmissing-braces: I'm not sure if this is a bug of Cang or a correct warning (relevant [SO](http://stackoverflow.com/questions/31555584/why-is-clang-warning-suggest-braces-around-initialization-of-subobject-wmis)).


.ci/install_macos.sh
```diff
@@ -0,0 +1,5 @@
+brew install boost
+brew install dartsim/dart/dartsim6
+brew install ompl --with-eigen
+brew install tinyxml2
+brew install yaml-cpp
```
mkoval: Nit: Why doesn't `rosdep` install these dependencies on macOS?

jslee02: Hm, is `rosdep` available on macOS?

Issue #197, opened 2017-04-27
Create custom FindRviz.cmake


psigen:
Due to the issue here:
https://github.com/personalrobotics/aikido/blob/master/src/rviz/CMakeLists.txt#L10

it seems like maybe we should create a `FindRviz.cmake` that contains this logic and put it in `./cmake` so that we can simply call `find_package(rviz)` as needed.

jslee02:
:+1: to the idea.

I wonder which one should be used for the library name between `interactive_markers` and `rvis`, though. :confused:

psigen:
Oh, probably should use `interactive_markers`, to override the existing non-working one?

jslee02:
Created PR #198 for this.

PR #198, opened 2017-04-27
Use custom CMake modules for interactive_markers and tf

jslee02:
as suggested by @psigen in #197

### DIFF
--- a/src/perception/CMakeLists.txt
+++ b/src/perception/CMakeLists.txt
@@ -19,12 +19,9 @@ aikido_check_package(std_msgs "perception" "std_msgs")
 find_package(visualization_msgs QUIET)
 aikido_check_package(visualization_msgs "perception" "visualization_msgs")
 
-# TODO: Workaround because find_package(tf) imports "tests" and "run_tests"
-# targets that conflict with the targets defined by Aikido.
-find_path(tf_INCLUDE_DIRS tf PATHS "/opt/ros/indigo/include"
-  PATH_SUFFIXES "interactive_markeres/interactive_marker_client.h")
-find_library(tf_LIBRARIES tf PATHS "/opt/ros/indigo/lib")
-evaluate_condition(tf_FOUND tf_INCLUDE_DIRS AND tf_LIBRARIES)
+# Note: Intentionally use "Module" mode because the config file imports "gtest",
+# "tests", and "run_tests" that conflict with the targets defined by Aikido.
+find_package(tf QUIET MODULE)
 aikido_check_package(tf "perception" "tf")
 
 #==============================================================================
--- a/src/rviz/CMakeLists.txt
+++ b/src/rviz/CMakeLists.txt
@@ -7,15 +7,9 @@ aikido_check_package(roscpp "rviz" "roscpp")
 find_package(geometry_msgs QUIET)
 aikido_check_package(geometry_msgs "rviz" "geometry_msgs")
 
-# TODO: Workaround because find_package(interactive_markers) imports "tests"
-# and "run_tests" targets that conflict with the targets defined by Aikido.
-find_path(interactive_markers_INCLUDE_DIRS interactive_markers
-  PATHS "/opt/ros/indigo/include"
-  PATH_SUFFIXES "interactive_markeres/interactive_marker_client.h")
-find_library(interactive_markers_LIBRARIES interactive_markers
-  PATHS "/opt/ros/indigo/lib")
-evaluate_condition(interactive_markers_FOUND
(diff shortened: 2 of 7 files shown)


src/perception/CMakeLists.txt
```diff
@@ -19,12 +19,9 @@ aikido_check_package(std_msgs "perception" "std_msgs")
 find_package(visualization_msgs QUIET)
 aikido_check_package(visualization_msgs "perception" "visualization_msgs")
 
-# TODO: Workaround because find_package(tf) imports "tests" and "run_tests"
-# targets that conflict with the targets defined by Aikido.
-find_path(tf_INCLUDE_DIRS tf PATHS "/opt/ros/indigo/include"
-  PATH_SUFFIXES "interactive_markeres/interactive_marker_client.h")
```
psigen: Do we need to put these path hints in the `Findtf.cmake` file?  They are definitely good defaults.

src/rviz/CMakeLists.txt
```diff
@@ -7,15 +7,9 @@ aikido_check_package(roscpp "rviz" "roscpp")
 find_package(geometry_msgs QUIET)
 aikido_check_package(geometry_msgs "rviz" "geometry_msgs")
 
-# TODO: Workaround because find_package(interactive_markers) imports "tests"
-# and "run_tests" targets that conflict with the targets defined by Aikido.
-find_path(interactive_markers_INCLUDE_DIRS interactive_markers
-  PATHS "/opt/ros/indigo/include"
-  PATH_SUFFIXES "interactive_markeres/interactive_marker_client.h")
-find_library(interactive_markers_LIBRARIES interactive_markers
-  PATHS "/opt/ros/indigo/lib")
```
psigen: Same with these path hints.

psigen: Note: Typo in `interactive_markeres`.  Maybe this was never being used.

src/perception/CMakeLists.txt
psigen: Also, I have no idea why the `interactive_marker_client.h` path suffix is used here for TF.

jslee02: > Also, I have no idea why the interactive_marker_client.h path suffix is used here for TF.

I believe that was a mistake. [I fixed it in `Findtf.cmake`](https://github.com/personalrobotics/aikido/pull/198/files#diff-b9abb4ef14835558cc1d5aabdf1035fbR20).

> Do we need to put these path hints in the Findtf.cmake file? They are definitely good defaults.

In `Findtf.cmake`, I used `tf.pc` that includes better hint, which is generally correct.

src/rviz/CMakeLists.txt
jslee02: > Note: Typo in interactive_markeres. Maybe this was never being used.

Which part is the typo exactly? There are many `interactive_markeres`. :sweat:

psigen: `interactive_markeres` -> `interactive_markers`, right?

Doesn't matter since this PR removes it, but suggests that the path never actually got used.

jslee02: > interactive_markeres -> interactive_markers, right?

Oh, right. Thanks! interactive_mark**e**res! Yeah, I also believe the path never got used, but leave there just in case with fixed one.

psigen:
I am pretty sure the only reason that ROS CMake files don't use this is because the mechanism only recently made it to the CMake repos on 14.04 and wouldn't have been supported when many of them were last touched.

I think we should use it here if these are effectively private CMake modules.  I don't think these `find_XXXX` directives will leak out through the `find_package(aikido)` call, right?

jslee02:
> I see that we use pkg_check_modules to find the dependencies, but then use find_path and find_library to set the x_INCLUDE_DIRS and x_LIBRARIES variables instead of using the output from pkg-config directly. Is this necessary because the output of pkg-config includes the offending gtest invocations? If so, can you add a comment to that effect?

That's actually for the case `pkg-config` failed to find the `.pc` file so that CMake does the second try with `find_path` and `find_library`. I believe `pkg-config` has nothing to do the offecding gtest invocations.

jslee02:
> I was considering creating an IMPORTed target for the dependency libraries and using target_include_directories and target_link_libraries to specify PUBLIC (or INTERFACE) dependencies. This would have the advantage of correctly percolating transitive dependencies through Aikido.

I like this idea even though I haven't used that mechanism yet. Let me create an issue for that.



> I don't think these find_XXXX directives will leak out through the find_package(aikido) call, right?

I believe so.

PR #200, opened 2017-04-28
Remove tag requirements for vector/matrix in YAML

### LINKED ISSUE #188: Consider removing tag requirements for vector and matrix in the yaml parser
In a yaml file, we currently need to specify tags for vector and matrix (i.e., `!Vector` and `!Matrix`) type in advance like:
```yaml
left:
  default: !Matrix &left_default
    [[0., 0., 1., 0.],
     [1., 0., 0., 0.],
     [0., 1., 0., .18],
     [0., 0., 0., 1.]]
  cylinder: *left_default
```
We could remove the tag requirement by checking if the node is sequence (for vector) or nested sequence (for matrix).

Related PR: #175

jslee02:
This PR resolves #188 by removing the custom tag requirements for vector and matrix type in YAML. This fix is inspired by the original parser [`perception/eigen_yaml.hpp`](https://github.com/personalrobotics/aikido/blob/4c08ac570005d74e5622f81f68c7f1bde32a382f/include/aikido/perception/eigen_yaml.hpp) that already resolved the issue. :disappointed_relieved: I thought it was identical to [`kenv`'s yaml parser](https://github.com/personalrobotics/kenv/blob/master/box2d_kenv/include/box2d_kenv/yaml_utils.h) but was not true.


include/aikido/util/detail/yaml_extension.hpp
```diff
@@ -135,13 +131,13 @@ struct convert<Eigen::
       throw YAML::RepresentationException(getMark(node), ss.str());
     }
 
-    if (node.Tag() == "!Vector")
+    if (node[0].Type() == YAML::NodeType::Scalar)
     {
       matrix.resize(rows, 1);
       for (Index i = 0; i < rows; ++i)
         matrix(i, 0) = node[i].template as<_Scalar>();
```
psigen: Do we want to check that _each_ row of the sequence is also a `Scalar`.  Otherwise, if we try to load a sequence like `[2, [2,3], 4, 5]` we will potentially get an unhelpful error message like `YAML::BadConversion`.  I am not sure if this is the case or not.

tests/util/test_yaml_extension.cpp
```diff
@@ -131,3 +131,57 @@ TEST(YamlEigenExtension, TsrTransforms)
...
+
+  Eigen::Isometry3d expectedTag124 = Eigen::Isometry3d::Identity();
+  expectedTag124.linear()
+      << -0.0068393994632501, -0.09054787493933   , -0.99586861832219  ,
+         -0.9999766064057   ,  0.00052349074982687,  0.0068200145734176,
+         -9.6209816943679e-5,  0.99589196617977   , -0.090549337061432 ;
+  expectedTag124.translation()
+      << 0.14460904118031, -0.00097599685803373, -0.033644917605727;
+  EXPECT_TRUE(tag124.matrix().isApprox(expectedTag124.matrix()));
+}
```
psigen: Should we also test that we get representation exceptions with invalid YAML?  Like mismatched sizes?

include/aikido/util/detail/yaml_extension.hpp
jslee02: :+1: I believe that case is covered by [this](https://github.com/personalrobotics/aikido/pull/200/files#diff-ee420700678d3f6c35f1b8428db1bd8eR163).

Another corner case would be when `node` doesn't have a child, which I'm not sure if we care about the case as well.

tests/util/test_yaml_extension.cpp
```diff
@@ -4,29 +4,62 @@
...
+  std::string notMatrix1 = "matrix: [[1, 2, 3], [1, 2, 3], 1]";
+  EXPECT_THROW(YAML::Load(notMatrix1)["matrix"].as<Eigen::Matrix3d>(),
       YAML::RepresentationException);
 
-  std::string incorrectSize = "vector: [1, 2, 3]";
-  EXPECT_THROW(YAML::Load(incorrectSize)["vector"].as<Eigen::Vector2d>(),
+  std::string notMatrix2 = "matrix: [[1, 2, 3], [1, 2, 3], [1, 2]]";
+  EXPECT_THROW(YAML::Load(notMatrix2)["matrix"].as<Eigen::Matrix3d>(),
       YAML::RepresentationException);
 }
```
mkoval: Nit: Please split this into a few separate tests, e.g. one for "incorrect size", another for "not sequence", etc.

My general preference would be to have one test per type of logical error. That way: (1) it is clear which case is not being handled correctly if a test fails and (2) it's easy to tell what a particular block of test code is trying to test.

tests/util/test_yaml_extension.cpp
```diff
@@ -131,3 +164,57 @@ TEST(YamlEigenExtension, TsrTransforms)
...
+        "      ,                                                       \n"
+        "      [                                                       \n"
+        "        0,                                                    \n"
+        "        0,                                                    \n"
+        "        0,                                                    \n"
+        "        1                                                     \n"
+        "      ]                                                       \n"
+        "    ]                                                         \n"
+        "  }                                                           \n"
+        "}                                                             \n";
```
mkoval: Nit: This line wrapping is terrible! :rage4: It is also odd that we're using a real configuration file here - that includes a crazy hard-coded transformation and references a file that will not exist on the build server (i.e. is inside the `pr_ordata` package) in an Aikido unit test.

We never try to load the file, so it's not technically wrong, but it does smell funny. 😬

tests/util/test_yaml_extension.cpp
```diff
@@ -4,29 +4,62 @@
...
+  EXPECT_THROW(YAML::Load(incorrectSize1)["vector"].as<Eigen::Vector2d>(),
+      YAML::RepresentationException);
+
+  std::string incorrectSize2 = "matrix: [[1, 2], [3, 4], [5, 6]]";
+  EXPECT_THROW(YAML::Load(incorrectSize2)["matrix"].as<Eigen::Matrix3d>(),
+      YAML::RepresentationException);
+
+  std::string notMatrix1 = "matrix: [[1, 2, 3], [1, 2, 3], 1]";
+  EXPECT_THROW(YAML::Load(notMatrix1)["matrix"].as<Eigen::Matrix3d>(),
       YAML::RepresentationException);
```
mkoval: Add the converse of this test, which starts with a scalar and has a `Sequence` as a later element. This is an interesting test case because we use `node[0].Type()` to decide whether we're deserializing a matrix or a vector.

include/aikido/util/detail/yaml_extension.hpp
jslee02: > Yes, we have to handle the case where node doesn't have a child. I believe the current implementation will SEGFAULT because we access node[0].Type() even when node.size() == 0.

I confirmed that this case is handled by [this](https://github.com/personalrobotics/aikido/pull/200/files#diff-ee420700678d3f6c35f1b8428db1bd8eR118) through a [test](https://github.com/personalrobotics/aikido/pull/200/files#diff-87b50c6224fda02940d3d364f0606c50R10). If the node doesn't have a child, then the node type is `Null`.

> I also agree agree with @psigen - we should have an additional check that node[i].Type() == YAML::NodeType::Scalar before we try the conversion node[i].template as<_Scalar>().
>
> @jslee02 I saw your link, but don't understand why the tests are relevant here. :thinking:

I thought he meant incorrect column size for each row. The suggested additional checks are added for both of vector and matrix cases.

tests/util/test_yaml_extension.cpp
jslee02: Done.

tests/util/test_yaml_extension.cpp
jslee02: I wanted to test with a real use case, but it seems not necessary. Updated.

tests/util/test_yaml_extension.cpp
jslee02: Done.

PR #206, opened 2017-05-02
Refactor Parabolic timer and add Parabolic smoother

dqyi11:
(1) Refactor Parabolic timer into 
(1.a) convert an interpolated into a spline
(1.b) convert a spline into a timed spline
(2) Add Parabolic smoother 
(2.a) convert a spline to a dynamic path
(2.b) convert a dynamic path to a spine
(2.c) add HauserParabolicSmoother class that handles "doShortcut", "doBlend" and "doShortcutAndBlend"


src/planner/parabolic/ParabolicUtil.hpp
```diff
@@ -0,0 +1,72 @@
...
+
+/// Convert a spline trajectory to a dynamic path
+/// \param _inputTrajectory a spline trajectory
+/// \param[out] _startTime the start time of the spline trajectory
+/// \param _maxVelocity maximum velocity for each dimension
+/// \param _maxAcceleration maximum acceleration for each dimension
+/// \return a dynamic path
+std::unique_ptr<ParabolicRamp::DynamicPath>
+    convertToDynamicPath(const aikido::trajectory::Spline* _inputTrajectory,
+                         double& _startTime,
```
mkoval: What is the purpose of this? :confused:

src/trajectory/Interpolated.cpp
```diff
@@ -100,6 +100,18 @@ void Interpolated::evaluateDerivative(double _t, int _derivative,
     _tangentVector.setZero();
   }
 
+  //if(_t==getStartTime())
```
mkoval: Nit: Remove dead code.

src/trajectory/Interpolated.cpp
```diff
@@ -100,6 +100,18 @@ void Interpolated::evaluateDerivative(double _t, int _derivative,
...
+  if(_t==mWaypoints[0].t)
```
mkoval: This is comparing two floating point numbers for exact equality, which is almost always a bug. What is the purpose of thsi check?

src/trajectory/Interpolated.cpp
```diff
@@ -100,6 +100,18 @@ void Interpolated::evaluateDerivative(double _t, int _derivative,
...
+  {
+      const auto segmentTime = mWaypoints[1].t - mWaypoints[0].t;
+      const auto alpha = (mWaypoints[1].t - mWaypoints[0].t) / segmentTime;
```
mkoval: It looks like `alpha` will always be `1` because the numerator and denominator are both `mWaypoints[1].t - mWaypoints[0].t`. What am I missing/

src/trajectory/Interpolated.cpp
```diff
@@ -100,6 +100,18 @@ void Interpolated::evaluateDerivative(double _t, int _derivative,
     _tangentVector.setZero();
   }
 
+  //if(_t==getStartTime())
+  if(_t==mWaypoints[0].t)
+  {
+      const auto segmentTime = mWaypoints[1].t - mWaypoints[0].t;
```
mkoval: This will SEGFAULT if `mWaypoints.size() < 2`.

src/trajectory/Interpolated.cpp
```diff
@@ -100,6 +100,18 @@ void Interpolated::evaluateDerivative(double _t, int _derivative,
...
+      const auto alpha = (mWaypoints[1].t - mWaypoints[0].t) / segmentTime;
+
+      mInterpolator->getDerivative(mWaypoints[0].state, mWaypoints[1].state,
+              _derivative, alpha, _tangentVector);
+      _tangentVector /= segmentTime;
+      return;
```
mkoval: Why do we explicitly `return` here? :confused:

src/trajectory/Interpolated.cpp
```diff
@@ -152,6 +164,16 @@ const statespace::StateSpace::State* Interpolated::getWaypoint(
 }
 
 //=============================================================================
+double Interpolated::getWaypointTime(size_t _index) const
+{
+  if (_index < mWaypoints.size())
+    //return getStartTime() + mWaypoints[_index].t;
+    return mWaypoints[_index].t;
```
mkoval: Won't this break if `getStartTime() != 0`?

src/trajectory/Interpolated.cpp
```diff
@@ -152,6 +164,16 @@ const statespace::StateSpace::State* Interpolated::getWaypoint(
 }
 
 //=============================================================================
+double Interpolated::getWaypointTime(size_t _index) const
+{
+  if (_index < mWaypoints.size())
+    //return getStartTime() + mWaypoints[_index].t;
```
mkoval: If the current implementation is correct, remove this dead code.

src/trajectory/Spline.cpp
```diff
@@ -198,5 +198,48 @@ Eigen::VectorXd Spline::evaluatePolynomial(
...
+  }
+  else
+    throw std::domain_error("Waypoint index is out of bounds.");
+}
+
+//=============================================================================
+void Spline::getWaypointDerivative(size_t _index, int _derivative,
+                                   Eigen::VectorXd& _tangentVector) const
+{
+  if (_index < getNumSegments())
```
mkoval: This should be `getNumWapoints()`, not `getNumSegements()`.

tests/planner/parabolic/test_ParabolicSmoother.cpp
```diff
@@ -0,0 +1,240 @@
...
+    mStraightLine->addWaypoint(0., state);
+
+    state.setValue(Vector2d(3., 4.));
+    mStraightLine->addWaypoint(1., state);
+
+  }
+
+  std::shared_ptr<R2> mStateSpace;
+  Eigen::Vector2d mMaxVelocity;
+  Eigen::Vector2d mMaxAcceleration;
```
mkoval: These are [fixed-size vectorizable types](https://eigen.tuxfamily.org/dox/group__TopicFixedSizeVectorizable.html), so you need to [annotate this class with `EIGEN_MAKE_ALIGNED_OPERATOR_NEW`](https://eigen.tuxfamily.org/dox/group__TopicStructHavingEigenMembers.html). Otherwise you will see runtime crashes on some architectures, namely `x86`.

tests/planner/parabolic/test_ParabolicSmoother.cpp
```diff
@@ -0,0 +1,240 @@
...
+
+TEST_F(ParabolicSmootherTests, convertInterpolatedToSpline)
+{
+    auto spline = convertToSpline(*mStraightLine.get());
+
+    auto splineState = mStateSpace->createState();
+    auto interpolatedState = mStateSpace->createState();
+    Eigen::VectorXd splineVec, interpolatedVec;
+    Eigen::VectorXd splineTangent, interpolatedTangent;
+    for(double t=spline->getStartTime();t<=spline->getEndTime();t+=1e-2)
```
mkoval: Use `StepSequence` to implement this stepping logic.

Also make `1e-2` a constant and decrease the value - perhaps to `1e-3`.

tests/planner/parabolic/test_ParabolicSmoother.cpp
```diff
@@ -0,0 +1,240 @@
...
+    {
+        spline->evaluate(t, splineState);
+        mStraightLine->evaluate(t, interpolatedState);
+        mStateSpace->logMap(splineState, splineVec);
+        mStateSpace->logMap(interpolatedState, interpolatedVec);
+        EXPECT_TRUE(interpolatedVec.isApprox(interpolatedVec));
```
mkoval: Nit: Use `EXPECT_EIGEN_EQUAL` here instead of `EXPECT_TRUE`. It will generate better error messages.

tests/planner/parabolic/test_ParabolicSmoother.cpp
```diff
@@ -0,0 +1,240 @@
...
+
+        spline->evaluateDerivative(t, 1, splineTangent);
+        mStraightLine->evaluateDerivative(t, 1, interpolatedTangent);
+        EXPECT_TRUE(splineTangent.isApprox(interpolatedTangent));
```
mkoval: Nit: Also use `EXPECT_EIGEN_EQUAL` here.

tests/planner/parabolic/test_ParabolicSmoother.cpp
```diff
@@ -0,0 +1,240 @@
...
+    }
+}
+
+TEST_F(ParabolicSmootherTests, doShort)
```
mkoval: Typo: `doShortcut`.

tests/planner/parabolic/test_ParabolicSmoother.cpp
```diff
@@ -0,0 +1,240 @@
...
+   state.setValue(p1);
+   nonStraightLine.addWaypoint(0., state);
+   state.setValue(p2);
+   nonStraightLine.addWaypoint(0.25, state);
+   state.setValue(p3);
+   nonStraightLine.addWaypoint(0.5, state);
+   state.setValue(p4);
+   nonStraightLine.addWaypoint(0.75, state);
+   state.setValue(p5);
+   nonStraightLine.addWaypoint(1., state);
```
mkoval: Nit: Could we implement this test using only three waypoints?

tests/planner/parabolic/test_ParabolicSmoother.cpp
```diff
@@ -0,0 +1,240 @@
...
+   }
+
+   // Position.
+   smoothedTrajectory->evaluate(smoothedTrajectory->getStartTime(), state);
+   EXPECT_TRUE(p1.isApprox(state.getValue()));
+
+   smoothedTrajectory->evaluate(smoothedTrajectory->getEndTime(), state);
+   EXPECT_TRUE(p5.isApprox(state.getValue()));
+
+   EXPECT_TRUE(shortenDist <= orignDist);
```
mkoval: Nit: Make this a strict inequality and increase the time limit until it succeeds with very high probability.

tests/planner/parabolic/test_ParabolicSmoother.cpp
```diff
@@ -0,0 +1,240 @@
...
+   smoothedTrajectory->evaluate(smoothedTrajectory->getEndTime(), state);
+   Eigen::VectorXd vec;
+   mStateSpace->logMap(state, vec);
+   EXPECT_TRUE(p6.isApprox(state.getValue()));
+
+   //EXPECT_TRUE(shortenDist <= orignDist);
+
+   double originTime = nonStraightLine.getDuration();
+   double shortenTime = smoothedTrajectory->getDuration();
+   EXPECT_TRUE(shortenTime<=originTime);
```
mkoval: Make this a strict inequality.

tests/planner/parabolic/test_ParabolicSmoother.cpp
```diff
@@ -0,0 +1,240 @@
...
+   EXPECT_TRUE(p1.isApprox(state.getValue()));
+
+   smoothedTrajectory->evaluate(smoothedTrajectory->getEndTime(), state);
+   EXPECT_TRUE(p5.isApprox(state.getValue()));
+
+   EXPECT_TRUE(shortenDist <= orignDist);
+
+   double originTime = nonStraightLine.getDuration();
+   double shortenTime = smoothedTrajectory->getDuration();
+   EXPECT_TRUE(shortenTime<=originTime);
```
mkoval: See my comment above. Make this a strict inequality.

Also: There's a lot of duplicate code here. Can we move some of it to a helper function?

src/planner/parabolic/ParabolicUtil.cpp
```diff
@@ -0,0 +1,228 @@
...
+     _inputTrajectory->getWaypoint(iwaypoint, currentState);
+
+     stateSpace->logMap(currentState, tangentVector);
+     milestones.emplace_back(toVector(tangentVector));
+  }
+
+  auto outputPath = make_unique<ParabolicRamp::DynamicPath>();
+  // Apply the adjoint limits
+  outputPath->Init(toVector(_maxVelocity), toVector(_maxAcceleration));
+  outputPath->SetMilestones(milestones);
```
mkoval: We need another version of this helper function - or an overload of this function - that optionally passes velocities to `SetMilestones`. We need that for all of the functions that accept non-piecewise linear splines.

src/planner/parabolic/ParabolicSmoother.cpp
```diff
@@ -0,0 +1,150 @@
...
+    bool _useVelocity)
+{
+  auto stateSpace = _inputTrajectory->getStateSpace();
+  HauserParabolicSmoother parabolicSmoother(_feasibilityCheck,
+                                            _timelimit,
+                                            _useVelocity);
+
+  double startTime = 0.0;
+  auto dynamicPath = convertToDynamicPath(_inputTrajectory, startTime,
+                                          _maxVelocity, _maxAcceleration);
```
mkoval: This call to `convertToDynamicPath` should preserve velocities.

src/planner/parabolic/ParabolicSmoother.cpp
```diff
@@ -0,0 +1,150 @@
...
+{
+  auto stateSpace = _inputTrajectory->getStateSpace();
+  HauserParabolicSmoother parabolicSmoother(_feasibilityCheck,
+                                            _timelimit,
+                                            _useVelocity,
+                                            _blendRadius,
+                                            _blendIterations);
+  double startTime = 0.0;
+  auto dynamicPath = convertToDynamicPath(_inputTrajectory, startTime,
+                                          _maxVelocity, _maxAcceleration);
```
mkoval: This call to `convertToDynamicPath` should preserve velocities.

src/planner/parabolic/ParabolicSmoother.cpp
```diff
@@ -0,0 +1,150 @@
...
+  auto stateSpace = _inputTrajectory->getStateSpace();
+  HauserParabolicSmoother parabolicSmoother(_feasibilityCheck,
+                                            _timelimit,
+                                            _useVelocity,
+                                            _blendRadius,
+                                            _blendIterations);
+
+  double startTime = 0.0;
+  auto dynamicPath = convertToDynamicPath(_inputTrajectory, startTime,
+                                          _maxVelocity, _maxAcceleration);
```
mkoval: This call to `convertToDynamicPath` should preserve velocities.

src/planner/parabolic/ParabolicUtil.hpp
dqyi11: @mkoval A spline trajectory has a start time (not necessarily starting from 0). However, ParabolicRamp::DynamicPath always assumes starting from 0.
I did not modify the DynamicPath class. So I used a start time variable to save the information when converting.
When I convert the DynamicPath back to a Spline, I will use the start time to recover.

src/trajectory/Interpolated.cpp
dqyi11: @mkoval  Could we assume that mWaypoints.size() is always >= 2 in an interpolated?

src/trajectory/Interpolated.cpp
dqyi11: @mkoval I am taking _t == start_time as a specific case. So I will return here to skip running the other part of the function. This is an unfinished discussion. Do you still think we should support the derivative here?

src/trajectory/Interpolated.cpp
dqyi11: In the Interpolated class, getStartTime() is taken as the time of the first way point.
"mWaypoints.front().t"
I will double check the correctness.

tests/planner/parabolic/test_ParabolicSmoother.cpp
dqyi11: No.... In Hauser's code, shortcut will be executed only when the waypoint number is larger than 3....

include/aikido/planner/parabolic/ParabolicTimer.hpp
```diff
@@ -34,6 +34,37 @@ std::unique_ptr<trajectory::Spline> computeParabolicTiming(
...
+/// \param _inputTrajectory input piecewise Geodesic trajectory
+/// \param _maxVelocity maximum velocity for each dimension
+/// \param _maxAcceleration maximum acceleration for each dimension
+/// \return time optimal trajectory that satisfies acceleration constraints
+std::unique_ptr<aikido::trajectory::Spline> computeParabolicTiming(
+      const aikido::trajectory::Spline* _inputTrajectory,
+      const Eigen::VectorXd& _maxVelocity,
+      const Eigen::VectorXd& _maxAcceleration);
+
+/// Convert an interpolated trajectory to a spline trajectory
```
dqyi11: @mkoval  The velocities are used in conversion. Do you think it is ignored?

src/planner/parabolic/ParabolicTimer.cpp
```diff
@@ -154,97 +105,26 @@ std::unique_ptr<trajectory::Spline> computeParabolicTiming(
...
-    assert(spline.getCoefficients().size() == 1);
-    const auto& coefficients = spline.getCoefficients().front();
-    outputTrajectory->addSegment(
-        coefficients, timeCurr - timePrev, segmentStartState);
-
-    timePrev = timeCurr;
-    positionPrev = positionCurr;
-    velocityPrev = velocityCurr;
-  }
+  double startTime = 0.0;
```
dqyi11: Because in converting to a dynamic path, the start time information is lost.
I do it this way to remind a user that he needs to save the start time if he wants to convert the dynamic path back to a spline.

src/planner/parabolic/ParabolicUtil.cpp
```diff
@@ -0,0 +1,228 @@
...
+  {
+     auto currentState = stateSpace->createState();
+     _inputTrajectory->getWaypoint(iwaypoint, currentState);
+
+     stateSpace->logMap(currentState, tangentVector);
+     milestones.emplace_back(toVector(tangentVector));
+  }
+
+  auto outputPath = make_unique<ParabolicRamp::DynamicPath>();
+  // Apply the adjoint limits
```
dqyi11: Yes, I am going to add velocities.

dqyi11:
@mkoval  All the comments are addressed.
The only concern is whether we should support the derivative at the start time of an interpolated trajectory, which defines the modifications in the interploated class.
I think we need the velocity to convert an interpolated trajectory to a spline trajectory, because a spline trajectory assumes the smoothness.
Please let me know what you think about it. Thanks!

src/planner/parabolic/HauserParabolicSmootherHelpers.cpp
```diff
@@ -0,0 +1,198 @@
...
+        if (std::fabs(rampNd.dx1[idof]) > ParabolicRamp::EpsilonV)
+        {
+            return false;
+        }
+    }
+    return true;
+}
+
+bool tryBlend(ParabolicRamp::DynamicPath& dynamicPath,
+              ParabolicRamp::RampFeasibilityChecker feasibilityChecker,
```
dqyi11: Only changed it to ParabolicRamp::RampFeasibilityChecker&. 
Because it calls ```dynamicPath.TryShortcut(t1, t2, feasibilityChecker);``` , which is ``` bool TryShortcut(Real t1,Real t2,RampFeasibilityChecker& check); ```

include/aikido/planner/parabolic/ParabolicSmoother.hpp
```diff
@@ -0,0 +1,124 @@
...
+
+/// Shortcut waypoints in a trajectory using parabolic splines.
+///
+/// This function smooths `_inputTrajectory' by iteratively sampling
+/// two waypoints in a trajectory and trying to find a shortcut using
+/// the following algorithm:
+///
+/// while _timelimit is not out:
+/// - Randomly sample two times uniformly at random.
+/// - Find two waypoints according two times in _inputTrajectory.
```
dqyi11: @mkoval   Should I then say "points" instead of "waypoints"?

dqyi11:
I have addressed all the comments.

The only thing left is whether we shall keep "useVelocity" in converting an interpolated to a spline.
This can be useful in the ParabolicSmoother.

In that, I convert a spline trajectory to a dynamic path without providing velocity.
Then dynamic path will re-evaluate the velocity.
If velocities are given, dynamic path will not re-evaluate the velocities.
Thus no new timing information will be added.
Does this make sense?

include/aikido/planner/parabolic/ParabolicTimer.hpp
```diff
@@ -30,17 +30,18 @@ namespace parabolic {
...
-      const Eigen::VectorXd& _maxVelocity,
-      const Eigen::VectorXd& _maxAcceleration);
+    const aikido::trajectory::Spline& _inputTrajectory,
+    const Eigen::VectorXd& _maxVelocity,
+    const Eigen::VectorXd& _maxAcceleration);
 
 /// Convert an interpolated trajectory to a spline trajectory
-/// This function requires the \c _inputTrajectory to use a \c GeodesicInterpolator.
+/// This function requires the \c _inputTrajectory to use a \c
+/// GeodesicInterpolator.
```
jslee02: Nit: Could you fix the wrapping? ClangFormat isn't good at it. :disappointed:

src/planner/parabolic/HauserParabolicSmootherHelpers.cpp
```diff
@@ -13,172 +13,184 @@ namespace planner {
...
+  SmootherFeasibilityCheckerBase(
+      aikido::constraint::TestablePtr testable, double checkResolution)
+    : testable_(std::move(testable))
+    , checkResolution_(checkResolution)
+    , statespace_(testable_->getStateSpace())
+    , interpolator_(statespace_)
+  {
+  }
+
+  virtual bool ConfigFeasible(ParabolicRamp::Vector const& x)
```
jslee02: Nit: Please use `const X&` pattern rather than `X const&`.

src/planner/parabolic/HauserParabolicSmootherHelpers.cpp
```diff
@@ -13,172 +13,184 @@ namespace planner {
...
-          statespace_(testable_->getStateSpace()),
-          interpolator_(statespace_)
+  SmootherFeasibilityCheckerBase(
+      aikido::constraint::TestablePtr testable, double checkResolution)
+    : testable_(std::move(testable))
+    , checkResolution_(checkResolution)
+    , statespace_(testable_->getStateSpace())
+    , interpolator_(statespace_)
+  {
+  }
```
jslee02: Nit: Please add `// Do nothing` comment to indicate that this function is intentionally empty (but initializing the member variables).

src/planner/parabolic/HauserParabolicSmootherHelpers.cpp
```diff
@@ -13,172 +13,184 @@ namespace planner {
...
+  virtual bool ConfigFeasible(ParabolicRamp::Vector const& x)
+  {
+    Eigen::VectorXd eigX = toEigen(x);
+    auto state = statespace_->createState();
+    statespace_->expMap(eigX, state);
+    return testable_->isSatisfied(state);
+  }
+
+  virtual bool SegmentFeasible(
+      ParabolicRamp::Vector const& a, ParabolicRamp::Vector const& b)
```
jslee02: Nit: Ditto

src/planner/parabolic/HauserParabolicSmootherHelpers.cpp
```diff
@@ -13,172 +13,184 @@ namespace planner {
...
-    }
+    return true;
+  }
 
 private:
-    aikido::constraint::TestablePtr          testable_;
-    double                                   checkResolution_;
-    aikido::statespace::StateSpacePtr        statespace_;
-    aikido::statespace::GeodesicInterpolator interpolator_;
+  aikido::constraint::TestablePtr testable_;
```
jslee02: Nit: Could you rename these private member variables using the pattern of `mX`? I believe we don't use the trailing underscore for member variables (even for private member variables).

src/planner/parabolic/HauserParabolicSmootherHelpers.cpp
```diff
@@ -13,172 +13,184 @@ namespace planner {
...
+  double checkResolution_;
+  aikido::statespace::StateSpacePtr statespace_;
+  aikido::statespace::GeodesicInterpolator interpolator_;
 };
 
-bool needsBlend(ParabolicRamp::ParabolicRampND const &rampNd)
+bool needsBlend(ParabolicRamp::ParabolicRampND const& rampNd)
```
jslee02: Nit: `const ParabolicRamp::ParabolicRampND&`

src/planner/parabolic/HauserParabolicSmootherHelpers.cpp
```diff
@@ -13,172 +13,184 @@ namespace planner {
...
-    // that waypoint indicies can change between iterations of the algorithm.
-    size_t const numRamps = dynamicPath.ramps.size();
-    double const tMax = dynamicPath.GetTotalTime();
-    double t = 0;
-
-    for (size_t iRamp = 0; iRamp < numRamps-1; ++iRamp)
+  // blending can completely remove waypoints from the trajectory in the case
+  // that two waypoints are closer than _blendRadius together - which means
+  // that waypoint indicies can change between iterations of the algorithm.
+  size_t const numRamps = dynamicPath.ramps.size();
```
jslee02: Nit: `const size_t`.

src/planner/parabolic/HauserParabolicSmootherHelpers.cpp
```diff
@@ -13,172 +13,184 @@ namespace planner {
...
+  double const tMax = dynamicPath.GetTotalTime();
```
jslee02: Nit: `const double`

src/planner/parabolic/HauserParabolicSmootherHelpers.cpp
```diff
@@ -13,172 +13,184 @@ namespace planner {
...
-            bool const success = dynamicPath.TryShortcut(t1, t2, feasibilityChecker);
-
-            if (success)
-            {
-                return true;
-            } else {
-                rampNd.blendAttempts++;
-            }
-        }
+      double const t1 = std::max(t - dtShortcut, -ParabolicRamp::EpsilonT);
```
jslee02: Nit: `const double`

src/planner/parabolic/HauserParabolicSmootherHelpers.cpp
```diff
@@ -13,172 +13,184 @@ namespace planner {
...
+      double const t2
```
jslee02: Nit: `const double`

src/planner/parabolic/HauserParabolicSmootherHelpers.cpp
```diff
@@ -13,172 +13,184 @@ namespace planner {
...
+          = std::min(t + dtShortcut, tMax + ParabolicRamp::EpsilonT);
+
+      bool const success = dynamicPath.TryShortcut(t1, t2, feasibilityChecker);
```
jslee02: Nit: `const bool`

src/planner/parabolic/HauserParabolicSmootherHelpers.cpp
```diff
@@ -13,172 +13,184 @@ namespace planner {
...
+
+  double dtShortcut = blendRadius;
+
+  // tryBlend always starts at the beginning of the trajectory.
+  // Without this bookkeeping, tryBlend would could any blend that fails
+  // multiple times.
+  for (int attempt = 0; attempt < blendIterations; ++attempt)
+  {
+    while (tryBlend(dynamicPath, feasibilityChecker, attempt, dtShortcut))
+      ;
```
jslee02: :scream: What's happening here?

src/planner/parabolic/HauserParabolicSmootherHelpers.cpp
jslee02: Also, please begin with a lower letter for member functions.

src/planner/parabolic/HauserParabolicSmootherHelpers.cpp
jslee02: Please rename to `segmentFeasible(...)`.

src/planner/parabolic/HauserParabolicSmootherHelpers.cpp
jslee02: Nit: At first, I thought this was a mistake, but now I got it. I would like to suggest make this code easier to read something like:
```cpp
bool success;
do 
{
  success = tryBlend(...);
} while (success);
```
Also, I'm not sure if which one is logically correct between `while(success)` and `while(!success)`. :confused:

include/aikido/planner/parabolic/ParabolicTimer.hpp
jslee02: Nit: I thought "So the conversion ..." was the continuing sentence without line breaking.

include/aikido/planner/parabolic/ParabolicSmoother.hpp
```diff
@@ -0,0 +1,137 @@
...
+/// from the parabolic ramp along any axis, and then checks for
+/// configuration and segment feasibility along that piecewise linear path.
+/// \_rngSeed seed used by a random generator for sampling time in shortcut.
+/// \return smoothed trajectory that satisfies acceleration constraints
+std::unique_ptr<trajectory::Spline> doShortcut(
+    const trajectory::Spline& _inputTrajectory,
+    aikido::constraint::TestablePtr _feasibilityCheck,
+    const Eigen::VectorXd& _maxVelocity,
+    const Eigen::VectorXd& _maxAcceleration,
+    aikido::util::RNG& _rngSeed,
```
mkoval: Nit: This *is* the RNG, not a seed for the RNG. 😉

src/planner/parabolic/HauserParabolicSmootherHelpers.cpp
```diff
@@ -0,0 +1,198 @@
...
+    Eigen::VectorXd eigA = toEigen(a);
+    Eigen::VectorXd eigB = toEigen(b);
+
+    auto testState = mStateSpace->createState();
+    auto startState = mStateSpace->createState();
+    auto goalState = mStateSpace->createState();
+    mStateSpace->expMap(eigA, startState);
+    mStateSpace->expMap(eigB, goalState);
+
+    aikido::util::VanDerCorput vdc{1, false, false, mCheckResolution};
```
mkoval: Nit: Add a comment explaining why `includeStartpoint` and `includeEndpoint` can be `false`.

src/planner/parabolic/HauserParabolicSmootherHelpers.cpp
```diff
@@ -0,0 +1,198 @@
...
+bool doShortcut(
+    ParabolicRamp::DynamicPath& dynamicPath,
+    aikido::constraint::TestablePtr testable,
+    double timelimit,
+    double checkResolution,
+    double tolerance,
+    aikido::util::RNG& rng)
+{
+  if (timelimit < 0.0)
+    throw std::runtime_error("Timelimit should be non-negative");
```
mkoval: Nits:
- Add a bounds check for `checkResolution`.
- Use `std::invalid_argument` instead of `std::runtime_error` for these exceptions.

src/planner/parabolic/HauserParabolicSmootherHelpers.cpp
```diff
@@ -0,0 +1,198 @@
...
+  if (tolerance < 0.0)
+    throw std::runtime_error("Tolerance should be non-negative");
+
+  SmootherFeasibilityCheckerBase base(testable, checkResolution);
+  ParabolicRamp::RampFeasibilityChecker feasibilityChecker(&base, tolerance);
+
+  std::chrono::time_point<std::chrono::system_clock> startTime, currentTime;
```
mkoval: Nits:
- `currentTime` is only used inside the loop, so move its definition to where it is first used.
- Initialize `startTime` on the same line where the variable it is defined.

src/planner/parabolic/HauserParabolicSmootherHelpers.cpp
```diff
@@ -0,0 +1,198 @@
...
+    double t1 = dist(rng);
+    double t2 = dist(rng);
+    dynamicPath.TryShortcut(t1, t2, feasibilityChecker);
+
+    currentTime = std::chrono::system_clock::now();
+    elapsedTime = std::chrono::duration_cast<std::chrono::duration<double>>(
+                      currentTime - startTime)
+                      .count();
+  }
+  return true;
```
mkoval: This `bool` seems to be meaningless. We should check whether `TryShortcut` ever reported success, i.e. returned `true`, and report that here.

src/planner/parabolic/HauserParabolicSmootherHelpers.cpp
```diff
@@ -0,0 +1,198 @@
...
+    aikido::constraint::TestablePtr testable,
+    double blendRadius,
+    int blendIterations,
+    double checkResolution,
+    double tolerance)
+{
+  if (blendIterations <= 0)
+    throw std::runtime_error("Blend iterations should be positive");
+  if (blendRadius <= 0.0)
+    throw std::runtime_error("Blend radius should be positive");
```
mkoval: Nits:
- Add bounds checks for `checkResolution` and `tolerance`.

src/planner/parabolic/HauserParabolicSmootherHelpers.cpp
mkoval: The current logic is correct. The core issue is that blending can completely remove waypoints from the trajectory, which makes bookkeeping via the time parameter extraordinarily difficult. We dodge that issue by allowing `tryBlend` to always start at the beginning of the trajectory. We continue to call `tryBlend` in a loop until it finds nothing left to blend.

src/planner/parabolic/ParabolicTimer.cpp
```diff
@@ -135,6 +39,50 @@ std::unique_ptr<trajectory::Spline> computeParabolicTiming(
...
+    const auto currentState = _inputTrajectory.getWaypoint(iwaypoint);
+    double currentTime = _inputTrajectory.getWaypointTime(iwaypoint);
+    const auto nextState = _inputTrajectory.getWaypoint(iwaypoint + 1);
+    double nextTime = _inputTrajectory.getWaypointTime(iwaypoint + 1);
+
+    stateSpace->logMap(currentState, currentVec);
+    stateSpace->logMap(nextState, nextVec);
+
+    _inputTrajectory.evaluateDerivative(currentTime, 1, currentVelocity);
+    _inputTrajectory.evaluateDerivative(nextTime, 1, nextVelocity);
```
mkoval: This is not meaningful.

It is not possible to convert an `Interpolated` to a `Spline` in the general case - it is only possible when the `Interpolator` used by the input evaluates a polynomial (i.e. is a spline itself). In practice, the only case we need to handle is the `GeodesicInterpolator`, which is equivalent to a piecewise linear spline.

A piecewise linear spline is not C1 continuous, ,so `evaluateDerivative` is undefined at waypoints. Since that is the case, I suggest:
1. verifying that `_inputTrajectory->getInterpolator()` is a `GeodesicInterpolator`
2. creating a `LinearSplineProblem` instead of a `CubicSplineProblem`
3. removing the derivative `addConstantConstraint(_, 1, _)` calls.

src/planner/parabolic/ParabolicTimer.cpp
```diff
@@ -154,94 +102,12 @@ std::unique_ptr<trajectory::Spline> computeParabolicTiming(
...
-    outputTrajectory->addSegment(
-        coefficients, timeCurr - timePrev, segmentStartState);
-
-    timePrev = timeCurr;
-    positionPrev = positionCurr;
-    velocityPrev = velocityCurr;
-  }
+  double startTime = _inputTrajectory.getStartTime();
+  auto dynamicPath = convertToDynamicPath(
+      _inputTrajectory, _maxVelocity, _maxAcceleration, false);
```
mkoval: This version should pass `_useVelocity = true`!

We can convert two types of inputs to a `DynamicPath`:

1. Untimed piecewise linear `Interpolated` trajectories that are untimed and are converted to timed `Spline` trajectories that stop at each waypoint [which is necessary to follow a piecewise linear trajectory under acceleration bounds]. In this case, `_useVelocity = false`.
2. Timed parabolic `Spline` trajectories that we want to shortcut or blend. In this case, it is important that we preserve the exact geometry of the input path, so `it is important that `_useVelocity = true`. Otherwise, `DyanmicRamp` will treat the `Spline`'s knot points as waypoints connected by piecewise linear segments.

src/planner/parabolic/ParabolicUtil.cpp
```diff
@@ -0,0 +1,230 @@
...
+using aikido::statespace::StateSpace;
+using dart::common::make_unique;
+
+using CubicSplineProblem
+    = aikido::util::SplineProblem<double, int, 4, Eigen::Dynamic, 2>;
+
+namespace aikido {
+namespace planner {
+namespace parabolic {
+
```
mkoval: Nit: These helper functions belong in a `detail` namespace.

src/planner/parabolic/ParabolicUtil.cpp
```diff
@@ -0,0 +1,230 @@
...
+  for (size_t iwaypoint = 0; iwaypoint < numWaypoints; ++iwaypoint)
+  {
+    auto currentState = stateSpace->createState();
+    _inputTrajectory.getWaypoint(iwaypoint, currentState);
+
+    stateSpace->logMap(currentState, currVec);
+    milestones.emplace_back(toVector(currVec));
+
+    _inputTrajectory.getWaypointDerivative(iwaypoint, 1, tangentVector);
+    velocities.emplace_back(toVector(tangentVector));
```
mkoval: Nit: Don't call `getWaypointDerivative` if `_useVelocity` is `false`.

src/planner/parabolic/ParabolicUtil.hpp
```diff
@@ -0,0 +1,70 @@
...
+
+/// Convert a spline trajectory to a dynamic path
+/// \param _inputTrajectory a spline trajectory
+/// \param[out] _startTime the start time of the spline trajectory
+/// \return a dynamic path
+std::unique_ptr<ParabolicRamp::DynamicPath>
+    convertToDynamicPath(const aikido::trajectory::Spline& _inputTrajectory,
+                         const Eigen::VectorXd& _maxVelocity,
+                         const Eigen::VectorXd& _maxAcceleration,
+                         bool _useVelocity = true);
```
mkoval: @dqyi11 I would still like to address this.

src/trajectory/Interpolated.cpp
```diff
@@ -100,6 +100,17 @@ void Interpolated::evaluateDerivative(double _t, int _derivative,
     _tangentVector.setZero();
   }
 
+  if(_t==mWaypoints[0].t)
```
mkoval: We're testing two floating point numbers of equality here. This is usually a bug, because the numbers may differ due to numerical precision issues. Please either add a tolerance or add a comment about why exact equality is appropriate here (e.g. because `t` was copied out of `mWaypoints[0].t`).

src/trajectory/Interpolated.cpp
mkoval: @dqyi11 Please address this.

src/trajectory/Spline.cpp
```diff
@@ -198,5 +198,48 @@ Eigen::VectorXd Spline::evaluatePolynomial(
...
 
+//=============================================================================
+size_t Spline::getNumWaypoints() const
+{
+    return getNumSegments() + 1;
+}
+
+//=============================================================================
+double Spline::getWaypointTime(size_t _index) const
+{
```
mkoval: This is resolved.

tests/planner/parabolic/test_ParabolicSmoother.cpp
```diff
@@ -0,0 +1,231 @@
...
+using aikido::planner::parabolic::doShortcutAndBlend;
+
+class ParabolicSmootherTests : public ::testing::Test
+{
+public:
+    EIGEN_MAKE_ALIGNED_OPERATOR_NEW
+protected:
+  void SetUp() override
+  {
+    mRng = aikido::util::RNGWrapper<std::mt19937>( std::random_device{}() );
```
mkoval: Please use a deterministic seed (e.g. `0`) in unit tests so their result is not stochastic.

tests/planner/parabolic/test_ParabolicSmoother.cpp
```diff
@@ -0,0 +1,231 @@
...
+
+    auto state = mStateSpace->createState();
+
+    state.setValue(Vector2d(1., 2.));
+    mStraightLine->addWaypoint(0., state);
+
+    state.setValue(Vector2d(3., 4.));
+    mStraightLine->addWaypoint(1., state);
+
+    mNonStraightLineLength = initNonStraightLine();
```
mkoval: It's odd to have one output (`mNonStraightLine`) in a member variable and a second `mNonStraightLineLength` as a return value. Please stick to one convention or the other for this function.

tests/planner/parabolic/test_ParabolicSmoother.cpp
```diff
@@ -0,0 +1,231 @@
...
+    return length;
+  }
+
+  double getLength(aikido::trajectory::Spline* spline)
+  {
+    double length = 0.0;
+    auto currState = mStateSpace->createState();
+    auto nextState = mStateSpace->createState();
+    Eigen::VectorXd currVec, nextVec;
+    for(size_t i=0; i<spline->getNumWaypoints()-1;++i)
```
mkoval: Nit: This will overflow if `getNumWaypoints() == 0`, which could cause the test to SEGFAULT instead of fail properly. Please refactor the loop so that is not an issue, e.g. by starting with `i = 1` and replacing `i` with `i - 1` in the body.

tests/planner/parabolic/test_ParabolicSmoother.cpp
```diff
@@ -0,0 +1,231 @@
...
+  std::shared_ptr<Interpolated> mNonStraightLine;
+
+  double mNonStraightLineLength;
+  double mTimelimit = 60.0;
+  const double mTolerance = 1e-5;
+};
+
+TEST_F(ParabolicSmootherTests, convertInterpolatedToSpline)
+{
+    auto spline = convertToSpline(*mStraightLine.get());
```
mkoval: Nit: `*mStraightLine.get()` is equivalent to `*mStraightLine`.

tests/planner/parabolic/test_ParabolicSmoother.cpp
```diff
@@ -0,0 +1,231 @@
...
+        mStraightLine->evaluate(t, interpolatedState);
+        mStateSpace->logMap(splineState, splineVec);
+        mStateSpace->logMap(interpolatedState, interpolatedVec);
+        EXPECT_EIGEN_EQUAL(splineVec, interpolatedVec, mTolerance);
+
+        spline->evaluateDerivative(t, 1, splineTangent);
+        mStraightLine->evaluateDerivative(t, 1, interpolatedTangent);
+        EXPECT_EIGEN_EQUAL(splineTangent, interpolatedTangent, mTolerance);
+    }
+}
```
mkoval: Add tests for:
- More than one segment, e.g. the non-striaghtline trajectory.
- A trajectory with a non-zero start time.

src/planner/parabolic/HauserParabolicSmootherHelpers.cpp
jslee02: That makes sense. It would be nice to add this as a comment at least to `tryBlend()`. 

I still prefer to change the code to be self-documenting. For example:
```cpp
bool needMoreBlending;
do 
{
  needMoreBlending = tryBlend(...);
} while (needMoreBlending);
```

src/trajectory/Interpolated.cpp
dqyi11: I'd prefer to add a comment here. Because it is only when _t is exactly the same with mWaypoints[0].t , we do something special. Also, @mkoval do you think we shall not support evaluate Derivative when _t < mWaypoints[0].t?

src/planner/parabolic/ParabolicTimer.cpp
dqyi11: @mkoval By what you suggested, here we should pass `_useVelocity = false`, right? Because we are converting an untimed piecewise linear `Spline` to a `DynamicPath` (after making changes by above suggestion).

src/trajectory/Interpolated.cpp
mkoval: > Because it is only when _t is exactly the same with mWaypoints[0].t , we do something special.

I still don't really understand why we're checking for exact equality. What if, due to numerical precision, we end up getting `mWaypoints[0].t - eps`? :confused: 

> Also, @mkoval do you think we shall not support evaluate Derivative when _t < mWaypoints[0].t?

I think we should support this. We had a long discussion about this last year and decided that we would not implement hard bounds checks in Aikido arrays, unless the underlying implementation absolutely requires it. The issue is that numerical precision can result in an algorithm querying the result slightly outside of the nominal bounds.

This numerical imprecision is often unavoidable. It's often more useful to return a reasonable value, e.g. by extrapolating the last segment or extending the trajectory with a constant segment than to report an error. @psigen and I got burned by this repeatedly when implementing `or_parabolicsmoother` - as you can see by the variety of `epsilon`s. :sweat_smile:

src/planner/parabolic/ParabolicTimer.cpp
mkoval: Yes, that is correct. :+1: 

Although, as per our discussion on Slack, we could  also remove this function entirely. We only really need (1) to support untimed `Interpolated` inputs and (2) to support timed `Spline` inputs.

src/trajectory/Interpolated.cpp
dqyi11: @mkoval When _t == mWaypoints[0].t, `int idx == getWaypointIndexAfterTime(_t)` returns 0 to idx. Then a domain error will be thrown. I needed the velocity of the start time of an `Interploated` to convert it to a parabolic cubic `Spline`. Since you suggested to convert it into a piecewise linear `Spline`, I think I do not need the velocities any more. 

Since we are not currently supporting the time before the start time now. As you pointed out previously that it is not continuous here thus no derivative shall be supported. Maybe I could just revert all the changes made here. Does this make sense?

mkoval: I am not 100% sure that I follow. For (1), the input `Interpolated` is piecewise linear so we do not need any velocities [as you noted]. If that is consistent with your understanding, then I agree that we should revert those changes.

dqyi11: OK. I will revert it.

src/planner/parabolic/ParabolicTimer.cpp
dqyi11: As discussed with @mkoval , in `ParabolicTimer', there should be 
(1) `computeParabolicTime()` that converts an untimed piecewise linear `Interplotaed` to a timed parabolic cubic `Spline`; and
(2) `convertToSpline` that converts an untimed piecewise linear `Interploated` to a untimed piecewise linear `Spline` (Not used now).

src/planner/parabolic/HauserParabolicSmootherHelpers.cpp
```diff
@@ -0,0 +1,216 @@
...
+  for (unsigned int i = 0; i < dynamicPath.ramps.size(); ++i)
+  {
+    dynamicPath.ramps[i].blendAttempts = 0;
+  }
+
+  double dtShortcut = blendRadius;
+
+  // tryBlend always starts at the beginning of the trajectory.
+  // Without this bookkeeping, tryBlend would could any blend that fails
+  // multiple times.
```
mkoval: Nit: Typo "would could." I think this should say "tryBlend would retry any blend that fails multiple times."

src/planner/parabolic/HauserParabolicSmootherHelpers.cpp
```diff
@@ -0,0 +1,216 @@
...
+  // We continue to call tryBlend() in a loop until it finds nothing left
+  // to blend.
+  for (int attempt = 0; attempt < blendIterations; ++attempt)
+  {
+    bool noMoreBlending = true;
+    do
+    {
+      noMoreBlending
+          = tryBlend(dynamicPath, feasibilityChecker, attempt, dtShortcut);
+    } while (noMoreBlending);
```
mkoval: Nit: The name `noMoreBlending` implies the opposite of what it actually is. I suggest renaming the flag to `needsMoreBlending`.

src/planner/parabolic/ParabolicUtil.cpp
```diff
@@ -0,0 +1,228 @@
...
+    return true;
+  }
+  else if (dynamic_cast<const R<5>*>(_stateSpace) != nullptr)
+  {
+    return true;
+  }
+  else if (dynamic_cast<const R<6>*>(_stateSpace) != nullptr)
+  {
+    return true;
+  }
```
mkoval: Resolved.

src/planner/parabolic/ParabolicUtil.cpp
mkoval: Resolved by splitting the `Spline` and `Interpolated` cases into separate overloads. 👍

Issue #117, opened 2016-11-22
Consider switching to Codecov for online coverage reporting


jslee02:
The graphical user interface of [Codecov](https://codecov.io/gh/dartsim/dart/branch/codecov) looks more organized and intuitive than [Coveralls](https://coveralls.io/github/dartsim/dart?branch=master).

Related issue: [DART #806](https://github.com/dartsim/dart/pull/806)

mkoval:
👍 Codecov has a much nicer UI. I've always thought that Coveralls looks a bit like the `Coveralls` project page looks a bit like a ransom note. 😁

@psigen Thoughts?

psigen:
Generally fine with this.  But let's try it out before we commit to it completely.

We should be able to add `codecov` support in a branch without removing the `coveralls` support, right?  I think they have non-overlapping configuration.  We can try that and see how it looks?

jslee02:
I've created a PR for this. Please take a look at it when you have a chance.

PR #208, opened 2017-05-05
Switch to codecov for online code coverage reporting

jslee02:
* Report comparison: [coveralls (before)](https://coveralls.io/github/personalrobotics/aikido?branch=master) vs [codecov (after)](https://codecov.io/gh/personalrobotics/aikido/branch/code_quality%2Fcodecov)
* [Project setup reference](https://github.com/codecov/example-cpp11-cmake)
* Related issue: #117

One more nice thing about codecov over coveralls is that codecov makes only one comment on a pull request and updates if necessary rather than posting new comments, which was like spammy. :sweat_smile:


jslee02:
Merging. We could revisit if something needs to be changed is found.

PR #214, opened 2017-06-01
Set default `_numOutputs` to 1 in `splitEngine`

### LINKED ISSUE #187: Add splitEngine overload for _numOutputs = 1
We often use `splitEngine` with `_numOutputs = 1`, which requires a disturbing amount of boilerplate. It would be nice to add an overload to handle this common case.

brianhou:
Resolves #187.

@mkoval is this what you meant? My initial thought was to have the two methods:
- `splitEngine(RNG& _engine, size_t _numOutputs, size_t _numSeeds = NUM_DEFAULT_SEEDS)` (already exists)
- `splitEngine(RNG& _engine, size_t _numSeeds = NUM_DEFAULT_SEEDS)` (call the former with `_numOutputs = 1`)

but this makes `splitEngine(RNG&, size_t)` ambiguous.

### DIFF
--- a/include/aikido/util/RNG.hpp
+++ b/include/aikido/util/RNG.hpp
@@ -137,7 +137,7 @@ Quaternion sampleQuaternion(
 /// \param _numSeeds number of seeds to use for initialization
 /// \return new random number generators
 std::vector<std::unique_ptr<util::RNG>> splitEngine(
-    RNG& _engine, size_t _numOutputs, size_t _numSeeds = NUM_DEFAULT_SEEDS);
+    RNG& _engine, size_t _numOutputs = 1, size_t _numSeeds = NUM_DEFAULT_SEEDS);
 
 } // namespace util
 } // namespace aikido
--- a/include/aikido/util/RNG.hpp
+++ b/include/aikido/util/RNG.hpp
@@ -11,7 +11,7 @@
 namespace aikido {
 namespace util {
 
-/// Default number of seeds to by \c splitEngine to seed new engines.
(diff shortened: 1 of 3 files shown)


jslee02:
According to the description of #187, it seems @brianhou 's initial thought is what @mkoval means. To avoid the function name's ambiguity, we could use different name. If the intention is to create _one object_ with the same type of another existing object, I would like to call the function `clone` (or `cloneFrom`).

Also, we could use `cloneRNGs` for multiple cloning version instead of `splitEngine`.

I'm open to using different name from what I proposed, but my point is to use different function name to avoid the ambiguity.

Or, we could remove all the default parameters of the original function.

brianhou:
Different function names sound reasonable, although I think it'd make more sense to keep the names similar. How about `cloneRngFrom` (for the new one-output version) and `cloneRngsFrom` (for the current `splitEngine`)?

I'd prefer to keep the default parameters; I don't want to arbitrarily choose a value for `_numSeeds` each time I use it :sweat_smile:

jslee02:
`cloneRngFrom` and `cloneRngsFrom` overall sound good to me, but have a minor concern that they use different capitalization (`RNG` vs `Rng`). I personally prefer `Rng` for any other acronyms, but it seems the current convention is `RNG`. So we may want to stay with it like `cloneRNGFrom` and `cloneRNGsFrom` unless we change the convention and apply it to all other names.

Also, out of curiosity, do we have a specific reason why not we have this functions as member functions?

PR #216, opened 2017-06-27
Fixes to work on real hardware

brianhou:
Issues:

- [x] `RosJointStateClient::jointStateCallback` was incorrectly handling empty buffers. Messages were being ignored because the timestamp was being read from the last record in an empty buffer.
- [x] For some reason, `RosJointStateClient::spin` doesn't seem to be waiting long enough for callbacks to be placed on its queue. I've temporarily added a duration of 1 second here.
- [x] The future returned by `RosTrajectoryExecutor::execute` never seems to be fulfilled. Calling `future::wait` waits forever (even after the trajectory seems to be successfully executed). I have a temporary hack in `pantry_loading` that sleeps for the duration of the trajectory instead of calling `future::wait`, but that needs to be fixed.

  It seems that `RosTrajectoryExecutor::transitionCallback` is sometimes not being called at all. On other trajectories, it _is_ being called and `isSuccessful` is true (the resulting message is `"Execution failed"`, which seems like a [typo](https://github.com/personalrobotics/aikido/blob/427f7be13b5b65a87316a2e7c7109dd8a0b07a37/src/control/ros/RosTrajectoryExecutor.cpp#L178)?) and the promise's value seems to be set. In both cases, the future still doesn't seem to be fulfilled. I thought that calling `future::wait` would return once a value is set by `promise::set_value`, but I could be misunderstanding.

   When `RosTrajectoryExecutor::transitionCallback` is being called, it also seems that it is called with `handle.getCommState() == DONE` well before the trajectory actually terminates (basically immediately). Is this expected behavior?
- [x] ~Calling `future::wait` on the future returned by `RosPositionCommandExecutor` results in the error message: `Exception thrown by callback: Promise already satisfied`.~

  This was because I forgot to start `right_hand_controller` 😓 Now the communication state is stuck at `ACTIVE` forever...

### DIFF
--- a/include/aikido/control/ros/RosTrajectoryExecutor.hpp
+++ b/include/aikido/control/ros/RosTrajectoryExecutor.hpp
@@ -40,11 +40,11 @@ public:
   virtual ~RosTrajectoryExecutor();
 
   /// Sends trajectory to ROS server for execution.
-  /// \param[in] traj Trajecotry to be executed.
+  /// \param[in] traj Trajectory to be executed.
   std::future<void> execute(trajectory::TrajectoryPtr traj) override;
 
   /// Sends trajectory to ROS server for execution.
-  /// \param[in] traj Trajectrory to be executed.
+  /// \param[in] traj Trajectory to be executed.
   /// \param[in] startTime Start time for the trajectory.
   std::future<void> execute(
       trajectory::TrajectoryPtr traj, const ::ros::Time& startTime);
--- a/src/control/ros/RosJointStateClient.cpp
+++ b/src/control/ros/RosJointStateClient.cpp
@@ -94,7 +94,7 @@ void RosJointStateClient::jointStateCallback(
         boost::circular_buffer<JointStateRecord>{mCapacity});
     auto& buffer = result.first->second;
 
-    if (_jointState.header.stamp < buffer.back().mStamp)
+    if (!buffer.empty() && _jointState.header.stamp < buffer.back().mStamp)
     {
       // Ignore out of order JointState message.
       ROS_WARN_STREAM(
--- a/src/control/ros/RosJointStateClient.cpp
+++ b/src/control/ros/RosJointStateClient.cpp
@@ -33,7 +33,7 @@ void RosJointStateClient::spin()
   std::lock_guard<std::mutex> skeletonLock{mSkeleton->getMutex()};
   std::lock_guard<std::mutex> bufferLock{mMutex};
 
-  mCallbackQueue.callAvailable();
+  mCallbackQueue.callAvailable(::ros::WallDuration(1.0));
 }
 
 //=============================================================================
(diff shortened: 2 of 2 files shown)


mkoval:
>  For some reason, `RosJointStateClient::spin` doesn't seem to be waiting long enough for callbacks to be placed on its queue. I've temporarily added a duration of 1 second here.

You are intended to call `spin()` regularly, e.g. at a fixed frequency in a background thread. There is no reason to think that calling `spin()` once after executing a trajectory will work correctly. Adding a timeout to `callAvailable()` is entirely the wrong solution. 😉

>  The future returned by `RosTrajectoryExecutor::execute` never seems to be fulfilled. Calling `future::wait` waits forever (even after the trajectory seems to be successfully executed).
> [...]
> It seems that `RosTrajectoryExecutor::transitionCallback` is sometimes not being called at all. 

The root issue here is the same as above: there is no background thread periodically calling `spin()`. Or, if there is a thread, the thread hung/crashed and is no longer running.

> On other trajectories, it is being called and `isSuccessful` is true (the resulting message is "Execution failed", which seems like a typo?) and the promise's value seems to be set.

The logic in `transitionCallback` is actually correct. An `FollowJointTrajectory` can be flagged as a failure in two different ways:

1. at the `actionlib` level because the goal was recalled, rejected, preempted, aborted, or lost (indicated by `GoalHandle::getTerminalState()`)
2. at the `FollowJointTrajectoryResult` level because of a logical error, e.g. an unknown joint was specified or the robot deviated too far from the path (indicated by `FollowJointTrajectoryResult::error_code`.

We only want to report success if `getTerminalState()` is `SUCCEEDED` _and_ `error_code` is `SUCCESSFUL`. All of the rest of the code in `transitionCallback` exists to munge all of the error flags (`getTerminalState()`, `error_code`, and `error_string`)  into one human-readable message.

> I thought that calling `future::wait` would return once a value is set by `promise::set_value`, but I could be misunderstanding.

That is correct. We are using `std::future`, which is part of the standard library, so I would be surprised if the error resides here. Are you sure `set_result` or `set_exception` is being called on `promise` associated with the `future` that you expect to return?

It is very possible there are bugs in `RosTrajectoryExecutor`, but I would be surprised if they are this fundamental. I successfully used the class that `RosTrajectoryExecutor` was based on successfully on Valkyrie without any of these issues.

brianhou:
Yep, that was definitely it. :sweat_smile: Fixed in https://github.com/personalrobotics/pantry-loading/pull/12/commits/3738173c75af9dca135264b7765b7a697e84e870.

Also, it seems that the hand issues stem from a lower level -- trying to just use `libbarrett_ros` without `aikido` doesn't successfully move the hand to a preshape either. I'm currently thinking there might be an issue in https://github.com/personalrobotics/pr_ros_controllers/tree/master/position_command_controller or something? Still need to look into this more.

> The logic in `transitionCallback` is actually correct. An `FollowJointTrajectory` can be flagged as a failure in two different ways...

I think I mostly understand this method, although https://github.com/personalrobotics/aikido/blob/427f7be13b5b65a87316a2e7c7109dd8a0b07a37/src/control/ros/RosTrajectoryExecutor.cpp#L178 still confuses me a bit. It seems that the message is set to `Execution failed` if `terminalState == TerminalState::SUCCEEDED` without considering the `FollowJointTrajectoryResult`. Would it make more sense to remove lines 176-179 and add the `<< "Execution failed."` to line 186?

brianhou:
I think these changes are uncontroversial, so I'm merging this now.

PR #217, opened 2017-07-17
Support only specifying some joints in the space.

brianhou:
Partially resolves https://github.com/personalrobotics/rewd_controllers/issues/13.


src/control/ros/Conversions.cpp
```diff
@@ -223,27 +239,44 @@ std::vector<const dart::dynamics::Joint*> findJointByName(
...
+    throw std::invalid_argument{message.str()};
+  }
+  else if (startPositions.size() > 0 &&
+           static_cast<std::size_t>(startPositions.size()) != numControlledJoints)
+  {
+    std::stringstream message;
+    message << "Incorrect number of joints in configuration: expected "
+            << numControlledJoints << ", got " << startPositions.size() << ".";
     throw std::invalid_argument{message.str()};
   }
```
jslee02: This logic works, but the two cases have overlapping cases that lead to redundant checks. I would like to simplify this as:
```cpp
  if (numTrajectoryJoints > numControlledJoints)
  {
    // throw exception
  }
  else if (numTrajectoryJoints < numControlledJoints)
  {
    if (static_cast<std::size_t>(startPositions.size()) != numControlledJoints)
    {
      // throw exception
    }
  }
  // When numTrajectoryJoints == numControlledJoints, don't care the size of
  // startPositions
```

src/control/ros/Conversions.cpp
```diff
@@ -223,27 +239,44 @@ std::vector<const dart::dynamics::Joint*> findJointByName(
...
 
   // Check that the names in jointTrajectory are unique.
   std::vector<std::string> joint_names;
   joint_names.reserve(numControlledJoints);
-  for (size_t i = 0; i < numControlledJoints; ++i)
+  for (size_t i = 0; i < numTrajectoryJoints; ++i)
     joint_names.emplace_back(jointTrajectory.joint_names[i]);
```
jslee02: Unrelated nitpick: I think this vector copy can be simplified using the copy constructor as
```cpp
  std::vector<std::string> joint_names(
      jointTrajectory.joint_names.begin(), jointTrajectory.joint_names.end());
```

src/control/ros/Conversions.cpp
```diff
@@ -281,11 +314,11 @@ std::unique_ptr<SplineTrajectory> toSplineJointTrajectory(
 
   // Map joint indices between jointTrajectory and space subspaces.
   std::map<size_t, size_t> rosJointToMetaSkeletonJoint;
```
jslee02: Unrelated nitpick: I think we could use `std::vector<std::pair<size_t, size_t>>` here instead of `std::map<...>` because we actually don't take the advantage of `std::map` as `std::map::find(...)` is not used. On the contrary, inserting items to `std::map` is more expensive than using `std::vector`.

src/control/ros/Conversions.cpp
```diff
@@ -281,11 +314,11 @@ std::unique_ptr<SplineTrajectory> toSplineJointTrajectory(
...
+  std::set<size_t> specifiedMetaSkeletonJoints;
```
jslee02: Nitpick: `std::unorderd_set` would be better since we don't worry about the order here.

src/control/ros/Conversions.cpp
```diff
@@ -311,6 +344,23 @@ std::unique_ptr<SplineTrajectory> toSplineJointTrajectory(
 
     rosJointToMetaSkeletonJoint.emplace(
         std::make_pair(trajIndex, metaSkeletonIndex));
+    specifiedMetaSkeletonJoints.insert(metaSkeletonIndex);
+  }
+
+  // Add unspecified joint mappings to rosJointToMetaSkeletonJoint
+  std::vector<size_t> unspecifiedMetaSkeletonJoints;
```
jslee02: Nitpick: We could reserve the size of `unspecifiedMetaSkeletonJoints` to `numControlledJoints - numTrajectoryJoints`.

src/control/ros/Conversions.cpp
```diff
@@ -311,6 +344,23 @@ std::unique_ptr<SplineTrajectory> toSplineJointTrajectory(
...
+  if (static_cast<std::size_t>(startPositions.size()) == numControlledJoints && numTrajectoryJoints < numControlledJoints)
```
jslee02: Nitpick: We can save this check by storing a flag whether we're in padding mode or not. The mode can be determined in the above error checking.

jslee02:
Also, shouldn't this PR target to `master` branch instead of `test/brianhou/rightarm` branch?

brianhou:
Thanks for helping with the changes. Looks good, but I'll double-check by testing things over the weekend before merging.

brianhou:
We can't test this until the left arm is fixed, so I'm just going to merge this.

Issue #95, opened 2016-04-29
Inconsistent naming: Sampleable/SO2Sampleable/SO2UniformSampler


psigen:
Currently, the sampling interface is called `Sampleable`.
We define a header called `SO2UniformSampler`.
Inside it is a class called `SO2Sampleable`.

We need to standardize on these things.  I posit the following naming convention:
- **virtual** and **abstract** classes adopt the _adjective_ name: `Sampleable`
- **concrete** classes adopt the _noun_ name: `Sampler`

This makes some sense when doing inheritance, as the interface describe what the class will do, but if a class does multiple things it may not be named after any of them, e.g.

```
SO2Sampler : Sampleable
TSR : Sampleable, Projectable
```

psigen:
This is related to #91 but more drastic 😓

jslee02:
Resolved by #224

Issue #91, opened 2016-04-26
Inconsistent naming: SO2Sampler vs SO3UniformSampler


mkoval:
We should decide on one of the two.

jslee02:
Resolved by #224

PR #224, opened 2017-09-29
Rename SO2Sampleable to SO2UniformSampler

jslee02:
addresses #91 and #95


include/aikido/constraint/uniform/SO2UniformSampler.hpp
```diff
@@ -31,6 +32,9 @@ class SO2Sampleable : public constraint::Sampleable
   std::unique_ptr<util::RNG> mRng;
 };
 
+using SO2Sampleable = SO2UniformSampler;
```
brianhou: I think we should either add a deprecation warning or just not have this old name at all. Since I don't think there's anything using this yet, I vote for the latter. :smiling_imp:

Issue #181, opened 2017-04-19
Consider splitting utils into common and io


jslee02:
The current `utils` component seems to play two roles: 
(1) common utilities used by most of other Aikido components
(2) file input-output utilities

I believe we want to keep the common component as light as possible regarding required external dependencies. For this, I would like to suggest splitting it into `common` and `io` components as
(1) `common`: a default (build always) component that requires only the essential dependencies (such as Eigen, DART).
(2) `io`: an optional component that probably requires additional dependencies related to file input-output such as `tinyxml2` and `yaml-cpp`.

mkoval:
I like this idea. 👍

jslee02:
Resolved by #225

PR #225, opened 2017-09-29
Split util namespace into common and io namespaces

jslee02:
addresses #181


brianhou:
Can you add a PR to `libherb` that uses the updated namespaces? Thanks!

jslee02:
> Can you add a PR to libherb that uses the updated namespaces?

:+1: personalrobotics/libherb#48

Issue #220, opened 2017-09-27
StepSequnce test fails on 32-bit Ubuntu


jslee02:
Here is the test error message:
```
/<<PKGBUILDDIR>>/tests/util/test_StepSequence.cpp:94: Failure
Value of: seq2.getMaxSteps()
  Actual: 5
Expected: 6
[  FAILED  ] StepSequence.MaxSteps (0 ms)
[ RUN      ] StepSequence.MaxStepsAlternateRanges
/<<PKGBUILDDIR>>/tests/util/test_StepSequence.cpp:109: Failure
Value of: seq2.getMaxSteps()
  Actual: 20
Expected: 21
```

I found this test failure working on uploading Aikido to the PRL PPA ([build log](https://launchpadlibrarian.net/338903511/buildlog_ubuntu-trusty-i386.aikido0_0.0.2.ppa1~trusty_BUILDING.txt.gz)).

yskim041:
aikido 0.1.0 got same error messages and one more from `test_RosTrajectoryConversions`.

```
[ RUN      ] StepSequence.MaxSteps
/<<PKGBUILDDIR>>/tests/common/test_StepSequence.cpp:94: Failure
Value of: seq2.getMaxSteps()
  Actual: 5
Expected: 6
[  FAILED  ] StepSequence.MaxSteps (0 ms)
[ RUN      ] StepSequence.MaxStepsAlternateRanges
/<<PKGBUILDDIR>>/tests/common/test_StepSequence.cpp:109: Failure
Value of: seq2.getMaxSteps()
  Actual: 20
Expected: 21
[  FAILED  ] StepSequence.MaxStepsAlternateRanges (0 ms)
...
[ RUN      ] ToRosJointTrajectoryTests.TrajectoryHasCorrectWaypoints
/<<PKGBUILDDIR>>/tests/control/ros/test_RosTrajectoryConversions.cpp:145: Failure
Value of: rosTrajectory2.points[i].time_from_start.toSec()
  Actual: 0.029999999000000003
Expected: timestep*i
Which is: 0.029999999999999999
[  FAILED  ] ToRosJointTrajectoryTests.TrajectoryHasCorrectWaypoints (0 ms)
```

brianhou:
I think @dqyi11 fixed these tests in in #305, but let's keep an eye out next time that we package AIKIDO.

jslee02:
Can we reopen this issue and close once we're sure if it's fixed in the next packaging?

[...]

PR #228, opened 2017-10-03
Packaging 0.1.0

jslee02:
Next time, let's update the changelog per PR. :sob:


jslee02:
@yskim041 Could you check if the debian package of this PR is uploaded without any errors? It would be good to release 0.1.0 after we confirmed that.

yskim041:
It was published only for amd64 without any errors. There were three errors for i386, and I added a comment about this on #220.

jslee02:
Thanks @yskim041 ! Merging this PR.

PR #229, opened 2017-10-03
pr template

yskim041:
I made `PULL_REQUEST_TEMPLATE.md` and `ISSUE_TEMPLATE.md` in `.github`. After adding these files in master, every PR will start with these templates.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,3 +1,7 @@
+[Insert your comment here]
+
+***
 Please make sure that these boxes are checked before closing your pull request.
 - [ ] Set milestone for this PR
-- [ ] Add a brief explanation about this PR in CHANGELOG.md
+- [ ] Added a brief explanation about this change in CHANGELOG.md
+- [ ] Added unit test for this change
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,4 @@
+Please make sure that these boxes are checked before closing your pull request. Thank you!
+- [ ] Set milestone for this PR
+- [ ] Add a brief explanation about this PR in CHANGELOG.md
+
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,4 +1,3 @@
-Please make sure that these boxes are checked before closing your pull request. Thank you!
+Please make sure that these boxes are checked before closing your pull request.
 - [ ] Set milestone for this PR
 - [ ] Add a brief explanation about this PR in CHANGELOG.md
-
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,3 @@
+Please make sure that these boxes are checked before closing your pull request.
```
jslee02: It would be nice to have a separate bar above this checklists so that the post comment can be easily distinguished.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,3 @@
...
+- [ ] Set milestone for this PR
+- [ ] Add a brief explanation about this PR in CHANGELOG.md
```
jslee02: I would like to add another item something like "Added unit test for this change".

jslee02: It more makes sense to me if we use past tense for the leading verb since checking box means we confirmed it's done.

yskim041: Please make sure that these boxes are checked before closing your pull request.
- [ ] Set milestone for this PR
- [ ] Added a brief explanation about this change in CHANGELOG.md
- [ ] Added unit test for this change
***
Here is an example. It is not possible to put new lines at the beginning of a template, so this layout is easier to make so far.

brianhou: I think imperatives make more sense, since we're telling people to do these things.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,7 +1,14 @@
...
\ No newline at end of file
+
+**Before creating a pull request**
+
+- [ ] Document new methods and classes
+- [ ] Format code with `make format`
+
+**Before merging a pull request**
+
+- [ ] Set version target
```
jslee02: It would be good to mention how to set the version target. It would be something like "Set version target by selecting milestone on the right side"

brianhou: Should this basically just be deciding between whether this PR goes into the next minor or major release? If so, we should add that as well.

If it's supposed to wait until a major release, are we planning to keep some PRs open until the next major release is ready?

jslee02: No. In that case, we might have a branch for it. Actually, if we're planning to release a major version and minor version (I don't believe this will happen soon though), `master` would be for the major version release, `release-0.3` (0.3 for example) would be for the minor version release. We just merge the PR into the right branch.

Targeting version for a PR just can be done by selecting proper branch. One of the reasons of using milestone is to collect PRs and issues in each milestone pages per version to see what changes have been made in the release like [this](https://github.com/personalrobotics/aikido/milestone/4?closed=1).

jslee02:
@yskim041 Please merge this PR if this looks good to you. I don't think we need to wait until the Travis test is done.

yskim041:
Looks good to me. Thanks!

PR #235, opened 2017-10-09
[WIP] return Colliding Bodies' Names if user given state is in collision

aditya-vk:
This PR forces the planning procedure to terminate if the user provided start state is in collision and prints the body nodes that are in collision.

WIP: Needs tests.
***

### DIFF
--- a/include/aikido/constraint/Testable.hpp
+++ b/include/aikido/constraint/Testable.hpp
@@ -14,12 +14,9 @@ public:
   virtual ~Testable() = default;
 
   /// Returns true if state satisfies this constraint.
+  /// \param _givenState: true if state is user-provided
   virtual bool isSatisfied(
       const statespace::StateSpace::State* _state, bool _givenState = false) const = 0;
-
-  // /// Returns true if state satisfies this constraint.
-  // virtual bool isSatisfiedGivenState(
-  //     const statespace::StateSpace::State* _state) const = 0;
 
   /// Returns StateSpace in which this constraint operates.
   virtual statespace::StateSpacePtr getStateSpace() const = 0;
--- a/include/aikido/constraint/Testable.hpp
+++ b/include/aikido/constraint/Testable.hpp
@@ -15,7 +15,11 @@ public:
 
   /// Returns true if state satisfies this constraint.
   virtual bool isSatisfied(
-      const statespace::StateSpace::State* _state) const = 0;
+      const statespace::StateSpace::State* _state, bool _givenState = false) const = 0;
+
+  // /// Returns true if state satisfies this constraint.
+  // virtual bool isSatisfiedGivenState(
+  //     const statespace::StateSpace::State* _state) const = 0;
 
   /// Returns StateSpace in which this constraint operates.
   virtual statespace::StateSpacePtr getStateSpace() const = 0;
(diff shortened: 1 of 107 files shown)


aditya-vk:
Great suggestions! 
Yeah, this is quite messy but was a quick hack-around to unblock some of the planning issues we were facing with `pantry-loading` demo (planner returns an `invalid start state` due to fingers colliding with each other when the hand is closed but we had no access to this information previously). Created a PR so this issue can be tracked and completed appropriately soon. I should've probably left it as WIP.

include/aikido/planner/ompl/detail/Planner-impl.hpp
```diff
@@ -24,6 +24,11 @@ trajectory::InterpolatedPtr planOMPL(
     double _maxPlanTime,
     double _maxDistanceBtwValidityChecks)
 {
+
```
jslee02: Nit: Please remove unnecessary this whitespace

include/aikido/planner/ompl/detail/Planner-impl.hpp
```diff
@@ -73,6 +78,10 @@ trajectory::InterpolatedPtr planOMPL(
...
```
jslee02: Nit: Please remove unnecessary this whitespace

src/planner/SnapPlanner.cpp
```diff
@@ -17,6 +17,10 @@ trajectory::InterpolatedPtr planSnap(
     const std::shared_ptr<aikido::constraint::Testable>& constraint,
     aikido::planner::PlanningResult& planningResult)
 {
+
```
jslee02: Nit: Please remove unnecessary this whitespace

src/planner/ompl/Planner.cpp
```diff
@@ -218,6 +218,10 @@ trajectory::InterpolatedPtr planCRRT(
     double _maxDistanceBtwProjections,
     double _minStepsize)
 {
+
```
jslee02: Nit: Please remove unnecessary this whitespace

src/planner/ompl/Planner.cpp
```diff
@@ -303,6 +307,10 @@ trajectory::InterpolatedPtr planCRRTConnect(
     double _minStepsize,
     double _minTreeConnectionDistance)
 {
+
```
jslee02: Nit: Please remove unnecessary this whitespace

include/aikido/constraint/Testable.hpp
```diff
@@ -14,8 +14,10 @@ class Testable
   virtual ~Testable() = default;
 
   /// Returns true if state satisfies this constraint.
+  /// \param _givenState: true if state is user-provided
```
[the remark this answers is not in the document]
jslee02: Nit: We don't use `:` (semicolon) for a parameter docstring.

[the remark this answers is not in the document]
jslee02: I guess `_state` is set to `false` when the checking state is provided by a planner, and, in that case, we don't want to terminate the planning due (could reject the sample though).

I think we don't need `_givenState` parameter for enforcing the planner to terminate. It's actually done by the caller (see [this](https://github.com/personalrobotics/aikido/pull/235/files#diff-af695089f1b391078609c33041298291R29)). `_givenState` is just using to determine whether a concrete `Testable` prints.

I agree with @mkoval that changing the way of receiving debug information.

brianhou:
I think we'll wind up rewriting this entirely per @mkoval's suggestion, so I'm just going to close this PR. The branch will still be somewhat useful until we have the new implementation, though.

PR #242, opened 2017-10-18
Add WorldInteractiveMarkerViewer.

brianhou:
This subclasses `aikido::rviz::InteractiveMarkerViewer` to create a viewer for a `dart::simulation::World`. This would automatically update the viewer as skeletons are added to the `World`, which is a first step toward using them properly in all of our demo scripts.

@psigen mentioned that `dart::simulation::World` has some additional overhead if we aren't doing things with dynamics. I'm using `World` for now for lack of something better in DART, but we should seriously think about a better longer-term solution.

Other issues:
- [x] There doesn't seem to be a mutex associated with `World`s, so there might be some thread-safety issues here?
- [x] Cloning a `World` doesn't replicate the transforms or the joint states. [This line](https://github.com/dartsim/dart/blob/master/dart/simulation/World.hpp#L94) in DART suggests that this is still a TODO...

***

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
   * Added World class: [#243](https://github.com/personalrobotics/aikido/pull/243)
 
+* RViz
+
+  * Added WorldInteractiveMarkerViewer: [#242](https://github.com/personalrobotics/aikido/pull/242)
+
 * Build & Testing & ETC
 
   * Changed to use size_t over std::size_t: [#230](https://github.com/personalrobotics/aikido/pull/230)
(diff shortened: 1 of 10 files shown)


brianhou:
When I'm using this new viewer class in the (poorly-named) block_stacking demo, I sometimes get this error when I exit the script:
> *** Error in `./devel/bin/block_stacking': corrupted double-linked list: 0x0000000000660e40 ***

[This StackOverflow answer](https://stackoverflow.com/a/28762523) suggests that heap overflow might be to blame, but I'm not sure where this could be happening. If anyone sees something that could be causing that issue, please let me know!

PR #243, opened 2017-10-18
Initial implementation of kinematic World

brianhou:
This is a first attempt at a kinematic version of `dart::simulation::World`. Using a DART `World` as our default environment representation forces users to instantiate a dynamics engine, which is unnecessary overhead for many of our tasks. We'll add more functionality in future PRs.

***

### DIFF
--- a/include/aikido/planner/World.hpp
+++ b/include/aikido/planner/World.hpp
@@ -11,41 +11,48 @@ class World
 {
 public:
   /// Construct a kinematic World.
+  /// \param name Name for the new World
   World(const std::string& name = "");
 
   virtual ~World();
+
+  /// Create a new World inside of a shared_ptr
+  /// \param name Name for the new World
+  static std::shared_ptr<World> create(const std::string& name = "");
 
   /// Create a clone of this World. All Skeletons will be copied over.
   /// \param newName Name for the cloned World
   std::shared_ptr<World> clone(const std::string& newName = "") const;
 
-  /// Create a new World inside of a shared_ptr
-  static std::shared_ptr<World> create(const std::string& name = "");
-
   /// Set the name of this World
+  /// \param newName New name for this World
   std::string setName(const std::string& newName);
 
   /// Get the name of this World
   const std::string& getName() const;
 
   /// Find a Skeleton by index
+  /// \param i Index of desired Skeleton
   dart::dynamics::SkeletonPtr getSkeleton(std::size_t i) const;
 
   /// Find a Skeleton by name
+  /// \param name Name of desired Skeleton
   dart::dynamics::SkeletonPtr getSkeleton(const std::string& name) const;
 
   /// Get the number of Skeletons
   std::size_t getNumSkeletons() const;
 
   /// Add a Skeleton to this World
+  /// \param skeleton Skeleton to add to the World
   std::string addSkeleton(const dart::dynamics::SkeletonPtr& skeleton);
 
   /// Remove a Skeleton from this World
+  /// \param skeleton Skeleton to remove from the World
   void removeSkeleton(const dart::dynamics::SkeletonPtr& skeleton);
 
   // TODO: Add methods for registering callbacks?
 
-  // Get the mutex that protects the state of this World.
+  /// Get the mutex that protects the state of this World.
   std::mutex& getMutex() const;
 
 protected:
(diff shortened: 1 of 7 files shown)


include/aikido/planner/World.hpp
```diff
@@ -0,0 +1,65 @@
...
+class World
+{
+public:
+  /// Construct a kinematic World.
+  World(const std::string& name);
+
+  virtual ~World() = default;
+
+  /// Create a clone of this World. All Skeletons will be copied over.
+  std::shared_ptr<World> clone(const std::string& name) const;
```
jslee02: Nit: Need docstring for the parameter.

Also, I like you used `newName` in the source file for this parameter. It would be good to use the same name here too.

src/planner/World.cpp
```diff
@@ -0,0 +1,134 @@
...
+  // TODO: handle name conflicts
+  // dart::simulation::World uses dart::common::NameManager
+
+  return skeleton->getName();
+}
+
+//==============================================================================
+void World::removeSkeleton(const dart::dynamics::SkeletonPtr& skeleton)
+{
+  std::lock_guard<std::mutex> lock(mMutex);
```
jslee02: Nit: It also looks safe even we lock after the below nullity check because it refers only the parameter.

include/aikido/planner/World.hpp
```diff
@@ -0,0 +1,66 @@
...
+#include <dart/dart.hpp>
+
+namespace aikido {
+namespace planner {
+
+class World
+{
+public:
+  /// Construct a kinematic World.
+  World(const std::string& name);
```
prasbg: If we are creating or cloning worlds inside of an optimizer or something, does it really make sense to need to generate names for all of them?  I think DART has some mechanisms for autogenerating IDs if names are not set, something like that might make sense here if this is just for semantic purposes.

tests/planner/test_World.cpp
```diff
@@ -0,0 +1,93 @@
...
+{
+public:
+  using SkeletonPtr = dart::dynamics::SkeletonPtr;
+
+  WorldTest()
+    : skel1{dart::dynamics::Skeleton::create("skel1")}
+    , skel2{dart::dynamics::Skeleton::create("skel2")}
+    , skel3{dart::dynamics::Skeleton::create("skel3")}
+  {
+    mWorld = std::make_shared<aikido::planner::World>("test");
```
prasbg: Should `World` have a similar `create()` helper to the above one that `Skeleton` has?

psigen:
Oops, posted with wrong account 🤷‍♀️

include/aikido/planner/World.hpp
```diff
@@ -0,0 +1,72 @@
...
+  /// Name of this World
+  std::string mName;
+
+  /// Skeletons in this World
+  std::vector<dart::dynamics::SkeletonPtr> mSkeletons;
+
+  mutable std::mutex mMutex;
+
+  /// NameManager for keeping track of Worlds
+  static dart::common::NameManager<World*> worldNameManager;
```
jslee02: I believe our convention is `mStaticVariable` even for static member variables.

include/aikido/planner/World.hpp
```diff
@@ -0,0 +1,72 @@
...
+
+  /// NameManager for keeping track of Skeletons
+  dart::common::NameManager<dart::dynamics::SkeletonPtr> mSkeletonNameManager;
+};
+
+typedef std::shared_ptr<World> WorldPtr;
```
jslee02: Nit: Let's use C++11 style of type aliasing: `using WorldPtr = std::shared_ptr<World>()`.

include/aikido/planner/World.hpp
```diff
@@ -0,0 +1,72 @@
...
+  World(const std::string& name = "");
+
+  virtual ~World();
+
+  /// Create a clone of this World. All Skeletons will be copied over.
+  /// \param newName Name for the cloned World
+  std::shared_ptr<World> clone(const std::string& newName = "") const;
+
+  /// Create a new World inside of a shared_ptr
+  static std::shared_ptr<World> create(const std::string& name = "");
```
jslee02: In case we provide one creator function, then it should be returning `std::unique_ptr` because we can convert `std::unique_ptr` to `std::shared_ptr` but the opposite is not allowed.

brianhou: Sure. I mostly copied this from DART :sweat_smile:

jslee02: Yeah, DART also needs to be tweaked. :disappointed: There is (or was) ongoing discussion: https://github.com/dartsim/dart/pull/845

PR #245, opened 2017-10-19
[WIP] Setup CodeDocs: Online doxygen-based API documentation publisher

jslee02:
***

- [x] Format code with `make format` (N/A)

- [x] Set version target by selecting a milestone on the right side (N/A)


brianhou:
It seems that this hosts the documentation at https://codedocs.xyz/personalrobotics/aikido/.

Would it make more sense to either (a) host it on GitHub Pages or (b) host it on the lab website?

jslee02:
AFAIK, both ways should be done manually, but CodeDocs is automated. It's updated per commit. The downside of using CodeDocs is that it always build for master branch.

I think we could (1) use CodeDocs for automated API documentation generation for master branch and (2) upload manually API documentation for released AIKIDO versions.

psigen:
I'm not sure I follow the reasoning here.

In general, I agree with @brianhou.  Depending on an externally built and hosted documentation system is not desirable because it constrains you to their quality of service.  I previously hosted some documentation on [ReadTheDocs](https://readthedocs.org/) for another project and ended up having horrible issues where the docs were slow to load, didn't build for days, or built inconsistently (had errors which I could not debug because I got different results when building locally).

https://github.com/personalrobotics/aikido/pull/245#issuecomment-338044465
> AFAIK, both ways should be done manually, but CodeDocs is automated. It's updated per commit. The downside of using CodeDocs is that it always build for master branch.

I'm not sure I understand why this is somehow specific to CodeDocs.

One very common approach is simply to put the documentation as a build step in `TravisCI` (or any equivalent CI) and simply give it a deploy key to your `gh-pages` or a server location.  I have done this before with great success.

It is also pretty easy to use build variables in CI to prefix the installation path of the output HTML artifacts (e.g. `master` goes to https://personalrobotics.github.io/aikido/docs/master`) and constrain documentation builds to only run on select branches (e.g. `master` and named tags).

----

**TL:DR:** I would _highly_ recommend using a doc generator that can be run locally to produce static HTML, then use manual or conventional CI processes to put the static HTML artifact on a web server of your choosing.  GitHub Pages works really well for this, and is also an intuitive place for people to look for documentation.

This gives us the most flexibility (if we decide to change generators), reliability (we can manually build, upload, and correct issues if needed), and customizability (we can control which documentation gets built and retained)

jslee02:
@psigen I totally agree with you. CodeDocs wouldn't be the best solution as a long run solution. This is a stopgap solution with minimal effort. 

I'm okay not to merge this, but I think we could start with this temporary solution that doesn't hurt the current codebase much then move on a better solution like you suggested. I would gratefully accept it once the setup is done. 😎

jslee02:
Closing in favor of @psigen 's [suggestion](https://github.com/personalrobotics/aikido/pull/245#issuecomment-338445870).

PR #247, opened 2017-10-22
Add methods for removing groups in NonColliding.

brianhou:
***

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 * Constraint
 
-  * Add methods for removing groups from NonColliding constraints. [#247](https://github.com/personalrobotics/aikido/pull/247)
+  * Added methods for removing groups from NonColliding constraints. [#247](https://github.com/personalrobotics/aikido/pull/247)
 
 * Planner
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ### 0.2.0 (201X-XX-XX)
 
+* Constraint
+
+  * Add methods for removing groups from NonColliding constraints. [#247](https://github.com/personalrobotics/aikido/pull/247)
+
 * Planner
 
   * Added World class: [#243](https://github.com/personalrobotics/aikido/pull/243)
--- a/include/aikido/constraint/NonColliding.hpp
+++ b/include/aikido/constraint/NonColliding.hpp
@@ -50,9 +50,20 @@ public:
       std::shared_ptr<dart::collision::CollisionGroup> _group1,
       std::shared_ptr<dart::collision::CollisionGroup> _group2);
 
+  /// Remove collision check between group1 and group2.
+  /// \param group1 First collision group.
+  /// \param group2 Second collision group.
+  void removePairwiseCheck(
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ### 0.2.0 (201X-XX-XX)
 
+* Constraint
+
+  * Add methods for removing groups from NonColliding constraints. [#247](https://github.com/personalrobotics/aikido/pull/247)
```
jslee02: Nit: The current convention is to use past tense for the leading verb.

include/aikido/constraint/NonColliding.hpp
```diff
@@ -50,10 +50,21 @@ class NonColliding : public Testable
       std::shared_ptr<dart::collision::CollisionGroup> _group1,
       std::shared_ptr<dart::collision::CollisionGroup> _group2);
 
+  /// Remove collision check between group1 and group2.
+  /// \param group1 First collision group.
```
jslee02: Nit: The parameter names in this docstring are not consistent with the actual parameter names. I believe the convention is not to use a leading underscore for parameter names, but it seems this file doesn't comply. Let's fix this later as wholesale.

src/constraint/NonColliding.cpp
```diff
@@ -65,11 +65,48 @@ void NonColliding::addPairwiseCheck(
...
+      groupsToPairwiseCheck.begin(),
+      groupsToPairwiseCheck.end(),
+      std::make_pair(_group1, _group2));
+  if (it != groupsToPairwiseCheck.end())
+  {
+    groupsToPairwiseCheck.erase(it);
+    return;
+  }
+
+  // Check for reverse pair
```
jslee02: I think we could remove this double checking by storing the groups in a pair (e.g., ascending or descending in the pointer addresses). For this, we should:
* change `addPairwiseCheck(...)` to add group pair in ascending order something like:
```cpp
  if (_group1 < _group2)
    groupsToPairwiseCheck.emplace_back(std::move(_group1), std::move(_group2));
  else
    groupsToPairwiseCheck.emplace_back(std::move(_group2), std::move(_group1));
```
* change `removePairwiseCheck(...)` to find the ascending-ordered group pair

More importantly, I think we could use `std::unordered_set` instead of `std::vector` for better performance in adding and removing because we don't need random access, and the order of pairs is not important. In case we switch to `std::unordered_set` we might directly use `std::unordered_set::erase` without finding the pair.

src/constraint/NonColliding.cpp
```diff
@@ -65,11 +65,48 @@ void NonColliding::addPairwiseCheck(
...
 }
 
+//==============================================================================
+void NonColliding::removeSelfCheck(
+    std::shared_ptr<dart::collision::CollisionGroup> _group)
+{
+  auto it
+      = std::find(groupsToSelfCheck.begin(), groupsToSelfCheck.end(), _group);
+  if (it != groupsToSelfCheck.end())
+    groupsToSelfCheck.erase(it);
```
jslee02: I think `groupsToSelfCheck` also can be `std::unordered_set` for the same reason with `groupsToPairwiseCheck`.

PR #251, opened 2017-10-27
Queue trajectories in TrajectoryExecutor

brianhou:
This PR adds methods to `aikido::control::TrajectoryExecutor` to support queuing of trajectories. This is useful for pipelining, since we no longer have to wait for a trajectory to finish executing before sending the next trajectory.

***

### DIFF
--- a/include/aikido/control/KinematicSimulationTrajectoryExecutor.hpp
+++ b/include/aikido/control/KinematicSimulationTrajectoryExecutor.hpp
@@ -33,7 +33,8 @@ public:
   /// \return future<void> for trajectory execution. If trajectory terminates
   ///        before completion, future will be set to a runtime_error.
   /// \throws invalid_argument if traj is invalid.
-  std::future<void> execute(trajectory::TrajectoryPtr traj, bool skip) override;
+  std::future<void> execute(
+      trajectory::TrajectoryPtr traj, bool skip = false) override;
 
   /// \copydoc PositionCommandExecutor::step()
   ///
--- a/include/aikido/control/KinematicSimulationTrajectoryExecutor.hpp
+++ b/include/aikido/control/KinematicSimulationTrajectoryExecutor.hpp
@@ -29,10 +29,11 @@ public:
   /// \param traj Trajectory to be executed. Its StateSpace should be a
   ///        MetaStateSpace over MetaSkeleton, and the dofs in the metaskeleton
   ///        should be all in skeleton passed to the constructor.
+  /// \param skip Whether to skip to the end of the trajectory
   /// \return future<void> for trajectory execution. If trajectory terminates
   ///        before completion, future will be set to a runtime_error.
   /// \throws invalid_argument if traj is invalid.
-  std::future<void> execute(trajectory::TrajectoryPtr traj) override;
+  std::future<void> execute(trajectory::TrajectoryPtr traj, bool skip) override;
 
   /// \copydoc PositionCommandExecutor::step()
   ///
--- a/include/aikido/control/TrajectoryExecutor.hpp
+++ b/include/aikido/control/TrajectoryExecutor.hpp
@@ -1,7 +1,11 @@
 #ifndef AIKIDO_CONTROL_TRAJECTORYEXECUTOR_HPP_
 #define AIKIDO_CONTROL_TRAJECTORYEXECUTOR_HPP_
 
+#include <atomic>
 #include <future>
+#include <mutex>
+#include <queue>
+#include <thread>
 #include <aikido/trajectory/Trajectory.hpp>
 #include "TrajectoryResult.hpp"
 
@@ -11,7 +15,7 @@ namespace control {
 class TrajectoryExecutor
 {
 public:
-  virtual ~TrajectoryExecutor() = default;
+  virtual ~TrajectoryExecutor();
 
   /// Execute traj and set future upon completion.
   /// \param traj Trajectory to be executed.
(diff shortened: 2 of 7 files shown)


src/control/TrajectoryExecutor.cpp
```diff
@@ -0,0 +1,82 @@
+#include <iostream>
```
jslee02: Nit: Remove `#include <iostream>` since it's not used.

include/aikido/control/KinematicSimulationTrajectoryExecutor.hpp
```diff
@@ -29,10 +29,12 @@ class KinematicSimulationTrajectoryExecutor : public TrajectoryExecutor
   /// \param traj Trajectory to be executed. Its StateSpace should be a
   ///        MetaStateSpace over MetaSkeleton, and the dofs in the metaskeleton
   ///        should be all in skeleton passed to the constructor.
+  /// \param skip Whether to skip to the end of the trajectory
```
mkoval: I think this belongs in a different _implementation_ of the `TrajectoryExecutor` interface, not in _each implementation_ of that interface. E.g. we should define a `InstantaneousTrajectoryExeuctor` (or a similar name) that implements `execute()` by jumping to the end of the trajectory.

include/aikido/control/TrajectoryExecutor.hpp
```diff
@@ -11,14 +15,51 @@ namespace control {
...
+  virtual ~TrajectoryExecutor();
 
   /// Execute traj and set future upon completion.
   /// \param traj Trajectory to be executed.
-  virtual std::future<void> execute(trajectory::TrajectoryPtr _traj) = 0;
+  /// \param skip Whether to skip to the end of the trajectory (only used by
+  ///        simulation trajectory executors)
+  virtual std::future<void> execute(
+      trajectory::TrajectoryPtr _traj, bool _skip = false)
+      = 0;
```
mkoval: What does `execute()` do if trajectories are queued for execution? Why would I ever call this over `queue()` below?

include/aikido/control/TrajectoryExecutor.hpp
```diff
@@ -11,14 +15,51 @@ namespace control {
...
+
+  /// Queue a trajectory for future execution. If there are no trajectories on
+  /// the queue or currently being executed, the trajectory should be executed
+  /// immediately.
+  /// \param traj Trajectory to queue and execute
+  void queue(trajectory::TrajectoryPtr _traj);
+
+  /// Toggle whether trajectories should be automatically dequeued and executed.
+  /// \param flag Whether to automatically execute trajectories from the queue
+  void setExecuteFromQueue(bool flag);
```
mkoval: Why would I ever want to set this to `false`?

include/aikido/control/TrajectoryExecutor.hpp
```diff
@@ -11,14 +15,51 @@ namespace control {
...
+
+  /// Block until all trajectories in the queue have been executed.
+  void executeAllFromQueue();
```
mkoval: Why is this function necessary? If we can't eliminate it entirely - which I think we should try to do - I suggest turning this into a non-blocking function that returns an `std::future<void>` that completes.

brianhou:
One limitation that I ran into with the current implementation is that the arm and hand have different Executors, which means that we can't queue actions from both at the same time. This is useful if we queue an arm trajectory and then a hand preshape, e.g. moving the arm to a glass and then closing the hand once the trajectory finishes.

mkoval:
I think the type of queuing you're looking for - i.e. to execute a multi-step plan - is best captured in a simple variant of the [command pattern](https://sourcemaking.com/design_patterns/command). This is precisely what `ExecutableSolution` is in Magi.

The key idea is to define an interface for executing an arbitrary action. Then, define a concrete implementation of the class for each action. E.g.:
```c++
class Action
{
  virtual std::future<void> execute(WorldPtr world) = 0;
};

class ExecuteTrajectoryAction : public Action
{
  ExecuteTrajectoryAction(TrajectoryExecutorPtr executor, TrajectoryPtr trajectory)
    : mExecutor{executor}
    , mTrajectory{trajectory}
  {
  }

  virtual std::future<void> execute(WorldPtr world) override
  {
    // Verify that 'world' is in the expected starting state.
    return mExecutor->execute(mTrajectory);
  }
}

// A similar concrete implementation for PositionCommandExecutor.
```

Once you have this framework in place, it's pretty easy to write a queue:
```c++
void execute_sequence(WorldPtr world, std::vector<std::unique_ptr<Action>> actions)
{
  for (const auto& action : actions) {
    auto future = action->execute(world);
    future.wait();
  }
}
```

In general, I strongly recommend taking a look at the architecture of `Action`, `Solution`, and `ExecutableSolution` in Magi. I think that was a _really_ powerful concept that hits a sweet spot between power (ability to parallelize planning, backtrack, etc) and complexity (not requiring general-purpose `World` cloning).

---

That being said, I still think it is valuable to add queuing to a trajectory executor for an _entirely different_ reason. It is useful to be able to preempt trajectories during execution to change the motion of the robot without stopping, e.g. if there are dynamic obstacles or a better plan becomes available. This requires the API to the controller (in our case: `TrajectoryExecutor`) to support the concepts of queuing and preemption.

I doubt this is a high priority, but it's something to keep in mind for the future. 🙂

gilwoolee:
@mkoval Thanks for the comments! I'm currently implementing MAGI-cpp (for AIKIDO). One question. In original MAGI where we had `Action`, `Solution`, `ExecutableSolution`, `action.execute` would call `self.plan` to get a `solution` which gets postprocessed to get an `ExecutableSolution`, which then gets executed. The concept of `ExecuteTrajectoryAction` seems to bypass all this. Is there any advantage to this?

I think both `ExecuteTrajectoryAction` and `ExecutePositionCommandAction` are such common cases that it makes sense to support them, but in general, we'd still want to follow what the original MAGI was doing, right?

include/aikido/control/TrajectoryExecutor.hpp
brianhou: I think it would've errored, but you're right that we'd always be calling `queue` over `execute`. I prefer the new `execute` semantics that you suggested.

mkoval:
> The concept of `ExecuteTrajectoryAction` seems to bypass all this. Is there any advantage to this?

That was not my intent. I was actually thinking of `ExecuteTrajectoryAction` as actually being a [albeit poorly named] `ExecutableSolution` for executing a trajectory. We would still have a menagerie of `Action` and `Solution` classes to wrap motion planning, post-processing, hand commands, etc. 😄

include/aikido/control/KinematicSimulationTrajectoryExecutor.hpp
gilwoolee: Maybe I'm misunderstanding this, but how do the concrete classes utilize both `TrajectoryExecutor` and `InstantaneousTrajectoryExecutor`? Don't we typically want to have one `Executor` but sometimes just want to jump to the end? e.g. How would it work with `magi.jump()` method?

brianhou:
Thank you for the feedback @mkoval! I think the useful parts of this PR have either been moved to #259, https://github.com/personalrobotics/magicpp/pull/1, or https://github.com/personalrobotics/magicpp/pull/8 so I'm going to close this.

PR #252, opened 2017-10-30
Fix WorldInteractiveMarkerViewer bug when removing Skeletons

brianhou:
If a `Skeleton` is removed and a new `Skeleton` with the same name is added quickly enough (before the viewer refreshes and removes the original `Skeleton`), the new `Skeleton` is not visualized. This PR fixes this bug by using the actual `SkeletonPtr` as the key in `WorldInteractiveMarkerViewer`, rather than its name.

At some point, we could fix this by registering callbacks to `World` that are called when adding or removing `Skeleton`s.

***

### DIFF
--- a/include/aikido/planner/World.hpp
+++ b/include/aikido/planner/World.hpp
@@ -39,6 +39,11 @@ public:
   /// \param name Name of desired Skeleton
   dart::dynamics::SkeletonPtr getSkeleton(const std::string& name) const;
 
+  /// Find a Skeleton by Skeleton, returning null if it is not in the World
+  /// \param skel Desired Skeleton
+  dart::dynamics::SkeletonPtr getSkeleton(
+      const dart::dynamics::SkeletonPtr skel) const;
+
   /// Get the number of Skeletons
   std::size_t getNumSkeletons() const;
 
--- a/include/aikido/planner/World.hpp
+++ b/include/aikido/planner/World.hpp
@@ -39,10 +39,10 @@ public:
   /// \param name Name of desired Skeleton
   dart::dynamics::SkeletonPtr getSkeleton(const std::string& name) const;
 
-  /// Find a Skeleton by Skeleton, returning null if it is not in the World
+  /// Returns true if the Skeleton is in this World.
   /// \param skel Desired Skeleton
-  dart::dynamics::SkeletonPtr getSkeleton(
-      const dart::dynamics::SkeletonPtr skel) const;
+  /// \return True if succeeded to find the Skeleton
+  bool hasSkeleton(const dart::dynamics::SkeletonPtr& skel) const;
 
   /// Get the number of Skeletons
   std::size_t getNumSkeletons() const;
--- a/include/aikido/rviz/WorldInteractiveMarkerViewer.hpp
+++ b/include/aikido/rviz/WorldInteractiveMarkerViewer.hpp
@@ -38,8 +38,8 @@ protected:
   /// Thread target for auto-updating the viewer
   void autoUpdate();
 
-  /// Mapping of Skeleton names to SkeletonMarkers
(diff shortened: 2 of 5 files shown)


include/aikido/planner/World.hpp
```diff
@@ -39,6 +39,11 @@ class World
   /// \param name Name of desired Skeleton
   dart::dynamics::SkeletonPtr getSkeleton(const std::string& name) const;
 
+  /// Find a Skeleton by Skeleton, returning null if it is not in the World
+  /// \param skel Desired Skeleton
+  dart::dynamics::SkeletonPtr getSkeleton(
```
jslee02: This function sounds like checking whether the skeleton exists in (registered to) this world. I would like to suggest changing the function name `containsSkeleton(...)` or `existsSkeleton(...)` (or `skeletonExists`) (we conventionally suffix `s` to the leading verb for an inquiry function) like [`NewtonsMethodProjectable::contains`](https://github.com/personalrobotics/aikido/blob/cbf1e5808325f75b3831a62760340e3a6d7261b3/include/aikido/constraint/NewtonsMethodProjectable.hpp#L45) and [`CatkinResourceRetriever::exists(...)`](https://github.com/personalrobotics/aikido/blob/2c24a9bd10a8412084d8979a4995a81fd4103259/include/aikido/io/CatkinResourceRetriever.hpp#L33) and returning `bool`.

I personally prefer `contains` over `exists` because `contains` would sound more natural when we use the function like:
```cpp
if (world.containsSkeleton(someSkeleton))
  ...
if (world.existsSkeleton(someSkeleton)) // or if(world.skeletonExists(...))
  ...
```

include/aikido/planner/World.hpp
```diff
@@ -39,6 +39,11 @@ class World
...
+      const dart::dynamics::SkeletonPtr skel) const;
```
jslee02: Nit: Let's use `const dart::dynamics::SkeletonPtr&` to avoid unnecessary increase of the reference count by copying `std::shared_ptr<...>`.

include/aikido/rviz/WorldInteractiveMarkerViewer.hpp
```diff
@@ -38,8 +38,8 @@ class WorldInteractiveMarkerViewer : public InteractiveMarkerViewer
   /// Thread target for auto-updating the viewer
   void autoUpdate();
 
-  /// Mapping of Skeleton names to SkeletonMarkers
-  std::map<std::string, SkeletonMarkerPtr> mSkeletonMarkers;
+  /// Mapping of Skeletons to SkeletonMarkers
+  std::map<dart::dynamics::SkeletonPtr, SkeletonMarkerPtr> mSkeletonMarkers;
```
jslee02: Although this is a stopgap solution (over [using callbacks](https://github.com/personalrobotics/aikido/pull/252#issue-269541118)), we might want to use `std::weak_pointer<...>` here because this class doesn't need to have the ownership as it just uses the skeletons only when they still exist in `mWorld`.

jslee02:
This is a good catch! I made some suggestions.

include/aikido/planner/World.hpp
mkoval: I don't think we can use `std::weak_ptr` here because it is not comparable or hashable. This makes sense because an unlocked `weak_ptr` can spontaneously degrade to `nullptr` if all owning `shared_ptr`s are destructed.

We could using a `const Skeleton*` as the key. But we'd need to be careful to not access the pointer after the `Skeleton` is destructed, e.g. by `lock()`ing a `std::weak_ptr` that we store inside the marker instance.

jslee02:
I changed the code according to the suggestions. @brianhou Could you take a look at the change for sure?

PR #258, opened 2017-11-17
Disable CI tests on macOS

jslee02:
Currently, Travis CI tests on macOS almost always timeout. This PR disables it as we don't have any actual use AIKIDO on macOS.

***

- [x] Format code with `make format` (N/A)

- [x] Set version target by selecting a milestone on the right side (N/A)


jslee02:
[It seems simply setting the Travis CI compiler configuration to Clang doesn't use Clang](https://travis-ci.org/personalrobotics/aikido/jobs/304013580#L3157-L3158). Probably this is related to the fact that we use docker image, and changing the compiler configuration doesn't affect on the default compiler of the docker image.

In this RP, I would like to remove testing AIKIDO using Clang. We might add it back with proper setting in a separate PR if necessary.

brianhou:
That seems somewhat strange; shouldn't we be able to set the `CMAKE_CXX_COMPILER` option via `CATKIN_CONFIG_OPTIONS`?

PR #259, opened 2017-11-19
Add Instantaneous and Queued TrajectoryExecutors

brianhou:
This is a better implementation than #251, following @mkoval's suggestions on that pull request. This PR also makes various changes to the `KinematicSimulation` and `Ros` `TrajectoryExecutors` implementations to make them more consistent.

***


brianhou:
This *almost* works when I'm running some scripts in simulation. I'm running into issues while using `QueuedTrajectoryExecutor` to wrap `KinematicSimulationTrajectoryExecutor`.

The problem is that `KinematicSimulationTrajectoryExecutor::execute` attempts to lock the meta-skeleton that it operates on [here](https://github.com/personalrobotics/aikido/blob/enhancement/brianhou/traj-queuing/src/control/KinematicSimulationTrajectoryExecutor.cpp#L59). For some reason, it's unable to acquire the lock when it's being wrapped by `QueuedTrajectoryExecutor`, although it works just fine when it's not being wrapped. I'm looking for other places where I might be neglecting to unlock a locked meta-skeleton (especially in `QueuedTrajectoryExecutor`), but no luck yet.

The actual queuing logic seems fine; if I comment out the skeleton locking/unlocking, the wrapped `KinematicSimulationTrajectoryExecutor` works.

brianhou:
Ack, just figured out what was happening. In `libherb`, we lock the entire robot meta-skeleton before calling `QueuedTrajectoryExecutor::step`. I'm calling the wrapped `TrajectoryExecutor::execute` from `QueuedTrajectoryExecutor::step`, which expects the arm meta-skeleton to be unlocked as it checks the dof names. I'll think about how to fix this.

brianhou:
@mkoval When you get a chance, would you mind looking over my proposed fix for this locking issue in https://github.com/personalrobotics/aikido/pull/259/commits/06672e5ba52bb5458dd388696aa813e9fe6070a9?

The short version is that the DOF validation that is performed by `[Instantaneous|KinematicSimulation]TrajectoryExecutor::execute` requires the robot `Skeleton` to be locked. However, `execute` is called from `QueuedTrajectoryExecutor::step`, which is called while the robot `Skeleton` is locked.

Instead, I'm now performing the DOF validation in `QueuedTrajectoryExecutor::execute`, rather than `QueuedTrajectoryExecutor::step`. Then, by the time the trajectory is actually sent to the underlying executor in `step`, it no longer needs to be validated and doesn't need to acquire the `Skeleton` lock.

include/aikido/control/InstantaneousTrajectoryExecutor.hpp
```diff
@@ -0,0 +1,53 @@
...
+  /// \throws invalid_argument if traj is invalid.
+  std::future<void> execute(trajectory::TrajectoryPtr traj) override;
+
+  // Do nothing.
+  void step() override;
+
+private:
+  ::dart::dynamics::SkeletonPtr mSkeleton;
+
+  std::unique_ptr<std::promise<void>> mPromise;
```
mkoval: This member variable seems to be unused.

src/control/KinematicSimulationTrajectoryExecutor.cpp
```diff
@@ -30,19 +31,24 @@ KinematicSimulationTrajectoryExecutor::KinematicSimulationTrajectoryExecutor(
 KinematicSimulationTrajectoryExecutor::~KinematicSimulationTrajectoryExecutor()
 {
   {
-    std::lock_guard<std::mutex> lockSpin(mMutex);
-    mInExecution = false;
+    std::lock_guard<std::mutex> lock(mMutex);
+    DART_UNUSED(lock); // Suppress unused variable warning
+
+    mInProgress = false;
```
mkoval: Set `mPromise` to an exception here to unblock anyone who is waiting on it.

src/control/KinematicSimulationTrajectoryExecutor.cpp
```diff
@@ -30,19 +31,24 @@ KinematicSimulationTrajectoryExecutor::KinematicSimulationTrajectoryExecutor(
...
-std::future<void> KinematicSimulationTrajectoryExecutor::execute(
+void KinematicSimulationTrajectoryExecutor::validate(
     trajectory::TrajectoryPtr traj)
 {
   if (!traj)
     throw std::invalid_argument("Traj is null.");
 
-  auto space = std::dynamic_pointer_cast<MetaSkeletonStateSpace>(
+  if (traj->metadata.executorValidated)
+    return;
```
mkoval: What's the purpose of this check? Is there any harm in doing the validation again?

src/control/QueuedTrajectoryExecutor.cpp
```diff
@@ -0,0 +1,91 @@
...
+    // Propagate the future's value or exception to the caller
+    try
+    {
+      mFuture.get();
+      promise->set_value();
+    }
+    catch (const std::exception& e)
+    {
+      promise->set_exception(std::current_exception());
+    }
```
mkoval: Why manually copy the result of `mFuture` to `promise`, rather than directly returning `mFuture` to the caller?

src/control/ros/RosTrajectoryExecutor.cpp
```diff
@@ -88,22 +91,32 @@ std::future<void> RosTrajectoryExecutor::execute(trajectory::TrajectoryPtr traj)
...
+void RosTrajectoryExecutor::validate(trajectory::TrajectoryPtr traj)
 {
-  using aikido::control::ros::toRosJointTrajectory;
   using aikido::statespace::dart::MetaSkeletonStateSpace;
 
   if (!traj)
     throw std::invalid_argument("Trajectory is null.");
 
+  if (traj->metadata.executorValidated)
+    return;
```
mkoval: See above. Why is this necessary?

include/aikido/control/QueuedTrajectoryExecutor.hpp
```diff
@@ -0,0 +1,57 @@
...
+  /// \param executor Underlying TrajectoryExecutor
+  QueuedTrajectoryExecutor(std::unique_ptr<TrajectoryExecutor> executor);
+
+  virtual ~QueuedTrajectoryExecutor();
+
+  // Documentation inherited.
+  void validate(trajectory::TrajectoryPtr traj) override;
+
+  /// Execute trajectory and set future upon completion. If another trajectory
+  /// is already running, queue the trajectory for later execution.
```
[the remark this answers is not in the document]
gilwoolee: Assuming that in the queue, one trajectory's endpoint is consistent with the beginning of the other, shouldn't we stop execution if one of them terminates with an error and set all the remaining promises to an exception? (This is what I do with `SequenceExecutableSolution` in magi)

include/aikido/control/TrajectoryExecutor.hpp
```diff
@@ -13,7 +12,13 @@ class TrajectoryExecutor
 public:
   virtual ~TrajectoryExecutor() = default;
 
-  /// Execute traj and set future upon completion.
+  /// Validate the traj in preparation for execution.
+  /// \param traj Trajectory to be validated
+  virtual void validate(trajectory::TrajectoryPtr traj) = 0;
```
gilwoolee: What exactly does it mean to validate a trajectory? Do we validate that it's collision free? Or that the robot's current state is in the same state as the beginning of the trajectory?

include/aikido/control/InstantaneousTrajectoryExecutor.hpp
```diff
@@ -0,0 +1,53 @@
...
+
+/// Instantaneously executes a trajectory in simulation by setting DOF positions
+/// to the end of the trajectory.
+class InstantaneousTrajectoryExecutor : public TrajectoryExecutor
+{
+public:
+  /// Constructor.
+  /// \param skeleton Skeleton to execute trajectories on.
+  ///        All trajectories must have dofs only in this skeleton.
+  InstantaneousTrajectoryExecutor(::dart::dynamics::SkeletonPtr skeleton);
```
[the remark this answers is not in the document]
brianhou: DART `MetaSkeleton`s don't come with an associated mutex, so we wouldn't be able to lock that mutex while checking whether the DOFs specified by the trajectory are in the skeleton. It seems necessary to lock the mutex while doing that. I'd otherwise be happy to use `MetaSkeleton` over `Skeleton`.

@jslee02 why don't `MetaSkeleton`s have mutexes?

include/aikido/control/InstantaneousTrajectoryExecutor.hpp
```diff
@@ -0,0 +1,53 @@
...
+  // Documentation inherited.
+  void validate(trajectory::TrajectoryPtr traj) override;
+
+  /// Instantaneously execute traj and set future upon completion.
+  /// \param traj Trajectory to be executed. Its StateSpace should be a
+  ///        MetaStateSpace over MetaSkeleton, and the dofs in the metaskeleton
+  ///        should be all in skeleton passed to the constructor.
+  /// \return future<void> for trajectory execution. If trajectory terminates
+  ///        before completion, future will be set to a runtime_error.
+  /// \throws invalid_argument if traj is invalid.
```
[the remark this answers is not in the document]
brianhou: It currently throws the `invalid_argument` immediately, rather than setting the `future<void>` to an exception. I was thinking that we should throw the error as early as possible, but I could see it in either place. What do you think?

include/aikido/control/InstantaneousTrajectoryExecutor.hpp
brianhou: It's used by `InstantaneousTrajectoryExecutor::execute`. I think it could be possible to avoid having this promise as a field in `InstantaneousTrajectoryExecutor` and just have it as a local variable in `execute`, if you think that would be better?

include/aikido/control/InstantaneousTrajectoryExecutor.hpp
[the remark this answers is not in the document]
jslee02: I don't think we have a good reason not to have a mutex in `MetaSkeleton`. Probably the mutex in `Skeleton` was added when `MetaSkeleton` wasn't yet introduced, and we just didn't add one to `MetaSkeleton` in introducing it?

@mxgrey Do you know why?

If we want to add a mutex to `MetaSkeleton`, I think we can just move the mutex in `Skeleton` to `MetaSkeleton`.

include/aikido/control/QueuedTrajectoryExecutor.hpp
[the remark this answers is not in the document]
brianhou: I think this is something to just add to the simulation trajectory executors. I've created #272 to track this.

src/control/KinematicSimulationTrajectoryExecutor.cpp
brianhou: This is how I'm attempting to avoid the deadlock issue I mentioned here: https://github.com/personalrobotics/aikido/pull/259#issuecomment-345492616.

In `libherb`, `Herb::step` acquires the lock on the robot skeleton before calling `QueuedTrajectoryExecutor::step`. If we re-validate, then we'll attempt to acquire the lock again which will cause a deadlock. I think a (dangerous?) alternative is to not lock the skeleton while doing this check.

src/control/QueuedTrajectoryExecutor.cpp
brianhou: I think there are two types of futures: ones from a promise made by the `QueuedTrajectoryExecutor` to the client, and ones from a promise made by the underlying `TrajectoryExecutor` to the `QueuedTrajectoryExecutor`.

I believe this distinction is necessary because the first type must be returned immediately after the client calls `QueuedTrajectoryExecutor::execute` so that the client has something to `wait` on. However, the second type isn't created until the underlying `TrajectoryExecutor::execute` is called (when the trajectory is popped off the queue), so it can't be returned immediately by `QueuedTrajectoryExecutor::execute`.

I think this means that I have to manually copy the result of `mFuture` to `promise`, but if there's another way I'd be happy to try that too!

In this implementation, `mPromiseQueue` holds all the promises that have been made by the `QueuedTrajectoryExecutor` to the client, while `mFuture` is the current future that the `QueuedTrajectoryExecutor` is waiting on (from the underlying `TrajectoryExecutor`).

include/aikido/control/TrajectoryExecutor.hpp
brianhou: Validating currently just means that the trajectory's DOFs match the DOFs from the skeleton that is being controlled by the `TrajectoryExecutor`. It could do more though -- I was thinking of it as a function for checking things that can be checked before executing the trajectory (like whether the DOFs match up). That said, I only added this to try and work around the deadlock issue.

I think verifying that the current state is in the same state as the beginning of the trajectory wouldn't fall under this category; we want to check that immediately before we actually execute the trajectory. I've added #272 for that.

I think validating whether the trajectory is collision-free shouldn't be the job of the executor, since it has no idea what the world looks like while it's executing.

brianhou:
Thanks for the comments, @mkoval! :heart: 

> In almost all cases, we should use MAGI to handle this queuing instead of TrajectoryExecutor.
> The best reason to add queuing to the TrajectoryExecutor API is to support trajectory queuing and preemption in hardware drivers - which this PR doesn't support.

Agreed. I'm not sure whether these changes will be useful, but I figured since we already had a somewhat-working version of the code that I'd clean it up in case it's a good jumping-off point for something else in the future. I've added `QueuedTrajectoryExecutor::abort`, but it seems difficult to actually abort an in-progress trajectory so this only aborts queued trajectories that haven't begun execution.

> What is the purpose of the trajectory tag added in this PR?

This stemmed out of a conversation with @siddhss5 where he mentioned that we should have additional metadata associated with our trajectories (like we did in `prpy`) about whether the trajectory was smooth, constrained, which planner was used, etc.

Adding this particular piece of metadata was just so that I could avoid adding an argument to `TrajectoryExecutor::execute` that decided whether or not the trajectory needed to be validated (again, to avoid the deadlock situation). It's not a particularly good reason :sweat:

mkoval:
> I'm not sure whether these changes will be useful, but I figured since we already had a somewhat-working version of the code that I'd clean it up in case it's a good jumping-off point for something else in the future.

I agree that this is potentially useful once we have more control over trajectory execution on real hardware in the future (cancelling, preempting, etc). 👍 

> I've added `QueuedTrajectoryExecutor::abort()`, but it seems difficult to actually abort an in-progress trajectory so this only aborts queued trajectories that haven't begun execution.

If you make `abort()` part of the generic `TrajectoryExecutor` API, then `QueuedTrajectoryExecutor` could call `abort()` on the underlying trajectory executor. This seems pretty straightforward for `KinematicTrajectoryExecutor`, since we could set a flag that interrupts the next iteration of the simulation loop.

We can eventually support this on `RosTrajectoryExecutor` by canceling the actionlib goal, once we support that on the ReWD controller side. That's probably outside the scope of this pull request. 🙃 

> Adding this particular piece of metadata was just so that I could avoid adding an argument to `TrajectoryExecutor::execute()` that decided whether or not the trajectory needed to be validated (again, to avoid the deadlock situation). It's not a particularly good reason 😓

Trajectory tags in general seem pretty useful for all the reasons that @siddhss5 gave. I'm not convinced that this is a great application of them, though. 😬 Unfortunately, I don't have a better solution to fixing the deadlock.

brianhou:
Thanks! I've added `abort` to the `TrajectoryExecutor` interface, and I'll merge this PR once Travis says we're good to go.

> I'm not convinced that this is a great application of [trajectory tags], though. Unfortunately, I don't have a better solution to fixing the deadlock.

The options I've thought of are either (1) splitting up the validation like this [and either having this silly trajectory tag, or just moving this boolean from a trajectory tag to an `execute` argument], (2) locking the skeleton in `Herb::step` and passing the owned lock to the executors, (3) not locking the skeleton in `Herb::step`, or (4) not locking the skeleton in the executors. I might revisit this later and see how annoying (2) would be.

mkoval:
My original intent with the `step()` interface was (3). None of the `step()` functions should acquire locks internally. Instead, they should document their requirements (e.g. "requires the Skeleton passed to the constructor to be locked") and rely on the caller to acquire the lock. In the case of HERB, that would be `Herb::step()` in `libherb`. Would that solve the problem?

brianhou:
Not really, we're actually still following your original intent. The tricky part is writing `QueuedTrajectoryExecutor::step` so that it doesn't need to acquire any locks (since it calls `TrajectoryExecutor::execute` which makes no such promises). I'm generally in favor of the architecture that we have currently.

Issue #257, opened 2017-11-17
Incorrect box constraint initialization in `computeJointVelocityFromTwist`


dqyi11:

dqyi11:
In current implementation, initial guess is unset (by default all elements are zero).
When upper bounds and lower bounds are same signs, initial guess will be out of bounds, thus `NLopt` will throw an invalid argument error.

We will use current velocities of MetaSkeleton as initial guess.
Initial guess will be updated according to upper and lower bounds.

mkoval:
Good catch! That explains why we were seeing that warning, even though having upper and lower bounds with the same sign is not an issue. Your fix makes sense to me.

dqyi11:
Fixed in #262

PR #262, opened 2017-11-20
Avoid invalid argument exceptions in computeJointVelocityFromTwist

dqyi11:
Add initial guess (using current MetaSkeleton velocities) to `NloptSolver` in `aikido::planner::vectorfield::computeJointVelocityFromTwist`.
Update initial guess according to velocity bounds to avoid `invalid argument` thrown from `nlopt`.
This is a fix to #257 .

***


src/planner/vectorfield/VectorFieldUtil.cpp
```diff
@@ -115,19 +116,22 @@ bool computeJointVelocityFromTwist(
...
+    initialGuess[i] = std::max(velocityLowerLimits[i], initialGuess[i]);
 
     if (position + _timestep * velocityUpperLimit
         > positionUpperLimit - _padding)
       velocityUpperLimits[i] = std::min(
-          velocityUpperLimits[i],
-          ((velocityUpperLimits[i] - _padding) - position) / _timestep);
+          velocityUpperLimit,
+          ((positionUpperLimit - _padding) - position) / _timestep);
+    initialGuess[i] = std::min(velocityUpperLimits[i], initialGuess[i]);
```
brianhou: Maybe we can use `std::clamp` instead?

src/planner/vectorfield/VectorFieldPlanner.cpp
```diff
@@ -196,29 +198,24 @@ std::unique_ptr<aikido::trajectory::Spline> planPathByVectorField(
...
   }
-  // Re-throw whatever error caused planning to fail.
-  else if (terminationError)
+
+  if (cacheIndex >= 0)
   {
-    std::rethrow_exception(terminationError);
+    return convertToSpline(knots, cacheIndex, _stateSpace);
   }
   else
```
brianhou: Does this ever happen? We need to make sure to avoid throwing exceptions from these planning methods. I think we should be printing some warning and returning `nullptr` instead.

src/planner/vectorfield/VectorFieldPlanner.cpp
```diff
@@ -196,29 +198,24 @@ std::unique_ptr<aikido::trajectory::Spline> planPathByVectorField(
            && terminationStatus
                   != VectorFieldPlannerStatus::CACHE_AND_TERMINATE);
 
-  if (cacheIndex >= 0)
+  // Print the termination condition.
```
brianhou: Is this refactor actually correct? This was a somewhat hasty change to stop throwing exceptions.

dqyi11: Yes. I think the logic is the same with previous one. The difference is that no exception will be thrown as long as a planned trajectory is returned.

src/planner/vectorfield/VectorFieldPlanner.cpp
dqyi11: It is possible but not currently reproducible in current test cases. 
It is a good idea to throw warning instead. I think all this part of code will be changed in a new version VFP.

src/planner/vectorfield/VectorFieldUtil.cpp
dqyi11: `std::clamp` seems a feature since `c++17`.

src/planner/vectorfield/VectorFieldPlanner.cpp
```diff
@@ -127,11 +127,13 @@ std::unique_ptr<aikido::trajectory::Spline> planPathByVectorField(
     {
       if (!_vectorFiledCb(_stateSpace, t, dq))
       {
+        dtwarn << "Failed evaluating VectorField." << std::endl;
```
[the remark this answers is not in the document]
dqyi11: Good catch. This part will be rewritten in the new vector field planner. Let's fix that in the new one?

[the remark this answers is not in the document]
jslee02: Can we simply remove this in this PR then?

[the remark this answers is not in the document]
dqyi11: Yes. It is moved now.

Issue #261, opened 2017-11-20
Need to be able to compare two Worlds' states.


gilwoolee:
We need a way to check whether two `world`'s are in the same state ("equal" from planning/execution perspective). This is useful for multi-step planning/execution, to check before execution whether the `world` is in the same state as it expected during planning. I think this can be done by adding a method that checks whether all skeletons have the same configuration.

brianhou:
Resolved by #265.

PR #263, opened 2017-11-20
World equal configuration

gilwoolee:
This PR addresses #261 .

I added `equalConfiguration` method but I am happy to change the name if anyone has better suggestions.

***

### DIFF
--- a/src/planner/World.cpp
+++ b/src/planner/World.cpp
@@ -151,5 +151,23 @@ std::mutex& World::getMutex() const
   return mMutex;
 }
 
+//==============================================================================
+bool World::equalConfiguration(const World* otherWorld) const
+{
+  if (otherWorld->mSkeletons.size() != mSkeletons.size())
+    return false;
+
+  for (size_t idx = 0; idx < mSkeletons.size(); ++idx)
+  {
+    if (mSkeletons[idx]->getName() != otherWorld->mSkeletons[idx]->getName())
+      return false;
+    if (mSkeletons[idx]->getConfiguration()
+      != otherWorld->mSkeletons[idx]->getConfiguration())
+      return false;
+  }
+
+  return true;
+}
+
 } // namespace planner
 } // namespace aikido
--- a/src/planner/World.cpp
+++ b/src/planner/World.cpp
@@ -152,21 +152,48 @@ std::mutex& World::getMutex() const
 }
 
 //==============================================================================
-bool World::equalConfiguration(const World* otherWorld) const
+bool World::State::equals(const State& other)
 {
-  if (otherWorld->mSkeletons.size() != mSkeletons.size())
+  if (configurations.size() != other.configurations.size())
     return false;
 
-  for (size_t idx = 0; idx < mSkeletons.size(); ++idx)
+  for (size_t idx = 0; idx < configurations.size(); ++idx)
   {
-    if (mSkeletons[idx]->getName() != otherWorld->mSkeletons[idx]->getName())
-      return false;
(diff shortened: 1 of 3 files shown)


src/planner/World.cpp
```diff
@@ -151,5 +151,23 @@ std::mutex& World::getMutex() const
...
 
+//==============================================================================
+bool World::equalConfiguration(const World* otherWorld) const
+{
+  if (otherWorld->mSkeletons.size() != mSkeletons.size())
+    return false;
+
+  for (size_t idx = 0; idx < mSkeletons.size(); ++idx)
+  {
+    if (mSkeletons[idx]->getName() != otherWorld->mSkeletons[idx]->getName())
```
brianhou: Should we be requiring that the `Skeleton`s are in the exact same order in the `World`s? Maybe we can allow permutations?

gilwoolee: The rationale for this was `World->getSkeleton(std::size_t i)` method. If `getSkeleton(i)` is used to get a skeleton and do something, it'd be better if the user can assume that `worlds` with `equalConfiguration` have the same ordering of skeletons.

gilwoolee:
On the second thought, I think what we need is not just comparing two worlds for equal configuration but a "World::Configuration" struct that stores the current configuration of the world, so that a user(e.g. `magi`)  can store the configuration of the world at the time of planning. It will also be helpful in serialiation/deserialization. Any thoughts @brianhou @jslee02 @mkoval ?

mxgrey:
Excuse me for jumping in on this (I occasionally glance at the PRs and issues on Aikido in case I see anything interesting), but at one point we had plans on creating a generalized `World::State` which would contain a collection of generalized `Skeleton::State` objects. The goal was to be able to take perfect snapshots of everything in the world in order to perfectly play back a simulation. I think the idea got somewhat sidelined due to uncertainty about small details like whether contact information should be preserved in the `World::State`.

All that being said, I think a simple `World::Configuration` struct would be really valuable and more straightforward to design than a `World::State` class, and would be appropriate to have as a built-in feature of DART if that's something you're willing to push upstream. Some day I'd still like to see a `World::State`, but `World::Configuration` would be a good (and simpler) utility.

gilwoolee:
@mxgrey Your feedback is always more than welcome! :)

What we need is conceptually similar to `World::State`, but in a way that planners need. The main difference between `dart::simulation::World` and `aikido::planner::World` is that in the latter we don't take a `step` with during which objects undergo physics-based simulation. Instead, we kinematically move the skeletons as given by trajectories, and planners may check collisions as defined by `aikido::constraint::CollisionFree`. @brianhou and I were discussing whether these constraints (or collision groups used by these constraints) should be part of each World.

Essentially, what we need from `World::State` is everything we need to from the World to playback plan, postprocess, and execution of trajectories, given that we store planning information (e.g. which planner, which planning method, etc.) as a metadata in some other form. 

In the current implementation of aikido, our World is independent of DART's World, so I think it's easier to implement `World::State` in aikido.

---
In the long run though, I think we might want to have an option of integrating `dart::simulation::World` and being able to take a `step`, e.g. for kinodynamic push-planning, in which case we might want to push this discussion into DART as well.

brianhou:
We decided to create our own `aikido::planner::World` because we didn't need everything that `dart::simulation::World` provides in terms of simulation state. The hope is that having a `World` in `aikido` allows us to prototype more quickly, and once we have a better idea of what we need (and what others might find useful) we could perhaps move it to DART as a `KinematicWorld`.

gilwoolee:
Closing this. Please see #265 .

PR #264, opened 2017-11-22
Feature/RcnnPose

yskim041:
A new perception class for R-CNN + Moped.

***


src/perception/RcnnMopedModule.cpp
```diff
@@ -0,0 +1,163 @@
...
+  , mCanUri(std::move(canUri))
+  , mReferenceLink(std::move(referenceLink))
+  , mTfListener(mNodeHandle)
+{
+  // Do nothing
+}
+
+
+//=============================================================================
+const dart::dynamics::SkeletonPtr makeBodyFromURDF(const std::string& uri,
```
brianhou: We have this helper function in a few places already, so perhaps we should create `aikido::io::util` and add this method there. When we do that, we should add a `resourceRetriever` argument (so we don't have the overhead of initializing the `CatkinResourceRetriever` multiple times) like here: https://github.com/personalrobotics/herb3_demos/blob/master/soda_handoff/src/soda_handoff.cpp#L36

src/perception/RcnnMopedModule.cpp
```diff
@@ -0,0 +1,163 @@
...
+        aikido::perception::convertROSPoseToEigen(marker_transform.pose);
+    // Get the transform as a pose matrix
+    Eigen::Isometry3d frame_pose =
+        aikido::perception::convertStampedTransformToEigen(transform);
+    // Compose to get actual skeleton pose
+    Eigen::Isometry3d skel_pose = frame_pose * marker_pose;
+    Eigen::Isometry3d link_offset = mReferenceLink->getWorldTransform();
+    skel_pose = link_offset * skel_pose;
+
+    std::string skel_name = marker_transform.text;
```
brianhou: I think this should be in a YAML database or something.

src/perception/RcnnMopedModule.cpp
```diff
@@ -0,0 +1,163 @@
...
+    else
+    {
+      skel_uri = "package://pr_ordata/data/objects/can.urdf";
+    }
+
+    dart::dynamics::SkeletonPtr objSkeletonPtr =
+        makeBodyFromURDF(skel_uri, skel_pose);
+    objSkeletonPtr->setName(skel_name);
+
+    dart::dynamics::SkeletonPtr env_skeleton = env->getSkeleton(skel_name);
```
brianhou: If there are multiple objects with the same `skel_name`, how do we disambiguate between them? Or does MOPED-lite ensure that each name is unique? How does it know that one Coke can is `can_coke(1)` vs `can_coke(2)`?

src/perception/RcnnMopedModule.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+    if (env_skeleton != nullptr)
+    {
+      env->removeSkeleton(env_skeleton);
```
brianhou: This might be related to my question above, but why are we removing this skeleton? It seems that these skeletons would be from a previous call to `detectObjects`, but what if MOPED-lite detected fewer cans than the last call? It seems like the extra cans wouldn't be removed.

include/aikido/perception/RcnnMopedModule.hpp
```diff
@@ -0,0 +1,50 @@
...
+namespace perception {
+
+class RcnnMopedModule : public PerceptionModule
+{
+public:
+  RcnnMopedModule(
+      ros::NodeHandle nodeHandle,
+      std::string markerTopic,
+      dart::common::ResourceRetrieverPtr resourceRetriever,
+      std::string canUri,
```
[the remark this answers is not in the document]
brianhou: Actually, it seems that this parameter / corresponding field aren't used at all. We should remove this.

src/perception/RcnnMopedModule.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+//=============================================================================
+bool RcnnMopedModule::detectObjects(
+    const aikido::planner::WorldPtr& env,
+    ros::Duration timeout,
+    ros::Time timestamp)
+{
+  bool any_valid = false;
+  
+  tf::TransformListener listener(mNodeHandle);
```
brianhou: Is this used?

include/aikido/perception/RcnnMopedModule.hpp
```diff
@@ -0,0 +1,50 @@
...
+
+  virtual ~RcnnMopedModule() = default;
+
+  // Documentation inherited
+  bool detectObjects(
+      const aikido::planner::WorldPtr& env,
+      ros::Duration timeout = ros::Duration(0.0),
+      ros::Time timestamp = ros::Time(0.0)) override;
+
+private:
```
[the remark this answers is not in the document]
yskim041: added documentation

include/aikido/perception/RcnnMopedModule.hpp
```diff
@@ -0,0 +1,50 @@
...
+#include <tf/transform_listener.h>
+#include <aikido/perception/PerceptionModule.hpp>
+
+namespace aikido {
+namespace perception {
+
+class RcnnMopedModule : public PerceptionModule
+{
+public:
+  RcnnMopedModule(
```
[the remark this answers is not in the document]
yskim041: added documentation

include/aikido/perception/RcnnMopedModule.hpp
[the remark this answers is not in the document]
yskim041: I added `ObjectDatabase` which is similar to `YamlAprilTagsDatabase`. I generalized this part by using `ObjectDatabase`, and made a corresponding JSON file in personalrobotics/pr-ordata#28 .

src/perception/RcnnMopedModule.cpp
yskim041: Removed this part. I used dart::utils::DartLoader instead.

src/perception/RcnnMopedModule.cpp
yskim041: Removed this TransformListener.

src/perception/RcnnMopedModule.cpp
yskim041: I made this part relies on `obj_id` provided from `RcnnPose` module. I will make `RcnnPose` handling all naming conflicts so that this receiver does not need to take care of the object detection.

src/perception/RcnnMopedModule.cpp
yskim041: I also think that it will be better if the perception module decides everything including removing a disappeared object. Now this `RcnnMopedModule` will remove a skeleton with `obj_id` only if it gets "None" in `obj_key`.

yskim041:
I addressed all issues. I want to generalize this module as a common receiver for any kinds of rcnn+<pose detection> algorithms I will implement separately.
After getting approvals from reviewers, I will change the file names from `RcnnMoped*` to `RcnnPose*`.

include/aikido/perception/ObjectDatabase.hpp
```diff
@@ -0,0 +1,57 @@
...
+  /// load configuration data from a JSON file at URI \c configDataURI.
+  /// \param[in] resourceRetriever The pointer to obtain the configuration file
+  /// \param[in] configDataURI The URI for the configuration information file
+  ObjectDatabase(
+      const dart::common::ResourceRetrieverPtr& resourceRetriever,
+      dart::common::Uri configDataURI);
+
+  virtual ~ObjectDatabase() = default;
+
+  bool getObjectByKey(
```
brianhou: Can you document this? In particular, it seems from the code that it should have `\param[out] obj_name` and `\param[out] obj_resource`.

include/aikido/perception/RcnnMopedModule.hpp
```diff
@@ -0,0 +1,80 @@
...
+
+namespace aikido {
+namespace perception {
+
+/// Instantiates the \c PerceptionModule specifically for RcnnPose.
+///
+/// It receives input from the RcnnPose running in a separate process,
+/// by subscribing to a visualization_msg::MarkerArray ROS topic published by
+/// the RcnnPose node. It uses a \c ObjectDatabase to resolve each marker to a
+/// \c dart::common::Uri, and updates the environment which is a list of \c
```
brianhou: "updates the environment which is an `aikido::planner::World`"

include/aikido/perception/RcnnMopedModule.hpp
```diff
@@ -0,0 +1,80 @@
...
+  ///   \param[in] resourceRetriever A CatkinResourceRetriever for resources
+  /// related to config files and models
+  ///	\param[in] referenceFrameId The desired reference frame for the
+  /// object pose
+  ///	\param[in] referenceLink A link on the robot with respect to which the
+  /// pose is transformed
+  RcnnPoseModule(
+      ros::NodeHandle nodeHandle,
+      std::string markerTopic,
+      std::shared_ptr<ObjectDatabase> configData,
```
brianhou: This parameter is missing documentation.

src/perception/ObjectDatabase.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+//==============================================================================
+ObjectDatabase::ObjectDatabase(
+    const dart::common::ResourceRetrieverPtr& resourceRetriever,
+    dart::common::Uri configDataURI)
+{
+  // Read the JSON file into string
+  if (!resourceRetriever)
+  {
+    throw std::invalid_argument("ResourceRetrieverPtr given is null!");
```
brianhou: I think `ResourceRetriever is null` matches the rest of the error messages in `aikido` better.

src/perception/ObjectDatabase.cpp
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  const dart::common::ResourcePtr resource
```
brianhou: I think this and the following section that puts the file in a string can be replaced by `resourceRetriever->readAll(configDataURI)`.

src/perception/ObjectDatabase.cpp
```diff
@@ -0,0 +1,85 @@
...
+  {
+    throw std::runtime_error(
+        std::string("Failed reading  URI - ") + configDataURI.toString());
+  }
+
+  // Load from string
+  mObjData = YAML::Load(content);
+}
+
+//==============================================================================
```
brianhou: It seems that this function will either throw an error or succeed. I think rather than returning `bool` (which isn't very helpful if we're throwing errors anyway), we can just make the return type `void`.

src/perception/ObjectDatabase.cpp
```diff
@@ -0,0 +1,85 @@
...
+bool ObjectDatabase::getObjectByKey(
+    const std::string& _obj_key,
+    std::string& obj_name,
+    dart::common::Uri& obj_resource)
+{
+  // Get name of object and pose for a given tag ID
+  YAML::Node obj_node = mObjData[_obj_key];
+  if (obj_node)
```
brianhou: I think we can restructure this as `if (!obj_node) ...` then just dedenting the block that is currently inside the `if`.

src/perception/RcnnMopedModule.cpp
```diff
@@ -0,0 +1,199 @@
...
+#include <Eigen/Geometry>
+#include <dart/utils/urdf/DartLoader.hpp>
+#include <ros/ros.h>
+#include <ros/topic.h>
+#include <visualization_msgs/Marker.h>
+#include <visualization_msgs/MarkerArray.h>
+#include <aikido/io/CatkinResourceRetriever.hpp>
+#include <aikido/perception/shape_conversions.hpp>
+
+using dart::dynamics::SkeletonPtr;
```
brianhou: It doesn't seem like this is being used.

src/perception/RcnnMopedModule.cpp
```diff
@@ -0,0 +1,199 @@
...
+}
+
+bool RcnnPoseModule::detectObjects(
+    const aikido::planner::WorldPtr& env,
+    ros::Duration timeout,
+    ros::Time timestamp)
+{
+  // Checks detected objects, looks up the database,
+  // and adds new skeletons to the world env
+  bool any_valid = false;
```
brianhou: Maybe `any_detected` would be a better name than `any_valid`.

src/perception/RcnnMopedModule.cpp
```diff
@@ -0,0 +1,199 @@
...
+    const auto& marker_stamp = marker_transform.header.stamp;
+    const auto& obj_key = marker_transform.text;
+    const auto& obj_id = marker_transform.ns;
+    const auto& detection_frame = marker_transform.header.frame_id;
+    if (!timestamp.isValid() || marker_stamp < timestamp)
+    {
+      continue;
+    }
+
+    // If obj_key is "None",
```
brianhou: When would `obj_key` be None? Does the RCNN keep track of that somehow?

src/perception/RcnnMopedModule.cpp
```diff
@@ -0,0 +1,199 @@
...
+        = aikido::perception::convertStampedTransformToEigen(transform);
+    // Compose to get actual skeleton pose
+    Eigen::Isometry3d obj_pose = frame_pose * marker_pose;
+    Eigen::Isometry3d link_offset = mReferenceLink->getWorldTransform();
+    obj_pose = link_offset * obj_pose;
+
+    bool is_new_obj;
+    dart::dynamics::SkeletonPtr obj_skeleton;
+
+    // Check if skel in skel_list
```
brianhou: `skel_list` -> `World`

src/perception/RcnnMopedModule.cpp
yskim041: `RcnnPose` will set `obj_key` None when an object disappears on its list.

include/aikido/perception/RcnnMopedModule.hpp
yskim041: added documentation for `configData`

src/perception/ObjectDatabase.cpp
yskim041: I agree with this point. I changed the return type to `void`, and removed all returns.

PR #265, opened 2017-11-23
Add World::State

gilwoolee:
This PR addresses #261 . Based on discussion on #263, this PR creates `WorldState` struct and provides `equals` method to compare with other `WorldState`s.

***


mxgrey:
Just to clarify some things about the semantics in DART:

`Skeleton::Configuration` follows the classical concept of a robot configuration, which is basically a glorified vector that describes only the relevant components of the robot state using as few parameters as possible. This is usually all that's needed for kinematic planning.

`Skeleton::State` contains a perfect and complete description of the state that the robot is in, making it suitable for perfect recreations of fully dynamic simulations. The `Skeleton::State` class is extensible, so if users add their own "Aspects" to the Skeleton or any of its BodyNodes or Joints, then the state information of those Aspects will also be captured by the `Skeleton::State`. The most concrete example of this can be seen in `SoftBodyNode`: The deformation of the soft body is captured in a `Skeleton::State`, but it is not captured in a `Skeleton::Configuration`.

Since the `World::State` implementation here is building off of `Skeleton::Configuration`, you might want to consider naming it `World::Configuration`. That would leave you with an opportunity to create a `World::State` later on which could build off of `Skeleton::State` and capture more than what this current implementation is able to.

gilwoolee:
Thanks for the comments @mxgrey .

Indeed, what you described is precisely why I chose to use `Skeleton::Configuration` instead of `Skeleton::State`. `Skeleton::Configuration` has everything we need from Skeleton for kinematic planning. I initially thought of calling ours as `World::Configuration`, but the reason why I decided to go with `World::State` is because, although the current version only contains skeleton configurations, `World::State` should contain everything we need for kinematic planning in `aikido`, which might include a set of collision groups(`dart::collision::CollisionGroup`) or collision constraints(`aikido::constraint::CollisionFree`), or joint limits, etc.

In fact, now might be a good time to discuss whether to put in collision groups or anything else, to make `World::State` contain everything we need for making planning calls. Any thoughts @mkoval @jslee02 @brianhou @mxgrey ?

Additionally, I think I'll add `WorldStateSaver`, just like `aikido::statespace::dart::MetaSkeletonStateSpaceSaver`, but it seems like once we have the former we might not need the latter too much. In general, shouldn't planners take `World` to restore after planning as well as `MetaSkeletonStateSpace` to plan with? Restoring `World` should include restoring `MetaSksletonState` .

mxgrey:
You may want to consider putting information like collision groups and constraints into a `World::Properties` data structure, assuming those are parameters that describe the scenario and don't change over time. That way, if you're recording or comparing states for a scenario, you can be a little leaner in terms of memory usage and the reduce the number of operations that are performed in a comparison.

Note that the `Skeleton::Properties` class records time-independent information like joint limits.

gilwoolee:
@brianhou @mkoval Could you review this?

include/aikido/planner/World.hpp
```diff
@@ -7,9 +7,22 @@
 namespace aikido {
 namespace planner {
 
+/// A Kinematic world that contains a set of skeletons.
+/// It is expected that worlds do not share the same skeletons.
 class World
 {
 public:
+  // Encapsulates the state of the World.
+  struct State
```
brianhou: Should `State` also have some notion of the `Skeleton`s it contains? Maybe the `Skeleton` names?

include/aikido/planner/World.hpp
```diff
@@ -7,9 +7,22 @@
...
+  {
+    std::vector<dart::dynamics::Skeleton::Configuration> configurations;
+
+    /// Returns true if two world states are the same.
```
brianhou: Is there any reason to use `equals` rather than `operator==`?

src/planner/World.cpp
```diff
@@ -151,5 +151,50 @@ std::mutex& World::getMutex() const
...
+
+  return state;
+}
+
+//==============================================================================
+void World::setState(const World::State& state)
+{
+  if (state.configurations.size() != mSkeletons.size())
+    throw std::invalid_argument(
+        "World::State and this World does not have the same number of "
```
brianhou: Nit: does -> do

src/planner/WorldStateSaver.cpp
```diff
@@ -0,0 +1,19 @@
+#include <aikido/planner/WorldStateSaver.hpp>
+
+namespace aikido {
+namespace planner {
+
+WorldStateSaver::WorldStateSaver(
+    WorldPtr world)
+  : mWorld{world}
```
brianhou: Would it make sense to initialize `mWorldState` here instead?

src/planner/WorldStateSaver.cpp
```diff
@@ -0,0 +1,19 @@
+#include <aikido/planner/WorldStateSaver.hpp>
+
+namespace aikido {
+namespace planner {
+
+WorldStateSaver::WorldStateSaver(
```
brianhou: Nit: can we also add basic tests for `WorldStateSaver`?

tests/planner/test_World.cpp
```diff
@@ -91,3 +92,48 @@ TEST_F(WorldTest, CloningPreservesSkeletonNamesAndConfigurations)
...
+  mWorld->addSkeleton(skel1);
+  skel1->createJointAndBodyNodePair<RevoluteJoint>();
+  skel1->setPosition(0, 0.5);
+
+  auto otherWorld = mWorld->clone();
+  otherWorld->getSkeleton(0)->setPosition(0, 1.0);
+
+  EXPECT_FALSE(mWorld->getState().equals(otherWorld->getState()));
+  EXPECT_FALSE(otherWorld->getState().equals(mWorld->getState()));
+}
```
brianhou: Nit: can we add a couple tests for `setState` to verify that it throws errors?

include/aikido/planner/World.hpp
mkoval: One option is to index `configurations` by `Skeleton` name, i.e. switch from an `std::vector` to an `std::unordered_map`.

include/aikido/planner/World.hpp
mkoval: I don't see any reason why we couldn't use `operator ==` here. If we define that, then we should also remember to define `operator !=`. 🙃 

Also, this likely should be flagged as `const`.

include/aikido/planner/WorldStateSaver.hpp
```diff
@@ -0,0 +1,38 @@
...
+/// RAII class to save and restore a World's state.
+class WorldStateSaver
+{
+public:
+  /// Construct a WorldStateSaver and save the current state of the
+  /// \c World. This state will be restored when
+  /// WorldStateSaver is destructed.
+  ///
+  /// \param _space WorldState to save/restore
+  explicit WorldStateSaver(WorldPtr world);
```
mkoval: We should be able to flag this argument as `const`.

Also, I don't see a strong reason to make this accept a smart pointer rather than a reference or raw pointer.

include/aikido/planner/WorldStateSaver.hpp
```diff
@@ -0,0 +1,38 @@
...
+
+  virtual ~WorldStateSaver();
+
+  // WorldStateSaver is uncopyable, must use std::move
```
mkoval: I don't see any harm in making this class copyable. What's the concern here?

src/planner/World.cpp
```diff
@@ -151,5 +151,50 @@ std::mutex& World::getMutex() const
...
+    return false;
+
+  for (size_t idx = 0; idx < configurations.size(); ++idx)
+  {
+    if (configurations[idx] != other.configurations[idx])
+      return false;
+  }
+
+  return true;
+}
```
mkoval: :neckbeard: I think this whole function could be replaced with:
```c++
return configurations == other.configurations;
``` 

Something is also a bit funny about this. Shouldn't we verify that the two `Configuration`s contain the same `Skeleton`s somehow?

src/planner/World.cpp
```diff
@@ -151,5 +151,50 @@ std::mutex& World::getMutex() const
...
+{
+  if (state.configurations.size() != mSkeletons.size())
+    throw std::invalid_argument(
+        "World::State and this World does not have the same number of "
+        "skeletons.");
+
+  for (size_t idx = 0; idx < state.configurations.size(); ++idx)
+  {
+    std::lock_guard<std::mutex> lock(mSkeletons[idx]->getMutex());
+    mSkeletons[idx]->setConfiguration(state.configurations[idx]);
```
mkoval: Just a thought, outside the scope of this pull request: At some point we may want to abandon using per-`Skeleton` locks and instead having a single lock for the entire `World`. This wasn't an option before because we didn't have a `World` object. 😉

src/planner/WorldStateSaver.cpp
gilwoolee: `WorldState` does not keep a pointer to a specific world, but when `WorldStateSaver` is destroyed, we typically want to restore the state of a `World` to a previous state. So I think we need to keep a pointer to the world?

src/planner/World.cpp
gilwoolee: We're not checking whether the skeletons are the same because the skeletons may be cloned ones. Should I check whether they have the same joint structure?

gilwoolee:
@brianhou could you give it a (hopefully) final pass? :)

gilwoolee:
@mkoval Thanks for the feedback! I agree that saving/restoring a smaller subset of the world which gets modified during planning is a better idea, which we're not currently doing in `magi`. I think it'd still be useful to be able to take a snapshot of the world's current state, for example, in serializing/deserializing, etc. We might want to support selective state saving/restoring in the near future, so I'll leave it to #276.

gilwoolee:
There also seems to be a bit of detailed comparison that needs to be done in order to say whether two skeletons have the same "state space", which is necessary when we are doing `==` operation or `setState` operation. I am leaving that for later and just comparing skeleton names for now.

mkoval:
I think comparing `Skeleton` names is fine for now. I agree that comparing `StateSpace`s is a giant can of worms - and I don't have a good solution in mind. Let's revisit that issue once we have some solid demos running in Aikido on HERB. Then we'll have some more context for the discussion.

src/planner/World.cpp
```diff
@@ -151,5 +152,52 @@ std::mutex& World::getMutex() const
...
+{
+  return !(*this == other);
+}
+
+//==============================================================================
+World::State World::getState() const
+{
+  World::State state;
+
+  for (auto skeleton : mSkeletons)
```
brianhou: I think this can be `const auto& skeleton`?

src/planner/World.cpp
```diff
@@ -151,5 +152,52 @@ std::mutex& World::getMutex() const
...
+  for (const auto& skeleton : mSkeletons)
+  {
+    auto name = skeleton->getName();
+    auto it = state.configurations.find(name);
+    if (it == state.configurations.end())
+      throw std::invalid_argument(
+          "Skeleton " + name + " does not exist in state.");
+  }
+
+  for (const auto& skeleton : mSkeletons)
```
brianhou: Could this loop be merged into the previous one?

src/planner/World.cpp
```diff
@@ -151,5 +152,52 @@ std::mutex& World::getMutex() const
...
+  {
+    std::lock_guard<std::mutex> lock(skeleton->getMutex());
+    skeleton->setConfiguration(state.configurations.at(skeleton->getName()));
```
brianhou: Can we use `state.configurations[skeleton->getName()]` instead of `at`? Since we've already checked that the keys should all exist, `at` seems unnecessary.

brianhou: Actually, maybe we could merge the two loops and then just use `at` instead of explicitly calling `find` and then indexing? @jslee02 is it inefficient to catch the (possible) `out_of_range` exception from `at` and then trigger our own more helpful exception?

src/planner/World.cpp
gilwoolee: I didn't want to merge these two because I don't want this to throw an error after modifying some of the skeletons.

src/planner/World.cpp
gilwoolee: Hmm. Can't do `[]` because `[]` is a non-const operator and I'm taking in `const State&`. 
https://stackoverflow.com/questions/15660838/error-no-match-for-operator-in-near-match

I'll keep it as is.

Issue #237, opened 2017-10-10
Add optional output parameter to Testable::isSatisfied


brianhou:
This would give us a programmatic way to determine why Testables aren't satisfied. This is helpful for multi-step planning algorithms, as well as general debugging.

@mkoval has a sketch of an implementation here: https://github.com/personalrobotics/aikido/pull/235#pullrequestreview-68305028

brianhou:
Resolved by #266.

PR #266, opened 2017-11-26
Testable Constraint Output

sniyaz:
See https://github.com/personalrobotics/aikido/issues/237. 

**Summary of changes:**

1) New TestableOutcome class and corresponding CollisionFreeOutcome derivative class for CollisionFree (NonColliding was renamed to CollisionFree in #256).

2) Change base constraint class (Testable) to add TestableOutcome (base constraint outcome class) as parameter to isSatisfied.

3) CollisionFree constraint now populates this field on a collision (self or otherwise).

**TODO**:
@mkoval, l I had a question about https://github.com/personalrobotics/aikido/pull/235, item 4. This is something I'd be happy to implement before merging, but I was curious what sort of use case this scenario would pop up in.

**Test Plan**

Also checked CollisionFreeOutcome's toString and isSatisfied methods in test_CollisionFree unit tests. Printed the return value of toString separately, and result seemed sane.

***


jslee02:
I added [the checklist](https://github.com/personalrobotics/aikido/blob/master/.github/PULL_REQUEST_TEMPLATE.md) that is for our PRs.

src/constraint/CollisionFree.cpp
```diff
@@ -42,16 +43,39 @@ bool CollisionFree::isSatisfied(
         groups.second.get(),
         mCollisionOptions,
         &collisionResult);
+
     if (collision)
+    {
+      if (_outcome != nullptr)
+      {
+        auto collisionOutcome = static_cast<CollisionFreeOutcome*>(_outcome);
+        auto collidingBodies = collisionResult.getCollidingBodyNodes();
```
jslee02: Nit: `const auto&` to avoid the unnecessary copy.

src/constraint/CollisionFree.cpp
```diff
@@ -42,16 +43,39 @@ bool CollisionFree::isSatisfied(
         groups.second.get(),
         mCollisionOptions,
         &collisionResult);
+
     if (collision)
+    {
+      if (_outcome != nullptr)
+      {
+        auto collisionOutcome = static_cast<CollisionFreeOutcome*>(_outcome);
```
jslee02: It's not guaranteed if `_outcome` is the type of `CollisionFreeOutcome*`. I think we should do one of the follows:
(1) use assertion (in debug mode) and `dynamic_cast` to check if `_outcome` is the right type. The downside would be that we still have problem in release mode
(2) run the routine only when it succeded to cast to the right type using `dynamic_cast` (warnings if the type is not the right one). The downside would be the overhaed of using `dynamic_cast`.
(3) throw exception when the type is not right. The downside would be that it has to use `dynamic_cast` anyway.

I'm not sure what would be the preferred way here. @brianhou , @mkoval, @gilwoolee thoughts?

jslee02:
@mkoval I also have a similar question about item 4 in [the comment](https://github.com/personalrobotics/aikido/pull/235#pullrequestreview-68305028).

In the case that we don't know the type of `Testable` at compile time, I think we also have a problem of not knowing what the outcome type should be. For this, I would suggest adding a virtual function `std::unique_ptr<Outcome> createOutcome()` to `Testable` so that we could create a right type of outcome from the testable like:
```c++
auto outcome = testable->createOutcome();
testable->isSatisfied(state, outcome);
std::cout << outcome->toString() << std::endl;
```

jslee02:
(sorry for not reviewing in multiple comments)

Also, I'm not sure if we really want to put `Outcome` class in a nested directory. Do we have a reason for putting the class in the nested directory?

src/constraint/CollisionFree.cpp
brianhou: Is the only difference between (2) and (3) whether we silently fail or raise an exception when the `TestableOutcome` is the wrong type? In that case, I think I'd prefer (3) over (2) to make sure we know that we've made a mistake.

My (limited) understanding is that we won't be passing a non-null `TestableOutcome` that often, so I'm also not as concerned about the overhead of using a `dynamic_cast` in release mode. Perhaps we can use (3) now and if it becomes performance-critical we can re-evaluate and perhaps switch to (1).

include/aikido/constraint/outcome/CollisionFreeOutcome.hpp
```diff
@@ -0,0 +1,27 @@
+#ifndef AIKIDO_CONSTRAINT_COLLISIONFREEOUTCOME_HPP_
+#define AIKIDO_CONSTRAINT_COLLISIONFREEOUTCOME_HPP_
+
+#include <sstream>
```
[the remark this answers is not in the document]
sniyaz: Could you elaborate on this a bit? If <sstream> is not included, then it looks like the string building in toString won't work (since a string stream is used there, and then converted to a string object).

sniyaz:
Awesome feedback! Working through this now.

src/constraint/CollisionFree.cpp
```diff
@@ -42,16 +43,39 @@ bool CollisionFree::isSatisfied(
...
+        auto collidingBodies = collisionResult.getCollidingBodyNodes();
+        for (const auto& elem : collidingBodies)
```
brianhou: Maybe this should iterate over `CollisionResult::getContacts()` instead. That will allow us to specify which _pairs_ of bodies are in collision, rather than squashing that information into a list of bodies that are in collision.

This isn't a problem now because our default `CollisionOption` only detects a maximum of one contact between two bodies, but we may as well support multiple contacts on the off-chance that becomes helpful.

tests/constraint/test_CollisionFree.cpp
```diff
@@ -64,6 +65,11 @@ class CollisionFreeTest : public ::testing::Test
     mStateSpace = std::make_shared<MetaSkeletonStateSpace>(group);
     mManipulator->enableSelfCollisionCheck();
     mManipulator->enableAdjacentBodyCheck();
+
+    // See what CollisionFreeOutcome's toString returns when no collision
```
brianhou: We should take advantage of having programmatic access to the information (rather than as a giant string). Maybe we can instead have a vector of body names that we check the outcome's vectors against.

src/constraint/CartesianProductTestable.cpp
```diff
@@ -53,7 +53,8 @@ statespace::StateSpacePtr CartesianProductTestable::getStateSpace() const
 
 //==============================================================================
 bool CartesianProductTestable::isSatisfied(
-    const aikido::statespace::StateSpace::State* _state) const
+    const aikido::statespace::StateSpace::State* _state,
+    TestableOutcome* _outcome) const
```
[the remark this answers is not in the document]
sniyaz: Thanks! 😄 This is a super handy thing to know.

include/aikido/constraint/outcome/CollisionFreeOutcome.hpp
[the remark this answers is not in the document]
jslee02: `sstream` is not exposed in this header file but used in the source file. So I suggest we include `<string>` as `std::string` is explicitly exposed in this header and include `<sstream>` in the source file where it's actually used. 

The motivation behind this is that this way prevents unnecessarily including `<sstream>` by including `CollisionFreeOutcome.hpp`.

src/constraint/CollisionFree.cpp
sniyaz: I agree with this logic (assuming that @brianhou's reasoning about (2) and (3) is correct). I'm going to with (3) right now, and we can change if people have strong feelings towards either approach.

sniyaz:
As of the latest commit, there are two big changes:

1) As per the suggestion of @brianhou, I have made CollisionFreeOutcome operate on Contact objects instead of just BodyNode names (for all of the advantages Brian gave, including seeing the exact collision pairs). This involved writing a method `getCollisionObjectName` to go from CollisionObject to a name. I'd appreciate feedback on this function in particular, since I'm not 100% sure this is the best way to implement that functionality.

2) CollisionFree will now barf and throw an exception when the TestableOutcome pointer passed to isSatisfied doesn't point to a CollisionFreeOutcome object.

I will go through soon and update the tests as @brianhou mentioned, but I'd like to have these changes rubber stamped before then.

sniyaz:
Updated tests to take advantage of programmatic access to data.

include/aikido/constraint/outcome/CollisionFreeOutcome.hpp
```diff
@@ -0,0 +1,62 @@
...
+#include "dart/collision/CollisionObject.hpp"
+#include "dart/collision/Contact.hpp"
+#include "dart/dynamics/BodyNode.hpp"
+#include "dart/dynamics/ShapeFrame.hpp"
+#include "dart/dynamics/ShapeNode.hpp"
+
+namespace aikido {
+namespace constraint {
+
+class CollisionFreeOutcome : public TestableOutcome
```
jslee02: Could we add a simple docstring for this class like saying this class is a testable outcome for `CollisionFree` class?

include/aikido/constraint/outcome/CollisionFreeOutcome.hpp
```diff
@@ -0,0 +1,62 @@
...
+  std::vector<dart::collision::Contact> mPairwiseContacts;
+
+  /// Holds Contact objects from self collisions.
+  std::vector<dart::collision::Contact> mSelfContacts;
+};
+
+} // namespace constraint
+} // namespace aikido
+
+#endif // ifndef AIKIDO_CONSTRAINT_COLLISIONFREEOUTCOME_HPP_
```
jslee02: Nit: `#endif // AIKIDO_CONSTRAINT_COLLISIONFREEOUTCOME_HPP_`

include/aikido/constraint/outcome/TestableOutcome.hpp
```diff
@@ -0,0 +1,24 @@
+#ifndef AIKIDO_CONSTRAINT_TESTABLEOUTCOME_HPP_
+#define AIKIDO_CONSTRAINT_TESTABLEOUTCOME_HPP_
+
+#include <string>
+
+namespace aikido {
+namespace constraint {
+
+class TestableOutcome
```
jslee02: Could we add docstring for this class?

src/constraint/CollisionFree.cpp
```diff
@@ -42,16 +43,47 @@ bool CollisionFree::isSatisfied(
...
+        for (const auto& elem : contacts)
+        {
+          collisionOutcome->markPairwiseContact(elem);
+        }
+      }
       return false;
+    }
   }
 
   for (auto group : mGroupsToSelfCheck)
```
jslee02: Nit: This is irrelevant to this PR, but let's avoid unnecessary copy here by changing this line to `for (const auto& group : mGroupsToSelfCheck)`.

Also `for (auto groups : mGroupsToPairwiseCheck)` should be `for (const auto& groups : mGroupsToPairwiseCheck)`

src/constraint/CollisionFree.cpp
```diff
@@ -42,16 +43,47 @@ bool CollisionFree::isSatisfied(
         groups.second.get(),
         mCollisionOptions,
         &collisionResult);
+
     if (collision)
+    {
+      if (_outcome != nullptr)
+      {
+        const auto& collisionOutcome
+            = dynamic_cast<CollisionFreeOutcome*>(_outcome);
```
jslee02: Nit: `dynamic_cast` is known to be expensive (relatively to `static_cast`) so it's preferable to avoid or use less if possible. According to this logic, dynamic casting will be called multiple times. I think this can be optimized by performing the dynamic casting once like:
```c++
CollisionFreeOutcome* collisionOutcome = nullptr;
  if (_outcome)
  {
    collisionOutcome = dynamic_cast<CollisionFreeOutcome*>(_outcome);
    if (!collisionOutcome)
    {
      throw std::invalid_argument(
          "TestableOutcome pointer is not of type CollisionFreeOutcome.");
    }
  }

  for (const auto& groups : mGroupsToPairwiseCheck)
  {
    collision = mCollisionDetector->collide(
        groups.first.get(),
        groups.second.get(),
        mCollisionOptions,
        &collisionResult);

    if (collision)
    {
      if (collisionOutcome)
      {
        const auto& contacts = collisionResult.getContacts();
        for (const auto& elem : contacts)
          collisionOutcome->markPairwiseContact(elem);
      }

      return false;
    }
  }

  for (const auto& group : mGroupsToSelfCheck)
  {
    collision = mCollisionDetector->collide(
        group.get(), mCollisionOptions, &collisionResult);
    
    if (collision)
    {
        if (collisionOutcome)
        {
          const auto& contacts = collisionResult.getContacts();
          for (const auto& elem : contacts)
            collisionOutcome->markSelfContact(elem);
        }

      return false;
    }
  }

  return true;
```

src/constraint/outcome/CollisionFreeOutcome.cpp
```diff
@@ -0,0 +1,82 @@
+#include <sstream>
+#include <aikido/constraint/outcome/CollisionFreeOutcome.hpp>
```
jslee02: Super nit: In a source file, we include the associated header very first then others in order of STL headers, 3rd party library headers, and AIKIDO headers.

include/aikido/constraint/outcome/CollisionFreeOutcome.hpp
```diff
@@ -0,0 +1,62 @@
...
+  bool isSatisfied() const override;
+
+  /// Returns a string with each pair of CollisionObject names on a separate
+  /// line. Each pair is also marked as being a normal collision or self
+  /// collision.
+  std::string toString() const override;
+
+  /// Store a Contact object from a pairwise collision. Used by CollisionFree
+  // to modify this object.
+  /// \param pairwiseContact Contact object to store.
```
jslee02: Nit: We prefer mark whether the parameter is input param or output param using `[in]` and `[out]` affixes to `\param`. Please change this to `/// \param[in] pairwiseContact` (also other parameters in this header).

include/aikido/constraint/outcome/CollisionFreeOutcome.hpp
```diff
@@ -0,0 +1,62 @@
...
+  /// Documentation inherited.
+  bool isSatisfied() const override;
+
+  /// Returns a string with each pair of CollisionObject names on a separate
+  /// line. Each pair is also marked as being a normal collision or self
+  /// collision.
+  std::string toString() const override;
+
+  /// Store a Contact object from a pairwise collision. Used by CollisionFree
+  // to modify this object.
```
jslee02: Nit: `/// to modify this object.` (need one more `/`)

include/aikido/constraint/outcome/CollisionFreeOutcome.hpp
```diff
@@ -0,0 +1,62 @@
...
+  /// Return a copy of the vector storing the Contact objects from self
+  /// collisions.
+  std::vector<dart::collision::Contact> getSelfContacts() const;
+
+  /// Gets the name of a CollisionObject. The name returned is that of the
+  /// corresponding BodyNode (if possible). If not, the name of the ShapeFrame
+  /// is returned instead. This is a helper for toString().
+  /// \param object object pointer to CollisionObject we want the name of.
+  std::string getCollisionObjectName(
+      dart::collision::CollisionObject* object) const;
```
jslee02: Nit: It seems `object` is not changed by this function. It would be nice to change the type to `const` so that we could pass const collision object as well.

src/constraint/CollisionFree.cpp
brianhou: I initially had the same thought, but I believe that since we immediately `return false` when there is a pairwise collision, there is no chance of `dynamic_cast`ing a second time.

jslee02: What if `outcome` is actually a pointer of `CollisionFreeOutcome`? I think in that case, `dynamic_cast` will be called multiple times for every iteration of the pairwise collisions.

src/constraint/CollisionFree.cpp
```diff
@@ -27,31 +27,63 @@ statespace::StateSpacePtr CollisionFree::getStateSpace() const
...
         groups.first.get(),
         groups.second.get(),
         mCollisionOptions,
         &collisionResult);
+
     if (collision)
+    {
+      if (outcome != nullptr)
+      {
+        const auto& collisionOutcome
```
brianhou: I don't think this should be `const`, since we're changing it below. I'm also unclear as to whether it should be `auto&` or just `auto` -- [this](https://stackoverflow.com/a/29860056) seems to suggest that `auto` is fine?

src/constraint/CollisionFree.cpp
```diff
@@ -27,31 +27,63 @@ statespace::StateSpacePtr CollisionFree::getStateSpace() const
...
-  for (auto group : mGroupsToSelfCheck)
+  for (const auto& group : mGroupsToSelfCheck)
   {
     collision = mCollisionDetector->collide(
         group.get(), mCollisionOptions, &collisionResult);
     if (collision)
+    {
+      if (outcome != nullptr)
+      {
+        const auto& collisionOutcome
```
brianhou: I think this should just be `auto collisionOutcome`, for the same reason as above.

sniyaz:
Implemented the `testable->createOutcome()` functionality as requested by @jslee02. This invovled writing a default implementation in Testable and an overriding one in CollisionFreeOutcome. Also created a DummyOutcome object (not intended to actually be used) that is used by the default createOutcome method in Testable.

Note: DummyOutcome throws an exception if `isSatisfied` is called to make it **very clear** to the user that the returned object is not meant for use. We can remove this if people would like to.

include/aikido/constraint/outcome/TestableOutcome.hpp
```diff
@@ -0,0 +1,24 @@
+#ifndef AIKIDO_CONSTRAINT_TESTABLEOUTCOME_HPP_
```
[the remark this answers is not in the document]
sniyaz: That makes sense. My main reasoning here was that all of the things in the main directory seem to be constraints, and these new files aren't constraints- they're supposed to correspond to the outputs of constraints. 

If nobody else has strong feelings either way, I'll move the files into the main constraint directory before merging.

include/aikido/constraint/FrameTestable.hpp
```diff
@@ -29,7 +29,13 @@ class FrameTestable : public Testable
   /// \param _state a MetaskeletonState to set the configuration of this
   ///        constraint's metaskeketon. This state's StateSpace should match
   ///        StateSpace returend by getStateSpace().
```
mkoval: Document the `TestableOutcome` argument. I suggest making it simply return whatever `TestableOutcome` is returned by the `_poseConstraint` passed to this class' constructor.

include/aikido/constraint/FrameTestable.hpp
```diff
@@ -29,7 +29,13 @@ class FrameTestable : public Testable
...
-  bool isSatisfied(const statespace::StateSpace::State* _state) const override;
+  bool isSatisfied(
+      const statespace::StateSpace::State* _state,
+      TestableOutcome* outcome = nullptr) const override;
+
+  /// Return an instance of DefaultOutcome, since this class doesn't have a
+  /// more specialized TestableOutcome derivative assigned to it.
+  std::unique_ptr<TestableOutcome> createOutcome() const override;
```
mkoval: Instead of this logic, I suggest returning the value of `_poseConstraint->createOutcome()` to preserve any useful debugging information that may contain. This should be strictly more informative than returning a `DefaultOutcome`. See my comment on `isSatisfied()` for details.

At some point in the future, we may need to wrap this outcome in a `FrameTestable` to, e.g. include some context about the joint configuration that was used to perform the check. But that can come later. 🙃

include/aikido/constraint/Testable.hpp
```diff
@@ -15,15 +18,21 @@ class Testable
 
   /// Returns true if state satisfies this constraint.
```
mkoval: Document the `TestableOutcome` parameter. Most importantly: (1) what type it must be and (2) what happens if it is not specified. 😉

include/aikido/constraint/DefaultOutcome.hpp
```diff
@@ -0,0 +1,35 @@
...
+#include "TestableOutcome.hpp"
+
+namespace aikido {
+namespace constraint {
+
+/// Simple default TestableOutcome derivative class. An instance of this class
+/// is returned when createOutcome() is called on an instance of a class that
+/// inherits Testable, but has no corresponding TestableOutcome derivative
+/// implemented.
+class DefaultOutcome : public TestableOutcome
```
mkoval: I suggest renaming this to `DefaultTestableOutcome`, since it inherits from `TestableOutcome` and is currently tied to `Testable` constraints. It's unclear if we'll be able to use the same API on other constraint types.

include/aikido/constraint/uniform/detail/RnBoxConstraint-impl.hpp
```diff
@@ -175,19 +175,40 @@ std::vector<ConstraintType> RBoxConstraint<N>::getConstraintTypes() const
...
+  }
+
   const auto value = mSpace->getValue(
       static_cast<const typename statespace::R<N>::State*>(state));
 
   for (auto i = 0; i < value.size(); ++i)
   {
     if (value[i] < mLowerLimits[i] || value[i] > mUpperLimits[i])
+    {
+      defaultOutcomeObject->setSatisfiedFlag(false);
```
mkoval: This will SEGFAULT if `defaultOutcomeObject` is `nullptr`.

include/aikido/constraint/uniform/detail/RnBoxConstraint-impl.hpp
```diff
@@ -175,19 +175,40 @@ std::vector<ConstraintType> RBoxConstraint<N>::getConstraintTypes() const
...
       return false;
+    }
   }
+
+  defaultOutcomeObject->setSatisfiedFlag(true);
```
mkoval: This will SEGFAULT if `defaultOutcomeObject` is `nullptr`. See my comment above.

include/aikido/constraint/uniform/detail/RnBoxConstraint-impl.hpp
```diff
@@ -175,19 +175,40 @@ std::vector<ConstraintType> RBoxConstraint<N>::getConstraintTypes() const
...
   return true;
 }
 
+//==============================================================================
+template <int N>
+std::unique_ptr<TestableOutcome> RBoxConstraint<N>::createOutcome() const
+{
+  return std::unique_ptr<TestableOutcome>(new DefaultOutcome());
```
mkoval: Nit: `new DefaultOutcome()` -> `new DefaultOutcome`.

include/aikido/constraint/uniform/detail/RnBoxConstraint-impl.hpp
```diff
@@ -175,19 +175,40 @@ std::vector<ConstraintType> RBoxConstraint<N>::getConstraintTypes() const
...
+    const statespace::StateSpace::State* state, TestableOutcome* outcome) const
 {
+  DefaultOutcome* defaultOutcomeObject = nullptr;
+  if (outcome)
+  {
+    defaultOutcomeObject = dynamic_cast<DefaultOutcome*>(outcome);
+    if (!defaultOutcomeObject)
+      throw std::invalid_argument(
+          "TestableOutcome pointer is not of type DefaultOutcome.");
+  }
```
mkoval: Nit: This logic is duplicated in quite a few places. I suggest putting it in a helper function, e.g.:
```c++
template <class Child, class Parent>
Child* checked_dynamic_cast<Child, Parent>(Parent* _parent)
{
  auto child = dynamic_cast<Child*>(_parent);
  if (!child && _parent)
  {
    std::stringstream message;
    message << "'" << std::type_info<Parent>::name() << "' is not of type "'"
            << std::type_info<Child>::name() <<"'.";
    throw std::invalid_argument{
  }
}
```
Then, you could reduce all of this boilerplate to just:
```c++
auto defaultOutcomeObject = checked_dynamic_cast<DefaultOutcome>(outcome);
```
(Some massaging may be required for the template type inference to work.)

Note that if you make `clear()` part of the base `TestableOutcome` API, then you could also put `clear()` inside that helper function! 😍

src/constraint/CollisionFree.cpp
```diff
@@ -27,35 +27,74 @@ statespace::StateSpacePtr CollisionFree::getStateSpace() const
...
     if (collision)
+    {
+      if (collisionFreeOutcome)
+      {
+        const auto& contacts = collisionResult.getContacts();
+        for (const auto& elem : contacts)
+        {
+          collisionFreeOutcome->markSelfContact(elem);
+        }
+      }
```
mkoval: Nit: If you make `CollisionFreeOutcome` a `friend class`, then this could be reduced to:
```c++
collisionFreeOutcome->mSelfContacts = collisionResult.getContacts();
```

src/constraint/CollisionFree.cpp
```diff
@@ -27,35 +27,74 @@ statespace::StateSpacePtr CollisionFree::getStateSpace() const
...
+
     if (collision)
+    {
+      if (collisionFreeOutcome)
+      {
+        const auto& contacts = collisionResult.getContacts();
+        for (const auto& elem : contacts)
+        {
+          collisionFreeOutcome->markPairwiseContact(elem);
+        }
```
mkoval: Nit: If you make `CollisionFreeOutcome` a `friend class`, then this could be reduced to:
```c++
collisionFreeOutcome->mContacts = collisionResult.getContacts();
```

sniyaz: Thanks for all the awesome feedback! :smile: I had a question about this one- when using  Eigen::aligned_allocator (as mentioned in a previous comment), wouldn't this be not possible? The two vectors appear to use different allocators.

sniyaz:
Pretty much there. Two biggest changes:

1. Made helper method to cast between types of TestableOutcome.

2. Make every Testable derivative class populate the isSatisfied output. Made every Testable derivative also implement createOutcome.

Various other tweaks as well (see comments). I think this should be ready to merge soon, so @jslee02 @brianhou @mkoval, let me know if you have any other thoughts.

src/constraint/CollisionFree.cpp
mkoval: I suspect this is actually a bug in DART! :grimacing: @mxgrey @jslee02 

In any case, you can still do the assignment with a bit more boilerplate:
```c++
const auto& contacts = collisionResult.getContacts();
collisionFreeOutcome.assign(std::begin(contacts), std::end(contacts));
```

include/aikido/constraint/CollisionFree.hpp
```diff
@@ -39,9 +40,16 @@ class CollisionFree : public Testable
   // Documentation inherited.
   statespace::StateSpacePtr getStateSpace() const override;
 
-  // Documentation inherited.
+  // Documentation inherited. As an important note, outcome is expected to be
```
jslee02: Nit: It would be great to expose this note by making this doxygen comment as:
```c++
/// \copydoc Testable::isSatisfied()
/// \note Outcome is expected to be an instance of ...
```
By doing above, we could see this note in addition to the original docstring of `Testable::isSatisfied()`.

include/aikido/constraint/CollisionFree.hpp
```diff
@@ -39,9 +40,16 @@ class CollisionFree : public Testable
...
+  // an instance of CollisionFreeOutcome. This method will cast outcome to
+  // a pointer of this type, and then populate the collision information
+  // (which bodies are in self/pairwise collision).
   bool isSatisfied(
-      const aikido::statespace::StateSpace::State* _state) const override;
+      const aikido::statespace::StateSpace::State* _state,
+      TestableOutcome* outcome = nullptr) const override;
+
+  // Documentation inherited. Returns an instance of CollisionFreeOutcome.
```
jslee02: Nit: ditto
```c++
/// \copydoc Testable::createOutcome()
/// Returns an instance of CollisionFreeOutcome.
```

include/aikido/constraint/uniform/detail/RnBoxConstraint-impl.hpp
```diff
@@ -175,19 +175,36 @@ std::vector<ConstraintType> RBoxConstraint<N>::getConstraintTypes() const
...
+      = dynamic_cast_if_present<DefaultTestableOutcome>(outcome);
+
   const auto value = mSpace->getValue(
       static_cast<const typename statespace::R<N>::State*>(state));
 
   for (auto i = 0; i < value.size(); ++i)
   {
     if (value[i] < mLowerLimits[i] || value[i] > mUpperLimits[i])
+    {
+      if (defaultOutcomeObject)
```
jslee02: I don't think we need the nullity check here because `dynamic_cast_if_present` throws an exception if `dynamic_cast<*>` is `nullptr`.

include/aikido/constraint/uniform/detail/RnBoxConstraint-impl.hpp
```diff
@@ -175,19 +175,36 @@ std::vector<ConstraintType> RBoxConstraint<N>::getConstraintTypes() const
...
+        defaultOutcomeObject->setSatisfiedFlag(false);
       return false;
+    }
   }
+
+  if (defaultOutcomeObject)
```
jslee02: ditto:
> I don't think we need the nullity check here because `dynamic_cast_if_present` throws an exception if `dynamic_cast<*>` is `nullptr`.

src/constraint/CartesianProductTestable.cpp
```diff
@@ -53,19 +53,36 @@ statespace::StateSpacePtr CartesianProductTestable::getStateSpace() const
...
+
   const auto state
       = static_cast<const statespace::CartesianProduct::State*>(_state);
 
   for (std::size_t i = 0; i < mStateSpace->getNumSubspaces(); ++i)
   {
     auto subState = mStateSpace->getSubState<>(state, i);
     if (!mConstraints[i]->isSatisfied(subState))
+    {
+      if (defaultOutcomeObject)
```
jslee02: ditto:
> I don't think we need the nullity check here because `dynamic_cast_if_present` throws an exception if `dynamic_cast<*>` is `nullptr`.

src/constraint/CartesianProductTestable.cpp
```diff
@@ -53,19 +53,36 @@ statespace::StateSpacePtr CartesianProductTestable::getStateSpace() const
...
+        defaultOutcomeObject->setSatisfiedFlag(false);
       return false;
+    }
   }
+
+  if (defaultOutcomeObject)
```
jslee02: ditto:
> I don't think we need the nullity check here because `dynamic_cast_if_present` throws an exception if `dynamic_cast<*>` is `nullptr`.

src/constraint/CollisionFree.cpp
```diff
@@ -27,35 +27,71 @@ statespace::StateSpacePtr CollisionFree::getStateSpace() const
...
 //==============================================================================
 bool CollisionFree::isSatisfied(
-    const aikido::statespace::StateSpace::State* _state) const
+    const aikido::statespace::StateSpace::State* _state,
+    TestableOutcome* outcome) const
 {
+  auto collisionFreeOutcome
+      = dynamic_cast_if_present<CollisionFreeOutcome>(outcome);
+
+  if (collisionFreeOutcome)
```
jslee02: ditto:
> I don't think we need the nullity check here because `dynamic_cast_if_present` throws an exception if `dynamic_cast<*>` is `nullptr`.

src/constraint/CollisionFree.cpp
```diff
@@ -27,35 +27,71 @@ statespace::StateSpacePtr CollisionFree::getStateSpace() const
...
   {
     collision = mCollisionDetector->collide(
         groups.first.get(),
         groups.second.get(),
         mCollisionOptions,
         &collisionResult);
+
     if (collision)
+    {
+      if (collisionFreeOutcome)
```
jslee02: ditto:
> I don't think we need the nullity check here because `dynamic_cast_if_present` throws an exception if `dynamic_cast<*>` is `nullptr`.

src/constraint/CollisionFree.cpp
```diff
@@ -27,35 +27,71 @@ statespace::StateSpacePtr CollisionFree::getStateSpace() const
...
   }
 
-  for (auto group : mGroupsToSelfCheck)
+  for (const auto& group : mGroupsToSelfCheck)
   {
     collision = mCollisionDetector->collide(
         group.get(), mCollisionOptions, &collisionResult);
     if (collision)
+    {
+      if (collisionFreeOutcome)
```
jslee02: ditto:
> I don't think we need the nullity check here because `dynamic_cast_if_present` throws an exception if `dynamic_cast<*>` is `nullptr`.

src/constraint/Satisfied.cpp
```diff
@@ -31,11 +31,23 @@ std::vector<constraint::ConstraintType> Satisfied::getConstraintTypes() const
...
 //==============================================================================
 bool Satisfied::isSatisfied(
-    const statespace::StateSpace::State* /*state*/) const
+    const statespace::StateSpace::State* /*state*/,
+    TestableOutcome* outcome) const
 {
+  auto defaultOutcomeObject
+      = dynamic_cast_if_present<DefaultTestableOutcome>(outcome);
+
+  if (defaultOutcomeObject)
```
jslee02: ditto:
> I don't think we need the nullity check here because `dynamic_cast_if_present` throws an exception if `dynamic_cast<*>` is `nullptr`.

src/constraint/TSR.cpp
```diff
@@ -210,11 +210,25 @@ std::unique_ptr<SampleGenerator> TSR::createSampleGenerator() const
...
 {
+  auto defaultOutcomeObject
+      = dynamic_cast_if_present<DefaultTestableOutcome>(outcome);
+
   Eigen::VectorXd dist;
   getValue(_s, dist);
-  return dist.norm() < mTestableTolerance;
+
+  bool isSatisfiedResult = dist.norm() < mTestableTolerance;
+  if (defaultOutcomeObject)
```
jslee02: ditto:
> I don't think we need the nullity check here because `dynamic_cast_if_present` throws an exception if `dynamic_cast<*>` is `nullptr`.

src/constraint/TestableIntersection.cpp
```diff
@@ -20,16 +20,33 @@ TestableIntersection::TestableIntersection(
...
+    TestableOutcome* outcome) const
 {
+  auto defaultOutcomeObject
+      = dynamic_cast_if_present<DefaultTestableOutcome>(outcome);
+
   for (auto c : mConstraints)
   {
     if (!c->isSatisfied(_state))
+    {
+      if (defaultOutcomeObject)
```
jslee02: ditto:
> I don't think we need the nullity check here because `dynamic_cast_if_present` throws an exception if `dynamic_cast<*>` is `nullptr`.

src/constraint/TestableIntersection.cpp
```diff
@@ -20,16 +20,33 @@ TestableIntersection::TestableIntersection(
...
+        defaultOutcomeObject->setSatisfiedFlag(false);
       return false;
+    }
   }
+
+  if (defaultOutcomeObject)
```
jslee02: ditto:
> I don't think we need the nullity check here because `dynamic_cast_if_present` throws an exception if `dynamic_cast<*>` is `nullptr`.

src/constraint/uniform/SE2BoxConstraint.cpp
```diff
@@ -137,19 +137,35 @@ statespace::StateSpacePtr SE2BoxConstraint::getStateSpace() const
...
+      = dynamic_cast_if_present<DefaultTestableOutcome>(outcome);
+
   Eigen::VectorXd tangent;
   mSpace->logMap(static_cast<const statespace::SE2::State*>(state), tangent);
 
   for (std::size_t i = mDimension - mRnDimension; i < mDimension; ++i)
   {
     if (tangent[i] < mLowerLimits[i] || tangent[i] > mUpperLimits[i])
+    {
+      if (defaultOutcomeObject)
```
jslee02: ditto:
> I don't think we need the nullity check here because `dynamic_cast_if_present` throws an exception if `dynamic_cast<*>` is `nullptr`.

src/constraint/uniform/SE2BoxConstraint.cpp
```diff
@@ -137,19 +137,35 @@ statespace::StateSpacePtr SE2BoxConstraint::getStateSpace() const
...
+        defaultOutcomeObject->setSatisfiedFlag(false);
       return false;
+    }
   }
+
+  if (defaultOutcomeObject)
```
jslee02: ditto:
> I don't think we need the nullity check here because `dynamic_cast_if_present` throws an exception if `dynamic_cast<*>` is `nullptr`.

tests/constraint/MockConstraints.hpp
```diff
@@ -14,11 +17,23 @@ class PassingConstraint : public aikido::constraint::Testable
...
   bool isSatisfied(
-      const aikido::statespace::StateSpace::State* /*state*/) const override
+      const aikido::statespace::StateSpace::State* /*state*/,
+      TestableOutcome* outcome = nullptr) const override
   {
+    auto defaultOutcomeObject
+        = aikido::constraint::dynamic_cast_if_present<DefaultTestableOutcome>(
+            outcome);
+
+    if (defaultOutcomeObject)
```
jslee02: ditto:
> I don't think we need the nullity check here because `dynamic_cast_if_present` throws an exception if `dynamic_cast<*>` is `nullptr`.

tests/constraint/MockConstraints.hpp
```diff
@@ -38,11 +53,23 @@ class FailingConstraint : public aikido::constraint::Testable
...
```
jslee02: ditto:
> I don't think we need the nullity check here because `dynamic_cast_if_present` throws an exception if `dynamic_cast<*>` is `nullptr`.

tests/constraint/test_FrameTestable.cpp
```diff
@@ -29,13 +30,27 @@ class EndEffectorTestable : public aikido::constraint::Testable
...
+
     auto st = static_cast<const SE3::State*>(_state);
     auto val = st->getIsometry();
     auto trans = val.translation();
-    return (
-        trans(0) > 0.5 && trans(0) < 1.5 && trans(1) > 0.0 && trans(1) < 1.0);
+
+    bool isSatisfiedResult
+        = trans(0) > 0.5 && trans(0) < 1.5 && trans(1) > 0.0 && trans(1) < 1.0;
+    if (defaultOutcomeObject)
```
jslee02: ditto:
> I don't think we need the nullity check here because `dynamic_cast_if_present` throws an exception if `dynamic_cast<*>` is `nullptr`.

tests/planner/ompl/OMPLTestHelpers.hpp
```diff
@@ -154,11 +156,25 @@ class MockProjectionConstraint : public aikido::constraint::Projectable,
...
+    auto defaultOutcomeObject
+        = aikido::constraint::dynamic_cast_if_present<DefaultTestableOutcome>(
+            outcome);
+
     auto state = static_cast<const CartesianProduct::State*>(_s);
     auto val = mStateSpace->getSubStateHandle<R3>(state, 0).getValue();
-    return std::fabs(val[0] - mValue) < 1e-6;
+
+    bool isSatisfiedResult = std::fabs(val[0] - mValue) < 1e-6;
+    if (defaultOutcomeObject)
```
jslee02: ditto:
> I don't think we need the nullity check here because `dynamic_cast_if_present` throws an exception if `dynamic_cast<*>` is `nullptr`.

tests/planner/ompl/OMPLTestHelpers.hpp
```diff
@@ -200,13 +221,22 @@ class MockTranslationalRobotConstraint : public aikido::constraint::Testable
     {
       if (val[i] < mLowerBounds[i] || val[i] > mUpperBounds[i])
       {
+        if (defaultOutcomeObject)
```
jslee02: ditto:
> I don't think we need the nullity check here because `dynamic_cast_if_present` throws an exception if `dynamic_cast<*>` is `nullptr`.

tests/planner/ompl/OMPLTestHelpers.hpp
```diff
@@ -200,13 +221,22 @@ class MockTranslationalRobotConstraint : public aikido::constraint::Testable
...
+          defaultOutcomeObject->setSatisfiedFlag(true);
         return true;
       }
     }
 
+    if (defaultOutcomeObject)
```
jslee02: ditto:
> I don't think we need the nullity check here because `dynamic_cast_if_present` throws an exception if `dynamic_cast<*>` is `nullptr`.

include/aikido/constraint/TestableOutcome.hpp
```diff
@@ -0,0 +1,51 @@
...
+  if (!childPtr)
+  {
+    std::stringstream message;
+    message << "TestableOutcome pointer is not of type " << typeid(Child).name()
+            << ".";
+    throw std::invalid_argument(message.str());
+  }
+
+  return childPtr;
+}
```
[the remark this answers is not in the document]
jslee02: > Where would be the best place to document this type of convention?

@mkoval For now, I believe the best place would be [`STYLE.md`](https://github.com/personalrobotics/aikido/blob/master/STYLE.md) adding a subsection under `C++ Style`.

include/aikido/constraint/CollisionFreeOutcome.hpp
```diff
@@ -0,0 +1,69 @@
...
+  /// Gets the name of a CollisionObject. The name returned is that of the
+  /// corresponding BodyNode (if possible). If not, the name of the ShapeFrame
+  /// is returned instead. This is a helper for toString().
+  /// \param[in] object object pointer to CollisionObject we want the name of.
+  std::string getCollisionObjectName(
+      const dart::collision::CollisionObject* object) const;
+
+protected:
+  /// Holds Contact objects from pairwise collisions.
+  std::vector<dart::collision::Contact> mPairwiseContacts;
```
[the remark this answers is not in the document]
jslee02: @mkoval `dart::collision::Contact` actually contains no types listed in the [link](https://eigen.tuxfamily.org/dox/group__TopicFixedSizeVectorizable.html); `Eigen::Vector3d` is not applicapable.

Sorry, I should have replied this before @sniyaz make the changes. :sob:

include/aikido/constraint/TestableOutcome.hpp
```diff
@@ -0,0 +1,36 @@
...
+  /// String representation of the outcome. Provides useful, programmatic
+  /// access to debug information.
+  virtual std::string toString() const = 0;
+};
+
+/// Helper function. Avoids repeating logic for casting TestableOutcome
+/// pointers down to pointers for a derivative class. Mostly used in the
+/// isSatisfied methods of classes that inherit Testable.
+template <class Child>
+Child* dynamic_cast_if_present(TestableOutcome* outcome);
```
jslee02: Suggestion: I think it would be more informative if the name of this function represents that it throws an exception when it fails to dynamic-cast, which is that main difference from `dynamic_cast`, something like `dynamic_cast_or_throw` or `strict_dynamic_cast`. How does this sound?

include/aikido/constraint/CollisionFreeOutcome.hpp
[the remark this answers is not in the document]
mkoval: You're right! 😱 I always forget that `Vector3d` is not a vectorizable because it contains an odd number of bytes. Sorry creating extra work. 😰

sniyaz:
Sorry for the pause- this was left unmerged because of the demo. I've addressed the final comments, and this should be ready to merge once and for all.

@brianhou or @jslee02, could one of you do the honors if these updates look good to you?

jslee02:
Oh, one minor thing. It would be great to add an item for this change to `CHANGELOG.md`.

jslee02:
@brianhou Thanks for updating `CHANGELOG.md`. Please merge this if this looks good to you as well.

Issue #267, opened 2017-11-26
VectorFieldPlanner 


dqyi11:
VectorField planner that plans a path by following defined joint velocities in configuration space and terminates by defined criteria.
- Use `boost::numeric::odeint::integrate_adaptive` that integrates joint velocities with adaptive step size
- Refactor implementation to follow the design in `prpy`;
[https://github.com/personalrobotics/prpy/blob/master/src/prpy/planning/vectorfield.py](https://github.com/personalrobotics/prpy/blob/master/src/prpy/planning/vectorfield.py)
- Add three vector fields 
  - [MoveEndEffectorOffset] that moves an end-effector to an offset by a direction;
  - [MoveEndEffectorPose] that moves an end-effector to a goal pose;
  - [MoveEndEffectorPoseAlong] that moves an end-effector along a given workspace trajectory.

jslee02:
@dqyi11 All the issues here resolved by #268? If so, could you please close this?

dqyi11:
Not yet, there will be another PR for 
> [MoveEndEffectorPoseAlong] that moves an end-effector along a given workspace trajectory.

jslee02:
I added a checklist to the description accordingly. Feel free to modify if necessary.

Issue #212, opened 2017-05-26
Implement collision checking for Spline trajectories


mkoval:
It would be very useful to implement a method that collision checks spline trajectories. The core piece of technology that we are missing is a function that takes a `trajectory::Spline` as input and returns a list of times such that every point on the trajectory is within some radius of the points obtained by evaluating the trajectory at those times. (That was a mouthful.)

We have something like this in the [`CheckRamp` function](https://github.com/personalrobotics/aikido/blob/master/src/external/hauser_parabolic_smoother/DynamicPath.cpp#L490) included with Hauser's parabolic smoother. This function constructs a piecewise linear approximation of a parabolic spline, then runs a bisection collision check on the resulting piecewise linear trajectory. This is a reasonable approach that is described in Section IV.E. of the [accompanying paper](https://pdfs.semanticscholar.org/c468/4d698ce0ba92706f5b70fdab9e09c1caa834.pdf).

If we had function, then it would be relatively straightforward to replace significant parts of the current parabolic smoothing implementation, which is currently built as an external project, with the Aikido equivalents. It would also let us experiment with other spline-based trajectory optimization and retiming algorithms.

PR #268, opened 2017-11-26
New vector field planner impelementation

dqyi11:
This PR address the enhancement #267 .
The changes include
- A new `ConfigurationSpaceVectorField` is created.
  - `MoveEndEffectorOffsetVectorField` is added
  - `MoveEndEffectorPoseVectorField` is added
- A new `VectorFieldPlanner` class is created 
  - Call `boost::numeric::odeint::integrate_adaptive` to integrate by adaptive step size
  - Add `planToEndEffectorPose`
- Added test cases for `planToEndEffectorPose`

**TODO** 
These will be added in another PR.
- Collision checking of a trajectory (or segment) will be added (using sampling based method in `prpy`)
- `MoveEndEffectorAlongWorkspacePathVectorField` will be added
- `planWorkspacePath` will be added


src/planner/vectorfield/VectorFieldUtil.cpp
```diff
@@ -52,6 +53,7 @@ std::unique_ptr<aikido::trajectory::Spline> convertToSpline(
 }
 
 //==============================================================================
+
```
jslee02: Nit: Adding empty line here is not our convention.

src/planner/vectorfield/VectorFieldUtil.cpp
```diff
@@ -62,26 +64,32 @@ DesiredTwistFunction::DesiredTwistFunction(
...
```
jslee02: Nit: Adding empty line here is not our convention.

src/planner/vectorfield/VectorFieldUtil.cpp
```diff
@@ -62,26 +64,32 @@ DesiredTwistFunction::DesiredTwistFunction(
...
 double DesiredTwistFunction::eval(const Eigen::VectorXd& _qd)
 {
   return 0.5 * (mJacobian * _qd - mTwist).squaredNorm();
 }
 
 //==============================================================================
+
```
jslee02: Nit: Adding empty line here is not our convention.

src/planner/vectorfield/VectorFieldUtil.cpp
```diff
@@ -62,26 +64,32 @@ DesiredTwistFunction::DesiredTwistFunction(
...
 void DesiredTwistFunction::evalGradient(
     const Eigen::VectorXd& _qd, Eigen::Map<Eigen::VectorXd> _grad)
 {
   _grad = mJacobian.transpose() * (mJacobian * _qd - mTwist);
 }
 
 //==============================================================================
+
```
jslee02: Nit: Adding empty line here is not our convention.

include/aikido/planner/vectorfield/VectorFieldPlanner.hpp
```diff
@@ -2,58 +2,88 @@
...
+  /// Generate a trajectory following the vector field along given time
+  ///
+  /// \param[in] _integrationTimeInterval Position in configuration space
+  /// \param[in] _timelimit Timelimit for integration calculation
+  /// \param[in] _useCollisionChecking Whether collision checking is
+  /// considered in planning.
+  /// \param[in] _useDofLimitChecking Whether Dof Limits are considered
+  /// in planning
+  /// \return A trajectory following the vector field
+  std::unique_ptr<aikido::trajectory::Spline> followVectorField(
```
jslee02: I would like to suggest changing this function name to `plan`. I believe the information that the trajectory is generated to follow the given vector field is encoded by the class name.

include/aikido/planner/vectorfield/VectorFieldPlanner.hpp
```diff
@@ -2,58 +2,88 @@
...
-    const aikido::statespace::dart::MetaSkeletonStateSpacePtr& _stateSpace,
-    const aikido::constraint::TestablePtr& _constraint,
-    double _timestep,
-    const VectorFieldCallback& _vectorFieldCb,
-    const VectorFieldStatusCallback& _statusCb);
+  /// Check status after every intergration step
+  ///
+  /// \param[in] _q Position in configuration space
+  /// \param[in] _t Current time being planned
+  void check(const Eigen::VectorXd& _q, double _t);
```
jslee02: I think `step` and `check` functions should be protected because they're just used internally in this class.

include/aikido/planner/vectorfield/VectorFieldPlanner.hpp
```diff
@@ -2,58 +2,88 @@
...
-/// \param[in] _t Planned time of a given duration
-/// \param[out] _dq Joint velocity calculated by a vector field and meta
-/// skeleton
-/// \return Whether vectorfield evaluation succeeds
-using VectorFieldCallback = std::function<bool(
-    const aikido::statespace::dart::MetaSkeletonStateSpacePtr& _stateSpace,
-    double _t,
-    Eigen::VectorXd& _dq)>;
+  VectorFieldPlanner(
+      const aikido::planner::vectorfield::ConfigurationSpaceVectorFieldPtr
```
jslee02: Nit: We don't need to specify the whole namespaces. This can be simplified as
```c++
const ConfigurationSpaceVectorFieldPtr
```

Please remove `aikido::[planner::[vectorfield]]` in this header.

include/aikido/planner/vectorfield/VectorFieldPlanner.hpp
```diff
@@ -2,58 +2,88 @@
...
+  double mTimelimit;
+
+  std::vector<Knot> mKnots;
+
+  double mInitialStepSize;
+  int mCacheIndex;
+  int mIndex;
+
+  bool mEnableCollisionCheck;
+  bool mEnableDofLimitCheck;
```
jslee02: It seems `mEnableCollisionCheck` and `mEnableDofLimitCheck` are not actually used because they are always overwritten in `followVectorField()`.

I think we should either (1) remove these member variables and always pass them as parameters of `followVectorField()` or (2) add getters and setters for them and remove those parameters from `followVectorField()`. I'm inclined to (2) because those parameters should represent _state_ of the planner but not temporary variables.

If we choose (2) then we might want to set those parameters in the constrcutor.

include/aikido/planner/vectorfield/VectorFieldPlanner.hpp
```diff
@@ -63,31 +93,81 @@ std::unique_ptr<aikido::trajectory::Spline> planPathByVectorField(
 /// \param[in] _constraint Trajectory-wide constraint that must be satisfied
 /// \param[in] _direction Direction of moving the end-effector
 /// \param[in] _distance  Distance of moving the end-effector
-/// \param[in] _linearVelocity Linear velocity of moving the end-effector
-/// \param[in] _linearTolerance How a planned trajectory is allowed to
+/// \param[in] _maxDistance Max distance of moving the end-effector
```
jslee02: I'm confused with what this function wants to do. Does this function plan a trajectory moving the EE for a certain distance (> distance & > maxDistance) along the given direction, which is my interpretation according to the logic of `checkPlanningStatus()`.

src/planner/vectorfield/VectorFieldPlanner.cpp
```diff
@@ -14,229 +16,248 @@ namespace planner {
...
-    if (static_cast<std::size_t>(dq.size()) != numDof)
-    {
-      throw std::length_error(
-          "Vector field returned an incorrect number of DOF velocities.");
-    }
+  if (status == VectorFieldPlannerStatus::TERMINATE
+      || status == VectorFieldPlannerStatus::CACHE_AND_TERMINATE)
+  {
+    throw VectorFieldTerminated(
+        "Planning was terminated by the StatusCallback.");
```
jslee02: Why do we treat the termination of planning (which I believe it's successfully terminated) as an exception?

include/aikido/planner/vectorfield/ConfigurationSpaceVectorField.hpp
```diff
@@ -0,0 +1,75 @@
...
+  /// \return Whether joint velocities are successfully computed
+  virtual bool getJointVelocities(Eigen::VectorXd& _qd) = 0;
+
+  /// Vectorfield planning status callback function
+  ///
+  /// \return Status of planning
+  virtual VectorFieldPlannerStatus checkPlanningStatus() = 0;
+
+  /// Meta skeleton spate space
+  virtual aikido::statespace::dart::MetaSkeletonStateSpacePtr
```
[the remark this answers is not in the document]
dqyi11: No. I don't know why I wrote that way. Good catch! Thanks!

include/aikido/planner/vectorfield/ConfigurationSpaceVectorField.hpp
```diff
@@ -0,0 +1,75 @@
...
+  getMetaSkeletonStateSpace()
+  {
+    return mStateSpace;
+  }
```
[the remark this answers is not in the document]
dqyi11: Should `ConstPtr` be defined in original header file? Fox example `using MetaSkeletonStateSpaceConstPtr = std::shared_ptr<const MetaSkeletonStateSpace>` in MetaSkeletonStateSpace.hpp?

include/aikido/planner/vectorfield/MoveEndEffectorOffsetVectorField.hpp
```diff
@@ -12,70 +12,58 @@ namespace vectorfield {
...
-      double _endTime,
-      double _timestep,
-      double _linearGain = 1.,
-      double _linearTolerance = 0.01,
-      double _angularGain = 1.,
-      double _angularTolerance = 0.01,
-      double _optimizationTolerance = 1e-3,
-      double _padding = 1e-5);
+      const Eigen::Vector3d& _direction,
+      double _distance,
```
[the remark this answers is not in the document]
dqyi11: This is following the naming in `prpy`. This is usually the expected distance to move. The planner will terminate once this distance is reached. It will fail it `_maxDistance` is reached.

include/aikido/planner/vectorfield/MoveEndEffectorOffsetVectorField.hpp
```diff
@@ -12,70 +12,58 @@ namespace vectorfield {
...
+      double _maxDistance = std::numeric_limits<double>::max(),
+      double _positionTolerance = 0.01,
+      double _angularTolerance = 0.15,
+      double _linearVelocityGain = 1.0,
+      double _initialStepSize = 1e-1,
+      double _jointLimitTolerance = 3e-2,
```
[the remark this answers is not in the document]
dqyi11: Yes. I will use `positionLimitPadding`.

include/aikido/planner/vectorfield/VectorFieldPlanner.hpp
dqyi11: :+1:

include/aikido/planner/vectorfield/VectorFieldPlanner.hpp
dqyi11: These two member variables are actually used in `check()` and `step()`. These two functions are only internally called inside `followVectorField()`. These two parameters are about how you shall follow a vector field (like integral time), so I put them as input arguments of `followVectorField()`. They are set to member variables so that `check()` and `step()` could access. Any suggestion?

include/aikido/planner/vectorfield/VectorFieldPlanner.hpp
dqyi11: Once the moved distance is larger than `distance`, it will terminate and return a trajectory.
But if the moved distance is larger than `maxDistance`, it violates and will fail.

include/aikido/planner/vectorfield/VectorFieldPlannerExceptions.hpp
```diff
@@ -9,18 +9,33 @@ namespace aikido {
 namespace planner {
 namespace vectorfield {
 
-/// Define runtime error for the termination of vectorfield planner.
-class VectorFieldTerminated : public std::runtime_error
+/// Define termination of vectorfield planner.
+class VectorFieldTerminated : public std::exception
```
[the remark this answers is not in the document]
dqyi11: `VectorFIeldTerminated` defines the end of the planning, which implies mission completed. I throw it as an exception, so that the integrator could be stopped. The input integral time is usually larger than it should be. 

`VectorFieldError` defines the error found in the planning process, so it inherits `std::runtime_error`.

include/aikido/planner/vectorfield/VectorFieldPlannerStatus.hpp
```diff
@@ -0,0 +1,26 @@
...
+/// CACHE_AND_CONTINUE - save the current trajectory and CONTINUE.
+///                      return the saved trajectory if TERMINATEd.
+/// CONTINUE - keep going
+/// CACHE_AND_TERMINATE - save the current trajectory and TERMINATE
+enum class VectorFieldPlannerStatus
+{
+  TERMINATE = -1,
+  CACHE_AND_CONTINUE = 0,
+  CONTINUE = 1,
+  CACHE_AND_TERMINATE = 2
```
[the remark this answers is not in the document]
dqyi11: No. I will remove them.

include/aikido/planner/vectorfield/VectorFieldUtil.hpp
```diff
@@ -67,22 +68,67 @@ class DesiredTwistFunction : public dart::optimizer::Function
...
-/// \param[in] _timestep How long will the computed joint velocities be executed
-/// \param[in] _padding Padding for joint limits
 /// \param[out] _jointVelocity Calculated joint velocities
 bool computeJointVelocityFromTwist(
     const Eigen::Vector6d& _desiredTwist,
     const aikido::statespace::dart::MetaSkeletonStateSpacePtr _stateSpace,
     const dart::dynamics::BodyNodePtr _bodyNode,
+    double _jointLimitTolerance,
+    const Eigen::VectorXd& _jointVelocityLowerLimits,
+    const Eigen::VectorXd& _jointVelocityUpperLimits,
```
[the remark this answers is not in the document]
dqyi11: The original reason is allowing set joint velocity limits as `INF` like `prpy`. 
But this will cause `nlopt` error. I thus added `_jointVelocityLimited` . 
I still keep these two arguments, assuming users might set joint velocity limits that are different with meta skeleton's. (for example, tighter bounds or looser bounds). What do you think?

include/aikido/planner/vectorfield/VectorFieldUtil.hpp
```diff
@@ -67,22 +68,67 @@ class DesiredTwistFunction : public dart::optimizer::Function
...
+Eigen::Vector4d computeGeodesicError(
+    const Eigen::Isometry3d& _currentTrans,
+    const Eigen::Isometry3d& _goalTrans);
+
+/// Calculate the geodesic distance between two transforms, being
+/// gd = norm( relative translation + r * axis-angle error )
+/// \param[in] _currentTrans Current transformation
+/// \param[in] _goalTrans Goal transformation
+/// \param[in] _r In units of meters/radians converts radians to meters
+double computeGeodesicDistanceBetweenTransforms(
```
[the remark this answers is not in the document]
dqyi11: haha, good catch. I copy them from `prpy`. Yes. You are right. I can just keep `computeGeodesicDistance`.

src/planner/vectorfield/VectorFieldPlanner.cpp
dqyi11: You are right. This is the way of stopping integration before it is done. Usually we set 10 sec to integrate but can reach an expected distance before that. Any other suggested way?

src/planner/vectorfield/VectorFieldUtil.cpp
```diff
@@ -62,26 +64,32 @@ DesiredTwistFunction::DesiredTwistFunction(
...
 double DesiredTwistFunction::eval(const Eigen::VectorXd& _qd)
 {
   return 0.5 * (mJacobian * _qd - mTwist).squaredNorm();
```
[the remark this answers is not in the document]
dqyi11: They are different. Here the objective is finding a joint velocity that could make the twist as close as to a desired Twist. The `checking planning status` is determining whether any constraint is violated or the goal is reached in following a vector field.

src/planner/vectorfield/VectorFieldUtil.cpp
```diff
@@ -95,49 +103,51 @@ bool computeJointVelocityFromTwist(
...
+      const double position = positions[i];
+      const double positionLowerLimit = positionLowerLimits[i];
+      const double positionUpperLimit = positionUpperLimits[i];
+      const double velocityLowerLimit = velocityLowerLimits[i];
+      const double velocityUpperLimit = velocityUpperLimits[i];
+
+      if (position + _maxStepSize * velocityLowerLimit
+          <= positionLowerLimit + _jointLimitTolerance)
+      {
+        velocityLowerLimits[i] = 0.0;
```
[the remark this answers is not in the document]
dqyi11: I change it back to this way, which is the same with `prpy` implementation and `muul` implementation.  Another reason is that now it is adaptive step size in integration, the step size is not the true step size will be taken in the next step of integration. Do you agree?

tests/planner/vectorfield/test_VectorFieldPlanner.cpp
```diff
@@ -141,67 +139,37 @@ class VectorFieldPlannerTest : public ::testing::Test
...
-        ->createShapeNodeWith<VisualAspect, CollisionAspect, DynamicsAspect>(
-            shape);
-  }
-
   //==============================================================================
   /// Creates a N link manipulator with the given dimensions where each joint is
   /// the specified type
-  SkeletonPtr createNLinkRobot(
-      int _n, Vector3d _dim, TypeOfDOF _type, bool _finished = false)
+  SkeletonPtr createNLinkRobot()
```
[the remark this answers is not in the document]
dqyi11: This supposed to be a temporary workaround. Could we make the changes in the enhancement of a robot helper class for unit test?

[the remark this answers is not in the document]
jslee02: We could do this in a future PR. Please create an issue for this to track.

include/aikido/planner/vectorfield/VectorFieldPlanner.hpp
dqyi11: Yes. You are right. I will update the doc strings accordingly :)

dqyi11:
@mkoval How about the below design of classes to support vector field planner in arbitrary `StateSpace`?

`VectorField` class provides `evaluateVelocity()` and `evaluateStatus()`.
`VectorFieldPlanner` class provides `evaluateTrajectory()` and `followVectorField()`.

Define an abstract vector field for any state space.
```
class VectorField
{
public:
  bool evaluateVelocity(const StateSpace::State* _state,
                        Eigen::VectorXd* _velocity) = 0;

  // Alternative: Status is only a function of the current state.
  VectorFieldPlannerStatus evaluateStatus(
    const StateSpace::State* _state) = 0;
};
```
Define a `MetaSkeletonStateSpaceVectorField` in robot's configuration space.
(Currently we do not have one, but it is possible that there will be applications that is directly defined in MetaSkeletonStateSpace, for example, something similar with `planToConfiguration`?) 
```
class MetaSkeletonStateSpaceVectorField : public VectorField
{
public:
  MetaSkeletonStateSpaceVectorField(
      aikido::statespace::dart::MetaSkeletonStateSpacePtr stateSpace);
  bool evaluateVelocity(const StateSpace::State* _state,
                        Eigen::VectorXd* _velocity);

  // Alternative: Status is only a function of the current state.
  VectorFieldPlannerStatus evaluateStatus(
    const StateSpace::State* _state);
};
```
Define a `BodyNodePoseVectorField` for a type of vector field in robot's configuration space that is created from workspace.
```
class BodyNodePoseVectorField : public MetaSkeletonStateSpaceVectorField
{
BodyNodePoseVectorField(
      aikido::statespace::dart::MetaSkeletonStateSpacePtr stateSpace,
      dart::dynamics::BodyNodePtr bodyNode);
};
```
For example, define a `MoveEndEffectorOffsetVectorField` for a specific `BodyNodePoseVectorField`.
```
class MoveEndEffectorOffsetVectorField : public BodyNodePoseVectorField
{
};
```
Correspondingly, we could have classes of `VectorFieldPlanner` (will be in `details`) as below. 
`VectorFieldPlanner` classes provides `evalutateTrajectory()` and `followVectorField()`.
```
class VectorFieldPlanner
{
public:
  VectorFieldPlanner(
      const VectorFieldPtr vectorField,
      const aikido::constraint::TestablePtr constraint);
              
  bool evaluateTrajectory(
      const aikido::trajectory::Trajectory& _trajectory,
      const aikido::constraint::TestablePtr constraint);

  std::unique_ptr<aikido::trajectory::Spline> followVectorField();
};
```
A `MetaSkeletonStateSpaceVectorFieldPlanner` for all the vector fields defined in robot's configuration space.
The `evaluateTrajector()` accepts only **timed** trajectory.
It checks the satisfactions of not only `CollisionFree` constraint and DOF limit constraint, but also joint velocity limit constraint.
```
class MetaSkeletonStateSpaceVectorFieldPlanner : public VectorFieldPlanner
{
public:
  VectorFieldPlanner(
      const MetaSkeletonStateSpaceVectorFieldPtr vectorField,
      const aikido::constraint::TestablePtr constraint);
};
```
It seems that `Testable` is for a `State` in state space. What is the best practice of using it to check velocity limit violation?

src/planner/vectorfield/MoveEndEffectorOffsetVectorField.cpp
```diff
@@ -3,145 +3,130 @@
...
   // Check for deviation from the straight-line trajectory.
-  const Vector3d linearError
-      = mTargetPose.translation() - currentPose.translation();
-  const Vector3d linearOrthogonalError
-      = linearError - linearError.dot(mLinearDirection) * mLinearDirection;
-  double linearOrthogonalMagnitude = linearOrthogonalError.norm();
+  const Vector4d geodesicError = computeGeodesicError(mStartPose, currentPose);
+  const double orientationError = geodesicError[0];
+  const Vector3d positionError = geodesicError.tail<3>();
+  double movedDistance = positionError.transpose() * mDirection;
```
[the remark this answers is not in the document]
dqyi11: Yes. It is normalized in the constructor.

dqyi11:
Changes include:
- **Integration time** is no longer a parameter but a constant. 
- **Non-`Rn` State Space Support** will be addressed in #274 
- **Non-`MetaSkeletonStateSpace` Support** is done. New classes are added for vector fields in arbitrary state space.
- **Joint Limits and Collision Checking** are now called by using `Testable` constraints.
`Eigen::VectorXd` is replaced with `aikido::statespace::StateSpace::State`. 
- **Constraint Validation** is done using a different resolution parameter with integration step size.
Our original plan was using a trajectory-wide constraint validation after vector field planner generates a path, so the constraint validation is following a rough observation interval (defined by the `stepper()`). But now `evaludateTrajectory()` is added to support trajectory-wide constraint validation. It is used to check in new segment in the planning process (needs a user-defined resolution parameter).

include/aikido/planner/vectorfield/VectorFieldIntegrator.hpp
```diff
@@ -0,0 +1,136 @@
...
+  virtual bool evaluateTrajectory(
+      const aikido::trajectory::Trajectory* trajectory,
+      const aikido::constraint::TestablePtr collisionFreeConstraint,
+      double evalStepSize)
+      = 0;
+
+  virtual bool convertStateToPositions(
+      const aikido::statespace::StateSpace::State* state,
+      Eigen::VectorXd& positions)
+      = 0;
```
mkoval: It's possible to implement this function on a generic state space simply by calling `logMap()`.

Note that this does not preserve the topology of non-Euclidean state spaces, i.e. the integrator may get "stuck" because it is unable to "wrap around" singularities. This is a fundamental issue until we resolve https://github.com/personalrobotics/aikido/issues/274 - and is true with the current implementation, which simply asks DART to perform the `logMap()` for us.

include/aikido/planner/vectorfield/VectorFieldIntegrator.hpp
```diff
@@ -0,0 +1,136 @@
...
+
+  virtual bool convertPositionsToState(
+      const Eigen::VectorXd& positions,
+      aikido::statespace::StateSpace::State* state)
+      = 0;
```
mkoval: It's possible to implement this function on a generic state space by calling `expMap()`. See my comment on `convertStateToPositions()` for more information

include/aikido/planner/vectorfield/VectorFieldIntegrator.hpp
```diff
@@ -0,0 +1,136 @@
...
+
+  /// Vectorfield callback function that returns joint velocities for
+  /// integration.
+  ///
+  /// \param[in] q Position in configuration space.
+  /// \param[out] qd Joint velocities in configuration space.
+  /// \param[in] t Current time being planned.
+  virtual void step(const Eigen::VectorXd& q, Eigen::VectorXd& qd, double t);
+
+  /// Check status after every intergration step.
```
mkoval: Typo: "intergration" -> "integration"

include/aikido/planner/vectorfield/VectorFieldUtil.hpp
```diff
@@ -41,47 +34,83 @@ class DesiredTwistFunction : public dart::optimizer::Function
...
-      const Eigen::VectorXd& _qd, Eigen::Map<Eigen::VectorXd> _grad) override;
+      const Eigen::VectorXd& qd, Eigen::Map<Eigen::VectorXd> grad) override;
 
 private:
+  /// Twist.
   Twist mTwist;
+
+  /// Jacobian of Meta Skeleton
   Jacobian mJacobian;
 };
```
mkoval: I believe this is an implementation detail of `computeJointVelocityFromTwist()`. I suggest moving this class to an anonymous namespace in this `.cpp` file.

include/aikido/planner/vectorfield/VectorFieldUtil.hpp
```diff
@@ -41,47 +34,83 @@ class DesiredTwistFunction : public dart::optimizer::Function
...
+    const Eigen::VectorXd& jointVelocityLowerLimits,
+    const Eigen::VectorXd& jointVelocityUpperLimits,
+    bool jointVelocityLimited,
+    double stepSize);
+
+/// Compute the twist in global coordinate that corresponds to the gradient of
+/// the geodesic distance between two transforms.
+///
+/// \param[in] currentTrans Current transformation.
+/// \param[in] goalTrans Goal transformation.
```
mkoval: Add a docstring for the return value. Particularly: document what order its elements are in.

include/aikido/planner/vectorfield/VectorFieldUtil.hpp
```diff
@@ -41,47 +34,83 @@ class DesiredTwistFunction : public dart::optimizer::Function
...
+Eigen::Vector6d computeGeodesicTwist(
+    const Eigen::Isometry3d& currentTrans, const Eigen::Isometry3d& goalTrans);
+
+/// Compute the error in gloabl coordinate between two transforms.
+///
+/// \param[in] currentTrans Current transformation.
+/// \param[in] goalTrans Goal transformation.
```
mkoval: Add a docstring for the return value. Particularly: document what order its elements are in.

include/aikido/planner/vectorfield/VectorFieldUtil.hpp
```diff
@@ -41,47 +34,83 @@ class DesiredTwistFunction : public dart::optimizer::Function
...
+Eigen::Vector4d computeGeodesicError(
+    const Eigen::Isometry3d& currentTrans, const Eigen::Isometry3d& goalTrans);
+
+/// Compute the geodesic distance between two transforms.
+/// gd = norm( relative translation + r * axis-angle error )
+///
+/// \param[in] currentTrans Current transformation.
+/// \param[in] goalTrans Goal transformation.
+/// \param[in] r In units of meters/radians converts radians to meters.
```
mkoval: Add a docstring for the return value.

include/aikido/planner/vectorfield/VectorFieldUtil.hpp
```diff
@@ -41,47 +34,83 @@ class DesiredTwistFunction : public dart::optimizer::Function
...
+double computeGeodesicDistance(
+    const Eigen::Isometry3d& currentTrans,
+    const Eigen::Isometry3d& goalTrans,
+    double r = 1.0);
```
mkoval: I don't think `r` should have a default. There's no reason to think that `1` is a good value for this - in fact, this is likely far too large for HERB because his entire workspace is only ~2 m wide.

include/aikido/planner/vectorfield/detail/VectorFieldPlannerExceptions.hpp
```diff
@@ -0,0 +1,81 @@
...
+};
+
+/// Define state in collision for the termination of vector field planner.
+class StateInCollisionError : public VectorFieldError
+{
+public:
+  /// Constructor.
+  ///
+  StateInCollisionError();
+};
```
mkoval: This could be thrown for any constraint violation, not just collision. I suggest renaming this to `ConstraintViolatedError`.

include/aikido/planner/vectorfield/detail/VectorFieldPlannerExceptions.hpp
```diff
@@ -0,0 +1,81 @@
...
+
+/// Define termination error of vectorfield planner due to DOF limit error.
+class DofLimitError : public VectorFieldError
+{
+public:
+  /// Constructor.
+  ///
+  DofLimitError();
+};
```
mkoval: It will no longer be possible to raise this error after removing `getBoundTestable()`. I suggest removing it.

We can restore fancy error messages later using a framework like https://github.com/personalrobotics/aikido/pull/266.

include/aikido/planner/vectorfield/MetaSkeletonStateSpaceVectorField.hpp
```diff
@@ -0,0 +1,68 @@
...
+namespace aikido {
+namespace planner {
+namespace vectorfield {
+
+/// This class defines a vector field in a configuration space.
+///
+/// Any vector field should inherit this class to implememnt functions
+/// for calculating joint velocities giving joint positions and time,
+/// and checking planning status.
+class MetaSkeletonStateSpaceVectorField : public VectorField
```
mkoval: I don't think this class is necessary - it doesn't seem to provide any value over `VectorField` itself. I think we can safely roll all of this functionality into `BodyNodePoseVectorField`.

include/aikido/planner/vectorfield/BodyNodePoseVectorField.hpp
```diff
@@ -0,0 +1,53 @@
...
+  /// \param[in] stateSpace MetaSkeleton state space.
+  /// \param[in] bn Body node of end-effector.
+  BodyNodePoseVectorField(
+      aikido::statespace::dart::MetaSkeletonStateSpacePtr stateSpace,
+      dart::dynamics::BodyNodePtr bodyNode);
+
+  // Documentation inherited.
+  virtual bool evaluateVelocity(
+      const aikido::statespace::StateSpace::State* state,
+      Eigen::VectorXd& qd) const override = 0;
```
mkoval: I was expecting this class to define `evaluateVelocity()` in terms of another function that returns the Cartesian velocity of the specified `BodyNode`. E.g.:
```c++
bool BodyNodePoseVectorField::evaluateVelocity(
      const aikido::statespace::StateSpace::State* state,
      Eigen::VectorXd& qd) const override = 0) override
{
  // Pseudocode. Actual implementation will require some massaging of data types.
  const auto cartesianState = mBodyNode->getPose(state);

  Eigen::Vector6d cartesianVelocity;
  if (!evaluateCartesianVelocity(cartesianState, cartesianVelocity))
    return false;

  return computeJointVelocityFromTwist(mBodyNode, cartesianVelocity, &qd);
}

// Implemented by the user:
virtual bool BodyNodePoseVectorField::evaluateCartesianVelocity(
  const Eigen::Isometry3d& _pose, Eigen::Vector6d& _cartesianVelocity) = 0;
```

In my opinion, the _whole point_ of this class is to allow the user to define a vector field over the pose of a `BodyNode` without worrying about the boilerplate required to call `computeJointVelocityFromTwist().

include/aikido/planner/vectorfield/BodyNodePoseVectorField.hpp
```diff
@@ -0,0 +1,53 @@
...
+
+  // Documentation inherited.
+  virtual VectorFieldPlannerStatus evaluateStatus(
+      const aikido::statespace::StateSpace::State* state) const override = 0;
```
mkoval: I was expecting this class to define `evaluateStatus()` in terms of another `evaluateCartesianStatus()` function that accepts the pose of the `BodyNode`. See my comment on `evaluateVelocity()` for details.

src/planner/vectorfield/MetaSkeletonStateSpaceVectorField.cpp
```diff
@@ -0,0 +1,65 @@
...
+  if (mMetaSkeleton->getPositionUpperLimits()
+      == mMetaSkeleton->getPositionLowerLimits())
+  {
+    throw std::invalid_argument("State space volume zero");
+  }
+  if (mMetaSkeleton->getVelocityUpperLimits()
+      == mMetaSkeleton->getVelocityLowerLimits())
+  {
+    throw std::invalid_argument("Velocity space volume zero");
+  }
```
mkoval: I don't think these checks are necessary. Bounds should be handled entirely by the `Testable` passed by the user.

include/aikido/planner/vectorfield/MetaSkeletonStateSpaceVectorFieldIntegrator.hpp
```diff
@@ -0,0 +1,67 @@
...
+
+/// VectorField Planner generates a trajectory by following a vector field
+/// defined in joint space.
+/// A planned trajectroy ends either by a predefined termination criterion or a
+/// integral time.
+///
+/// This class defines two callback functions for a integrator.
+/// step() provides joint velocities the vector field planner should follow,
+/// check() is called after each integration step to check planner status.
+class MetaSkeletonStateSpaceVectorFieldIntegrator : public VectorFieldIntegrator
```
mkoval: I am not sure why this class is necessary. I think we can safely delete this once we implement `convertStateToPositions()` and `convertPositionsToState()` in the base `VectorField` class.

dqyi11:
- **Missing functions on `BodyNodePoseVectorField`** are added. `evaluateVelocity()` and `evaluateStatus()` are added that operate on the BodyNode's pose.
- **`MetaSkeletonStateSpaceVectorField`** and **`MetaSkeletonStateSpaceVectorFieldIntegrator`** are removed. 
- **`Incremental Constraint Validation`** is supported. Now a trajectory is created from a new knot to the last knot in list, and is sent to `evaluateTrajectory` for constraint satisfaction. 
- Docstrings are updated.

TODO:
-**Collision Free and Bounds Constraints**. This will be updated after #266 is merged.
-**Incorrect Constraint Validation**. 
@mkoval  suggested
> The constraint validation currently discretizes the trajectory using an integration time resolution. This is incorrect because integration time is meaningless - we should be collision checking at a fixed DOF resolution. 
This will be updated after #212 is merged. The current solution is adding a user-defined check resolution in `followVectorField()`.

include/aikido/planner/vectorfield/VectorFieldIntegrator.hpp
```diff
@@ -0,0 +1,136 @@
...
+
+protected:
+  /// Evaluate whether a trajectory is eligible
+  ///
+  ///
+  virtual bool evaluateTrajectory(
+      const aikido::trajectory::Trajectory* trajectory,
+      const aikido::constraint::TestablePtr collisionFreeConstraint,
+      double evalStepSize)
+      = 0;
```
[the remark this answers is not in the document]
dqyi11: My current solution is adding an argument in `followVectorField()`, which will be used as the resolution in constraint checking.

include/aikido/planner/vectorfield/VectorFieldIntegrator.hpp
```diff
@@ -0,0 +1,107 @@
...
+
+/// VectorField Planner generates a trajectory by following a vector field
+/// defined in joint space.
+/// A planned trajectroy ends either by a predefined termination criterion or a
+/// integral time.
+///
+/// This class defines two callback functions for a integrator.
+/// step() provides joint velocities the vector field planner should follow,
+/// check() is called after each integration step to check planner status.
+class VectorFieldIntegrator
```
mkoval: I still don't understand why this is a class. Like I said before, I don't see any reason why you would ever want to call `followVectorField()` more than once on an instance of this class.

I suggest pulling `followVectorField()` out into a free function. If we need to maintain state between integration steps, e.g. the partially-complete trajectory, then we can instantiate a `VectorFieldIntegrator` class inside that function. However, that class is an *implementation detail* that belongs *in the `.cpp` file*.

include/aikido/planner/vectorfield/VectorFieldIntegrator.hpp
```diff
@@ -0,0 +1,107 @@
...
+      planner::PlanningResult* planningResult);
+
+protected:
+  /// Vectorfield callback function that returns joint velocities for
+  /// integration.
+  ///
+  /// \param[in] q Position in configuration space.
+  /// \param[out] qd Joint velocities in configuration space.
+  /// \param[in] t Current time being planned.
+  virtual void step(const Eigen::VectorXd& q, Eigen::VectorXd& qd, double t);
```
mkoval: Nit: This does not need to be `virtual`.

include/aikido/planner/vectorfield/VectorFieldIntegrator.hpp
```diff
@@ -0,0 +1,107 @@
...
+
+  /// Check status after every integration step.
+  ///
+  /// \param[in] q Position in configuration space.
+  /// \param[in] t Current time being planned.
+  virtual void check(const Eigen::VectorXd& q, double t);
```
mkoval: Nit: This does not need to be `virtual`.

include/aikido/planner/vectorfield/detail/VectorFieldPlannerExceptions.hpp
```diff
@@ -0,0 +1,81 @@
...
+/// Define termination of vectorfield planner.
+// When VectorFieldTerminated is throws, it means
+// that planning is finished.
+class VectorFieldTerminated : public std::exception
+{
+public:
+  /// Constructor.
+  ///
+  /// \param[in] _whatArg Error string.
+  VectorFieldTerminated(const std::string& whatArg);
```
mkoval: This should be flagged `explicit`.

include/aikido/planner/vectorfield/detail/VectorFieldPlannerExceptions.hpp
```diff
@@ -0,0 +1,81 @@
...
+#include <dart/dynamics/dynamics.hpp>
+
+namespace aikido {
+namespace planner {
+namespace vectorfield {
+
+/// Define termination of vectorfield planner.
+// When VectorFieldTerminated is throws, it means
+// that planning is finished.
+class VectorFieldTerminated : public std::exception
```
mkoval: This is an implementation detail of `followVectorField()` used to terminate the integrator. As such, we should take this out of the header file and put it in the `.cpp` file where it is used in an anonymous `namespace.`

include/aikido/planner/vectorfield/detail/VectorFieldPlannerExceptions.hpp
mkoval: This is still not necessary.

include/aikido/planner/vectorfield/detail/VectorFieldUtil.hpp
```diff
@@ -0,0 +1,123 @@
...
+};
+
+/// Convert a sequence of waypoint and time pairs into a trajectory.
+///
+/// \param[in] A seqeunce of waypoint and time pairs.
+/// \param[in] stateSpace State space of output trajectory.
+/// \return A trajectory.
+std::unique_ptr<aikido::trajectory::Spline> convertToSpline(
+    const std::vector<Knot>& knots,
+    const aikido::statespace::StateSpacePtr stateSpace);
```
mkoval: This is the only function that uses `Knot`. As such, I suggest moving the definition of the `Knot` class into this file.

Nit: It is superfluous to flag `stateSpace` as `const` because it is passed by value.

include/aikido/planner/vectorfield/detail/VectorFieldUtil.hpp
```diff
@@ -0,0 +1,123 @@
...
+    const Eigen::VectorXd& jointVelocityUpperLimits,
+    bool jointVelocityLimited,
+    double stepSize);
+
+/// Compute the twist in global coordinate that corresponds to the gradient of
+/// the geodesic distance between two transforms.
+///
+/// \param[in] fromTrans Current transformation.
+/// \param[in] toTrans Goal transformation.
+/// \return Geodesic twist.
```
mkoval: Document the convention about how to interpret the elements of this vector. Which elements represent linear versus angular velocities? What coordinate frame is this expressed in?

include/aikido/planner/vectorfield/detail/VectorFieldUtil.hpp
```diff
@@ -0,0 +1,123 @@
...
+Eigen::Vector6d computeGeodesicTwist(
+    const Eigen::Isometry3d& fromTrans, const Eigen::Isometry3d& toTrans);
+
+/// Compute the error in gloabl coordinate between two transforms.
+///
+/// \param[in] fromTrans Current transformation.
+/// \param[in] toTrans Goal transformation.
+/// \return Geodesic error.
```
mkoval: Document the convention about how to interpret the elements of this vector. Which elements represent linear versus angular velocities? What coordinate frame is this expressed in?

include/aikido/planner/vectorfield/detail/VectorFieldUtil.hpp
```diff
@@ -0,0 +1,123 @@
...
+Eigen::Vector4d computeGeodesicError(
+    const Eigen::Isometry3d& fromTrans, const Eigen::Isometry3d& toTrans);
+
+/// Compute the geodesic distance between two transforms.
+/// gd = norm( relative translation + r * axis-angle error )
+///
+/// \param[in] fromTrans Current transformation.
+/// \param[in] toTrans Goal transformation.
+/// \param[in] r In units of meters/radians converts radians to meters.
+/// \return Geodesic distance.
```
mkoval: What are the units on the return value?

include/aikido/planner/vectorfield/detail/VectorFieldUtil.hpp
```diff
@@ -0,0 +1,123 @@
...
+/// and linear velocity.
+/// \param[in] stateSpace MetaSkeleton state space.
+/// \param[in] bodyNode Body node of the end-effector.
+/// \param[in] jointLimitPadding If less then this distance to joint
+/// limit, velocity is bounded in that direction to 0.
+/// \param[in] jointVelocityLowerLimits Joint velocity lower bounds.
+/// \param[in] jointVelocityUpperLimits Joint velocity upper bounds.
+/// \param[in] jointVelocityLimited Whether joint velocity bounds are
+/// considered in optimization.
+/// \param[in] stepSize Step size in second.
```
mkoval: What is this used for? Can we remove it?

include/aikido/planner/vectorfield/detail/VectorFieldUtil.hpp
```diff
@@ -0,0 +1,123 @@
...
+///
+/// \param[out] jointVelocity Calculated joint velocities.
+/// \param[in] desiredTwist Desired twist, which consists of angular velocity
+/// and linear velocity.
+/// \param[in] stateSpace MetaSkeleton state space.
+/// \param[in] bodyNode Body node of the end-effector.
+/// \param[in] jointLimitPadding If less then this distance to joint
+/// limit, velocity is bounded in that direction to 0.
+/// \param[in] jointVelocityLowerLimits Joint velocity lower bounds.
+/// \param[in] jointVelocityUpperLimits Joint velocity upper bounds.
```
mkoval: These arguments should only be required if `jointVelocityLimited` is `true`. Are they ignored otherwise?

Instead of having a `jointVelocityLimitReached` flag, consider passing those arguments by `const *` and ignoring the velocity limits if they are `nullptr`.

src/planner/vectorfield/VectorFieldPlanner.cpp
```diff
@@ -1,278 +1,204 @@
...
+        std::bind(&detail::VectorFieldIntegrator::step, integrator, _1, _2, _3),
+        initialQ,
+        0.,
+        integrationTimeInterval,
+        initialStepSize,
+        std::bind(&detail::VectorFieldIntegrator::check, integrator, _1, _2));
   }
-
-  for (std::size_t i = 0; i < skeleton->getNumDofs(); ++i)
+  catch (const detail::VectorFieldTerminated& e)
```
mkoval: Add a comment explaining that `VectorFieldTerminated` is an exception that we raise internally to terminate integration, i.e. it does not indicate that an error has actually occurred.

src/planner/vectorfield/VectorFieldPlanner.cpp
```diff
@@ -1,278 +1,204 @@
...
     }
-  } while (terminationStatus != VectorFieldPlannerStatus::TERMINATE
-           && terminationStatus
-                  != VectorFieldPlannerStatus::CACHE_AND_TERMINATE);
+    return nullptr;
+  }
 
-  // Print the termination condition.
-  if (terminationError)
+  if (integrator->getCacheIndex() < 0)
```
mkoval: I am not sure - should this be `<= 0`? I am having trouble thinking through what we want to happen if only one waypoint is cached, since that doesn't really define a valid trajectory. 🤔 In either case - it would be nice to add a comment here explaining why we use the check that we do.

src/planner/vectorfield/VectorFieldUtil.cpp
```diff
@@ -6,157 +9,189 @@ namespace aikido {
...
-          = std::max(velocityLowerLimits[i], feasibleVelocityLowerLimit);
+      const double position = positions[i];
+      const double positionLowerLimit = positionLowerLimits[i];
+      const double velocityLowerLimit = velocityLowerLimits[i];
+
+      if (position + stepSize * velocityLowerLimit
+          <= positionLowerLimit + jointLimitPadding)
+      {
+        velocityLowerLimits[i] = 0.0;
+      }
```
mkoval: This check needs to live outside the `if (jointVelocityLowerLimits != nullptr)` block. We want to enforce joint limits even if we don't enforce velocity limits.

The simplest way to implement this might be to initialize `jointVelocityLowerLimits` a vector of `+infinity()` constants if it is not passed as an argument, then proceed as before. You might want to double check with @jslee02 that the optimizer will behave as expected if we do this.

src/planner/vectorfield/VectorFieldUtil.cpp
```diff
@@ -6,157 +9,189 @@ namespace aikido {
...
-          = std::min(velocityUpperLimits[i], feasibleVelocityUpperLimit);
+      const double position = positions[i];
+      const double positionUpperLimit = positionUpperLimits[i];
+      const double velocityUpperLimit = velocityUpperLimits[i];
+
+      if (position + stepSize * velocityUpperLimit
+          >= positionUpperLimit - jointLimitPadding)
+      {
+        velocityUpperLimits[i] = 0.0;
+      }
```
mkoval: This should happen even if no upper joint velocity limits are specified. See my comment on the lower joint velocity limits for details.

src/planner/vectorfield/detail/BodyNodePoseVectorFieldConstraint.hpp
```diff
@@ -0,0 +1,44 @@
...
+
+namespace aikido {
+namespace planner {
+namespace vectorfield {
+namespace detail {
+
+/// This class implements a constraint for body node pose vector field.
+/// It checks a given state constraint and a bound constraint of state space.
+///
+class BodyNodePoseVectorFieldConstraint : public aikido::constraint::Testable
```
mkoval: I don't see why this class is necessary. It seems like we could achieve the same result by creating a `CompoundTestable` constraint that wraps `stateConstraint` and the joint limit constraint returned by calling `createTestableBounds()` on the `MetaSkeletonStateSpace`.

src/planner/vectorfield/detail/VectorFieldIntegrator.cpp
```diff
@@ -0,0 +1,172 @@
...
+{
+  return mKnots;
+}
+
+//==============================================================================
+void VectorFieldIntegrator::step(
+    const Eigen::VectorXd& q, Eigen::VectorXd& qd, double /*_t*/)
+{
+  aikido::statespace::StateSpace::State* state
+      = mVectorField->getStateSpace()->allocateState();
```
mkoval: Nit: I suggest allocating this as a member variable, so we can reuse the same memory between steps. You can stored a `ScopedState` to let RAII handle the deallocation for you.

src/planner/vectorfield/detail/VectorFieldIntegrator.cpp
```diff
@@ -0,0 +1,172 @@
...
+  mVectorField->getStateSpace()->expMap(q, state);
+
+  // compute joint velocities
+  bool success = mVectorField->evaluateVelocity(state, qd);
+  mVectorField->getStateSpace()->freeState(state);
```
mkoval: You can eliminate this manual `freeState()` call if you call `createState()` instead of `allocateState()`. That function returns a `ScopedState`, which is a RAII wrapper around `State`.

This line also disappears if we store the state as a member variable. See my suggestion above.

src/planner/vectorfield/detail/VectorFieldIntegrator.cpp
```diff
@@ -0,0 +1,172 @@
...
+    throw IntegrationFailedError();
+  }
+}
+
+//==============================================================================
+void VectorFieldIntegrator::check(const Eigen::VectorXd& q, double t)
+{
+  if (mTimer.getElapsedTime() > mTimelimit)
+  {
+    // throw TimeLimitError();
```
mkoval: This should not be commented out. 😉

src/planner/vectorfield/detail/VectorFieldIntegrator.cpp
```diff
@@ -0,0 +1,172 @@
...
+    }
+    if (!mVectorField->evaluateTrajectory(
+            *trajSegment.get(),
+            mConstraint,
+            mConstraintCheckResolution,
+            trajEvalStartTime))
+    {
+      throw VectorFieldTerminated("State not in constraint.");
+    }
+  }
```
mkoval: This constraint validation logic is correct, but will result in _a lot_ of extra constraint validation checks if the the integration step size is significantly smaller than the collision checking resolution. This is common if the robot is near a singularity because the ODE is relatively stiff.

A better way to implement this is to maintain the time `mLastEvaluation` at which the last constraint validation check was performed. In this function, we could collision check the trajectory from `mLastEvaluation` to `t` _excluding the endpoint_, then update `mLastEvaluation = t`. Finally, when planning terminates, we need to check the unevaluated portion `[mLastEvaluation, getEndTime()]` of the trajectory.

I think this is fine for now - this doesn't affect correctness. If you don't want to make this change now, then just file an issue so it doesn't slip through the cracks. 😉

src/planner/vectorfield/detail/VectorFieldIntegrator.cpp
```diff
@@ -0,0 +1,172 @@
...
+    segment.push_back(mKnots.back());
+    segment.push_back(knot);
+    auto trajSegment = convertToSpline(segment, mVectorField->getStateSpace());
+
+    if (mKnots.size() > 1)
+    {
+      trajEvalStartTime = mConstraintCheckResolution;
+    }
+    if (!mVectorField->evaluateTrajectory(
+            *trajSegment.get(),
```
mkoval: Nit: This is equivalent to `*trajSegment`.

src/planner/vectorfield/detail/VectorFieldIntegrator.cpp
```diff
@@ -0,0 +1,172 @@
...
+            mConstraint,
+            mConstraintCheckResolution,
+            trajEvalStartTime))
+    {
+      throw VectorFieldTerminated("State not in constraint.");
+    }
+  }
+  mKnots.push_back(knot);
+  mIndex += 1;
```
mkoval: It seems like `mIndex` always equals `mKnots.size()`. Is the variable necessary? 🤔

include/aikido/planner/vectorfield/MoveEndEffectorPoseVectorField.hpp
```diff
@@ -0,0 +1,78 @@
...
+  ///
+  /// \param[in] stateSpace MetaSkeleton state space.
+  /// \param[in] bn Body node of end-effector.
+  /// \param[in] goalPose Desired end-effector pose.
+  /// \param[in] poseErrorTolerance Constraint error tolerance in meters.
+  /// \param[in] r Conversion of radius to meters in computing Geodesic
+  /// distance.
+  /// \param[in] linearVelocityGain Linear velocity gain in workspace.
+  /// \param[in] angularVelocityGain Angular velocity gain in workspace.
+  /// \param[in] initialStepSize Initial step size.
```
mkoval: Clarify that this is the *maximum* step size used to guarantee that the integrator does not step out of joint limits, not the initial step size.

src/planner/vectorfield/detail/VectorFieldIntegrator.cpp
```diff
@@ -0,0 +1,172 @@
...
+    {
+      trajEvalStartTime = mConstraintCheckResolution;
+    }
+    if (!mVectorField->evaluateTrajectory(
+            *trajSegment.get(),
+            mConstraint,
+            mConstraintCheckResolution,
+            trajEvalStartTime))
+    {
+      throw VectorFieldTerminated("State not in constraint.");
```
mkoval: This should be a `ConstraintViolatedError`.

src/planner/vectorfield/detail/BodyNodePoseVectorFieldConstraint.hpp
dqyi11: I am not sure whether there is a standard on creating a `CompoundTestable`. I could not find any in `aikido`. But this class actually implements  a "compoundTestable" that wraps `stateConstraint` and the joint limit constraint returned by calling `createTestableBounds()` on the `MetaSkeletonStateSpace`.

src/planner/vectorfield/VectorFieldUtil.cpp
dqyi11: I remember that I had tested that setting `infinity()` to bounds will cause `invalid argument` error in calling `NlOpt solver`. The sample code I could find in `dart` uses `HUGE_VAL`. I think we could users decide what the bounds are like and set accordingly. I added back `enforceJointVelocityLimits` to enable/disable joint velocity limits here.

src/planner/vectorfield/MoveEndEffectorPoseVectorField.cpp
```diff
@@ -0,0 +1,64 @@
...
+}
+
+//==============================================================================
+bool MoveEndEffectorPoseVectorField::evaluateCartesianVelocity(
+    const Eigen::Isometry3d& pose, Eigen::Vector6d& cartesianVelocity) const
+{
+  using aikido::planner::vectorfield::computeGeodesicTwist;
+  Eigen::Vector6d desiredTwist = computeGeodesicTwist(pose, mGoalPose);
+  desiredTwist.head<3>() *= mAngularVelocityGain;
+  desiredTwist.tail<3>() *= mLinearVelocityGain;
```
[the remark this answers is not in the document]
dqyi11: I have removed these two parameters. The original intent of adding these two parameters is providing some tuning parameters of a resulting vector field.

src/planner/vectorfield/MoveEndEffectorOffsetVectorField.cpp
```diff
@@ -4,144 +4,98 @@
...
-      _stateSpace,
-      mBodynode,
-      mOptimizationTolerance,
-      mTimestep,
-      mPadding,
-      _dq);
-  return result;
+  Eigen::Vector6d desiredTwist = computeGeodesicTwist(pose, mStartPose);
+  desiredTwist.tail<3>() = mDirection;
+  cartesianVelocity = desiredTwist;
```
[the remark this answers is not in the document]
dqyi11: Yes. This is removed when moving from `muul` design to `prpy` design.
In `muul`, `MoveHandStraightVectorField` defines a timed workspace path by a distance and linear velocity, which is similar with the requirement of `PlanWorkspacePath()` in `prpy`, so linear gain and rotation gain are used.
Now `planToEndEffectorOffset` moves a range [minimum distance, maximum distance). It is not easy to estimate rotational error and linear orthogonal error, because no goal pose is known.
I guess that is the reason that no control gain is used in `planToEndEffectorOffset()` of `prpy`.

In a coming `PlanWorkspacePath` vector field, linear and rotation control gains are used in correcting the deviation from a reference workspace path. 

Does this make sense?

include/aikido/planner/vectorfield/BodyNodePoseVectorField.hpp
```diff
@@ -0,0 +1,124 @@
...
+  /// Meta skeleton state space.
+  aikido::statespace::dart::MetaSkeletonStateSpacePtr mMetaSkeletonStateSpace;
+
+  /// Meta Skeleton
+  dart::dynamics::MetaSkeletonPtr mMetaSkeleton;
+
+  /// BodyNode
+  dart::dynamics::BodyNodePtr mBodyNode;
+
+  /// Initial step size of integrator.
```
mkoval: Nit: This comment should be updated to reflect the fact that this is a maximum step size.

src/planner/vectorfield/MoveEndEffectorOffsetVectorField.cpp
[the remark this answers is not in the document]
mkoval: Yes, that makes sense. 👍 Thanks for the explanation.

src/planner/vectorfield/VectorFieldPlanner.cpp
```diff
@@ -1,278 +1,237 @@
...
+  catch (const detail::VectorFieldTerminated& e)
   {
-    std::stringstream ss;
-    if (skeleton->getPositionLowerLimit(i) > skeleton->getPositionUpperLimit(i))
-    {
-      ss << "Position lower limit is larger than upper limit at DOF " << i;
-      throw std::invalid_argument(ss.str());
-    }
-    if (skeleton->getVelocityLowerLimit(i) > skeleton->getVelocityUpperLimit(i))
+    // dtwarn << e.what() << std::endl;
```
mkoval: Remove dead code.

PR #277, opened 2017-12-08
[WIP] Introduce pr-ompl dependency

aditya-vk:
This PR takes the first step to enable using any of the planners developed at PRL. Specifically the planners that have an `OMPL` interface but those that are still not a part of `OMPL`. This PR is directly linked to https://github.com/personalrobotics/pr-ompl/pull/4

This is to ensure that planners that have been developed out of `AIKIDO` framework can still be be used via `AIKIDO`. `CRRT` and `CRRTConnect` come under this category which can be moved out of `AIKIDO` into `pr-ompl`.

Each of the planners `<P>` in `pr-ompl`, if the current framework in AIKIDO is followed, would have a corresponding `plan<P>()` method in AIKIDO which sets the planner-specific parameters and calls `planOMPL`. I would like advice on whether this is ideal or there can be a better restructuring to minimize the number of such individual functions (given that planners have their own specific parameters).

In this PR I implemented only LRA* -> graph-based planner that generalizes LazySP and LWA*. It needs a little clean up in the sense that some of the parameters being passed to it (constraints:: params etc.) are irrelevant to it. The test for it was written using a locally available graphml file. Since it currently takes graphml file as input, not sure how to test the function without creating a graph in place and saving it, as an alternative. The test has been commented out for now.

***


CMakeLists.txt
```diff
@@ -124,7 +124,7 @@ initialize_component_helpers(aikido)
 include(CMakePackageConfigHelpers)
 include(FindPkgConfig)
 
-find_package(Boost REQUIRED COMPONENTS filesystem)
+find_package(Boost REQUIRED COMPONENTS filesystem graph program_options system)
```
brianhou: I haven't had a chance to actually review this PR (or the corresponding PR in `pr-ompl`), but I think we should keep these kind of dependencies _out_ of AIKIDO if they're only needed in `pr-ompl`.

As I think you were saying offline, I think it'd be best if this was an optional dependency on `pr-ompl`, rather than requiring that `pr-ompl` be bundled with AIKIDO. It seems that having this organization where we add our methods directly to `planner/ompl/Planner.cpp` wouldn't quite work in that scenario? I feel like we'd need to have them in a separate `PRPlanner` file that we can choose to compile depending on whether `pr-ompl` is actually available, but maybe someone who's more familiar with CMake can weigh in.

aditya-vk: This dependency is not required in AIKIDO actually, my bad - removed them in a recent commit. I currently have `pr-ompl` as an optional dependency but yes, moving `pr-ompl` methods out of `planner/ompl/Planner.cpp` will allow `OMPL` and `pr-ompl` dependencies to be separate.

include/aikido/planner/ompl/Planner.hpp
```diff
@@ -209,6 +210,51 @@ trajectory::InterpolatedPtr planCRRTConnect(
...
+    statespace::StateSpacePtr _stateSpace,
+    statespace::InterpolatorPtr _interpolator,
+    distance::DistanceMetricPtr _dmetric,
+    constraint::SampleablePtr _sampler,
+    constraint::TestablePtr _validityConstraint,
+    constraint::TestablePtr _boundsConstraint,
+    constraint::ProjectablePtr _boundsProjector,
+    double _maxDistanceBtwProjections,
+    std::string _roadmapPath,
+    double _lookahead,
```
mkoval: I think this should be an `int`. 🤔 At least from my understanding of the paper.

include/aikido/planner/ompl/Planner.hpp
```diff
@@ -209,6 +210,51 @@ trajectory::InterpolatedPtr planCRRTConnect(
...
+    double _greediness,
```
mkoval: I think this also should be an `int`.

include/aikido/planner/ompl/Planner.hpp
```diff
@@ -209,6 +210,51 @@ trajectory::InterpolatedPtr planCRRTConnect(
...
+    const statespace::StateSpace::State* _goal,
+    statespace::StateSpacePtr _stateSpace,
+    statespace::InterpolatorPtr _interpolator,
+    distance::DistanceMetricPtr _dmetric,
+    constraint::SampleablePtr _sampler,
+    constraint::TestablePtr _validityConstraint,
+    constraint::TestablePtr _boundsConstraint,
+    constraint::ProjectablePtr _boundsProjector,
+    double _maxDistanceBtwProjections,
+    std::string _roadmapPath,
```
mkoval: Nit: Take this by `const &`.

include/aikido/planner/ompl/Planner.hpp
```diff
@@ -209,6 +210,51 @@ trajectory::InterpolatedPtr planCRRTConnect(
...
+/// start to a goal region given a graph to search on
+/// \param _start The start state
+/// \param _goal The goal state
+/// \param _statespace The StateSpace that the planner must plan within
+/// \param _interpolator An Interpolator defined on the StateSpace. This is used
+/// to interpolate between two points within the space.
+/// \param _dmetric A valid distance metric defined on the StateSpace
+/// \param _sampler A Sampleable that can sample states from the
+/// StateSpace. Warning: Many OMPL planners internally assume this sampler
+/// samples uniformly. Care should be taken when using a non-uniform sampler.
```
mkoval: This warning only makes sense for `planOMPL`. You should either reword to rephrase it specifically for LRA* or remove it entirely.

src/planner/ompl/Planner.cpp
```diff
@@ -168,7 +170,15 @@ trajectory::InterpolatedPtr planOMPL(
...
     double _maxPlanTime)
 {
-  _planner->setProblemDefinition(_pdef);
+  try
+  {
+    _planner->setProblemDefinition(_pdef);
+  }
+  catch(::ompl::Exception &ex)
+  {
+    std::cout << "Error" << std::endl;
```
mkoval: Don't print to `std::out`. if you want to return this error message, I suggest either wrapping it in an Aikido exception or putting it in a `PlannerResult` output parameter.

tests/planner/ompl/test_OMPLPlanner.cpp
```diff
@@ -1,3 +1,4 @@
+#include <LRAstar.hpp>
```
mkoval: Where does this header come from? It should probably in a directory, rather than in the root include path.

include/aikido/planner/ompl/Planner.hpp
```diff
@@ -209,6 +210,51 @@ trajectory::InterpolatedPtr planCRRTConnect(
...
+/// \param _maxDistanceBtwProjections The maximum distance (under dmetric)
+/// between projecting and validity checking two successive points on a tree
+/// extension
+/// \param _roadmapPath The path to the graph to be used
+/// \param _lookahead The number of edges over which lazy search is performed.
+/// The lookahead helps interpolate between LWA* and LazySP
+/// \param _greediness The number of edges along the lazy path to evaluate
+/// \param _maxPlanTime The maximum time to allow the planner to search for a
+/// solution
+trajectory::InterpolatedPtr planLRAstar(
```
mkoval: This function should probably be moved into a `planning::ompl::pr` namespace, so we can still build the `planning::ompl` module without requiring the user to install `pr_ompl`. This also provides a nice migration path as planners get merged into the core of OMPL: they can be promoted up a namespace.

PR #278, opened 2017-12-08
Refactor DART StateSpaces

brianhou:
This PR refactors AIKIDO's DART state spaces to decouple the state space from the underlying DART `Joint` or `MetaSkeleton`, as suggested by @mkoval and @psigen in #249.

**This will be a hugely backwards-incompatible API change. Any code that uses `MetaSkeletonStateSpace::getMetaSkeleton`, `JointStateSpace::getJoint`, or `MetaSkeletonStateSpaceSaver` will need to be fixed.**

Quick summary of changes:
- `JointStateSpace::getJoint` has been removed and `JointStateSpace::getProperties` has been added.
- `MetaSkeletonStateSpace::getMetaSkeleton` has been removed and `MetaSkeletonStateSpace::getProperties` has been added. `MetaSkeletonStateSpace::Properties` also holds information about how to map (`joint index`, `joint dof index`) to `metaskeleton dof index`, rather than re-computing each time we need to convert between states and positions.
- `MetaSkeletonStateSpaceSaver` has been renamed to `MetaSkeletonStateSaver`. I think this makes more sense, since it was most importantly used to save the current `MetaSkeleton` position (which would no longer be in a `MetaSkeletonStateSpace`).
- Many constraints (e.g. `constraint::CollisionFree`, `constraint::InverseKinematicsSampleable`) now take in an additional `MetaSkeleton` argument. This is necessary so they have a concrete skeleton that they can test with.
- Conversions between `SplineTrajectory` and ROS `JointTrajectory` are somewhat different. The joint names in the `JointTrajectory` now come from the DOF names, rather than the Joint names. I thought this was simpler, but could be convinced to figure out how to use Joint names instead if this is the wrong thing to do. (It seems that each joint in a ROS `JointTrajectory` can only take on one value, so maybe it's closer to what we mean by DOF?)

Note: one of the `VectorFieldPlanner` tests is currently failing, but I figured it's not really worth debugging since the improved `VectorFieldPlanner` will be ready soon and will need to be ported to the new `MetaSkeletonStateSpace` anyway.

***


include/aikido/statespace/dart/MetaSkeletonStateSpace.hpp
```diff
@@ -18,26 +19,96 @@ namespace dart {
...
+  };
+
   using CartesianProduct::State;
   using CartesianProduct::ScopedState;
 
   /// Constructs a state space for a DART \c MetaSkeleton.
   ///
   /// \param _metaskeleton target \c MetaSkeleton
-  MetaSkeletonStateSpace(::dart::dynamics::MetaSkeletonPtr _metaskeleton);
+  MetaSkeletonStateSpace(::dart::dynamics::MetaSkeleton* _metaskeleton);
```
mkoval: Mark this as `explicit`. It should have been before, too! 😱

include/aikido/statespace/dart/MetaSkeletonStateSpace.hpp
```diff
@@ -18,26 +19,96 @@ namespace dart {
...
 
-  /// Get the \c MetaSkeleton associated with this state space.
+  /// Gets the MetaSkeleton properties associated with this state space.
   ///
-  /// \return \c MetaSkeleton associated with this state space
-  ::dart::dynamics::MetaSkeletonPtr getMetaSkeleton() const;
+  /// \return MetaSkeleton properties associated with this state space
+  const Properties getProperties() const;
```
mkoval: Nit: `const` is not meaningful.

include/aikido/statespace/dart/MetaSkeletonStateSaver.hpp
```diff
@@ -0,0 +1,43 @@
...
+      ::dart::dynamics::MetaSkeletonPtr _metaskeleton);
+
+  virtual ~MetaSkeletonStateSaver();
+
+  // MetaSkeletonStateSaver is uncopyable, must use std::move
+  MetaSkeletonStateSaver(const MetaSkeletonStateSaver&) = delete;
+  MetaSkeletonStateSaver& operator=(const MetaSkeletonStateSaver&) = delete;
+
+  MetaSkeletonStateSaver(MetaSkeletonStateSaver&&) = default;
+  MetaSkeletonStateSaver& operator=(MetaSkeletonStateSaver&&) = default;
```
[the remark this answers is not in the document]
brianhou: I copied this from https://github.com/personalrobotics/aikido/blob/master/include/aikido/statespace/ScopedState.hpp#L30 when I initially added the MetaSkeletonStateSpaceSaver. I think this was to prevent copying in favor of moving? I'm not sure when we would want to do either though.

include/aikido/statespace/dart/MetaSkeletonStateSpace.hpp
```diff
@@ -18,26 +19,96 @@ namespace dart {
...
+    /// Number of joints in the MetaSkeleton
+    std::size_t mNumJoints;
+
+    /// Names of DOFs in the MetaSkeleton
+    std::vector<std::string> mDofNames;
+
+    /// Mapping from Joint index and Joint DOF index to MetaSkeleton DOF index
+    std::unordered_map<std::size_t,
+                       std::unordered_map<std::size_t, std::size_t>>
+        mIndexMap;
```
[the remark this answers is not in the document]
brianhou: This was because pairs don't come with a hash function and I was being sloppy... I'll fix it.

brianhou:
I've updated this PR with the new `VectorFieldPlanner` implementation, so there are no longer any conflicts with master. Can someone (@gilwoolee or @jslee02) take one last pass over this before I merge?

src/constraint/FramePairDifferentiable.cpp
```diff
@@ -7,14 +7,22 @@ namespace constraint {
...
     DifferentiablePtr _relPoseConstraint)
-  : mJacobianNode1(std::move(_jacobianNode1))
+  : mMetaSkeletonStateSpace(std::move(_metaSkeletonStateSpace))
+  , mMetaSkeleton(std::move(_metaskeleton))
+  , mJacobianNode1(std::move(_jacobianNode1))
   , mJacobianNode2(std::move(_jacobianNode2))
   , mRelPoseConstraint(std::move(_relPoseConstraint))
-  , mMetaSkeletonStateSpace(std::move(_metaSkeletonStateSpace))
 {
+  if (!mMetaSkeletonStateSpace)
```
gilwoolee: Check compatibility between `statespace` and `metaskeleton`.

src/planner/vectorfield/MoveEndEffectorOffsetVectorField.cpp
```diff
@@ -13,6 +13,7 @@ namespace vectorfield {
 //==============================================================================
 MoveEndEffectorOffsetVectorField::MoveEndEffectorOffsetVectorField(
     aikido::statespace::dart::MetaSkeletonStateSpacePtr stateSpace,
+    dart::dynamics::MetaSkeletonPtr metaskeleton,
```
gilwoolee: Check for compatibility between statespace and metaskeleton.

src/planner/vectorfield/VectorFieldPlanner.cpp
```diff
@@ -192,6 +195,7 @@ std::unique_ptr<aikido::trajectory::Spline> planToEndEffectorOffset(
 //==============================================================================
 std::unique_ptr<aikido::trajectory::Spline> planToEndEffectorPose(
     const aikido::statespace::dart::MetaSkeletonStateSpacePtr& stateSpace,
+    dart::dynamics::MetaSkeletonPtr metaskeleton,
```
gilwoolee: Check for compatibility.

brianhou:
@jslee02 Merging master (with the `TrajectoryMarker` introduced in #288) causes the build to fail. Can you investigate? Part of the issue is because `MetaSkeletonStateSpaceSaver` was changed to `MetaSkeletonStateSaver`. Also, it seems that the `MetaSkeleton` within `TrajectoryMarker` shouldn't be `const` since we change it in `TrajectoryMarker::updatePoints`.

Issue #284, opened 2017-12-09
Consider adding name field to Trajectory


jslee02:
I think it would be good to be able to assign a name to a trajectory. This would be useful to know which trajectory is which when visualizing multiple trajectories in a viewer (e.g., rviz). This can be done by adding a name field to the associated visualization object (e.g., `TrajectoryMarker`), but I think it would be worth to have them in `Trajectory` class.

Thought?

psigen:
In general, I am a little wary of assigning metadata like names to pure data structures like `Trajectory`.  It makes it a bit awkward in situations like:

1) A trajectory is being created and consumed inside an optimizer.  What should the name be?
2) A trajectory is being stored into a database by ID.  Does the name have to be synchronized with the database ID when it is saved/loaded/copied?  Can the name be a binary string (since that might be its unique identifier)?
3) A trajectory is cloned a number of times.  Should its name change?  Does the name need to be unique?

With that said, it can be user-friendly to keep the name around as a metadata field for debugging. If it is really necessary to maintain a name, absolutely _nothing_ in DART or its plugins should assume the field:
1) is populated
2) is unique
3) is meaningful

mkoval:
I agree with @psigen here. I generally like the idea of keeping name in visualization classes for all of the reasons he outlined above.

However, I do think it would be tremendously useful to other types of metadata with a trajectory, e.g. how it was planned, what constraints it was checked against, and where those constraint checks were performed. This is very similar to the concept of "trajectory tags" we had in PrPy that @brianhou began to resurrect in https://github.com/personalrobotics/aikido/pull/259.

---

This is _precisely_ the problem that DART `Aspect`s https://github.com/dartsim/dart/pull/659 were intended to solve. I would be thrilled if we added `Aspect`s to Aikido's `Trajectory` class. I don't see any issue with creating a `VisualizationAspect` that could hold the trajectory's display name.

PR #288, opened 2017-12-12
Add TrajectoryMarker

jslee02:
This PR adds `aikido::rviz::TrajectoryMarker` for visualizing trajectories in RViz.

The inputs are:
* `aikido::trajectory::Trajectory`: A C-space trajectory used to compute task-space trajectory
* `dart::dynamics::Skeleton`: Target skeleton the trajectory is applied to.
* `dart::dynamics::Frame`: Target frame to visualize the trajectory of the origin.

`aikido::rviz::TrajectoryMarker` has several rendering options:
* trajectory color (RBGA)
* trajectory thickness
* number of line-segments of the task-space trajectory

Example:
image

***

### DIFF
--- a/include/aikido/rviz/InteractiveMarkerViewer.hpp
+++ b/include/aikido/rviz/InteractiveMarkerViewer.hpp
@@ -13,6 +13,7 @@
 #include <aikido/constraint/TSR.hpp>
 #include <aikido/rviz/SmartPointers.hpp>
 #include <aikido/rviz/TSRMarker.hpp>
+#include <aikido/trajectory/Trajectory.hpp>
 
 namespace aikido {
 namespace rviz {
@@ -51,6 +52,28 @@ public:
       int nSamples = 10,
       const std::string& basename = "");
 
+  /// Adds trajectory marker to this viewer.
+  ///
+  /// \param[in] trajectory C-sapce (or joint-space) trajectory.
+  /// \param[in] skeleton Target DART meta skeleton that the C-space trajectory
+  /// will be applied to compute the visualizing task-space trajectory.
+  /// \param[in] frame Target DART frame where the trajectory of its origin
+  /// will be visualized.
+  /// \param[in] rgba Color and alpha of the visualizing task-space trajectory.
+  /// Default is [RGBA: 0.75, 0.75, 0.75, 0.75].
+  /// \param[in] thickness Thickness of the visualizing task-space trajectory.
+  /// Default is 0.01.
+  /// \param[in] numLineSegments Number of line-segments of the visualizing
+  /// task-space trajectory. Default is 16.
+  /// \return Trajectory marker added to this viewer.
+  TrajectoryMarkerPtr addTrajectoryMarker(
+      trajectory::ConstTrajectoryPtr trajectory,
+      const dart::dynamics::MetaSkeleton& skeleton,
+      const dart::dynamics::Frame& frame,
+      const Eigen::Vector4d& rgba = Eigen::Vector4d::Constant(0.75),
+      double thickness = 0.01,
+      std::size_t numLineSegments = 16u);
+
   void setAutoUpdate(bool flag);
   void update();
 
@@ -60,6 +83,7 @@ protected:
   interactive_markers::InteractiveMarkerServer mMarkerServer;
   std::set<SkeletonMarkerPtr> mSkeletonMarkers;
   std::set<FrameMarkerPtr> mFrameMarkers;
+  std::set<TrajectoryMarkerPtr> mTrajectoryMarkers;
 
   std::atomic_bool mRunning;
   std::atomic_bool mUpdating;
--- a/include/aikido/rviz/InteractiveMarkerViewer.hpp
+++ b/include/aikido/rviz/InteractiveMarkerViewer.hpp
@@ -5,6 +5,7 @@
 #include <set>
 #include <thread>
 
+#include <dart/common/NameManager.hpp>
 #include <dart/dynamics/Frame.hpp>
 #include <dart/dynamics/SmartPointer.hpp>
 #include <interactive_markers/interactive_marker_server.h>
@@ -84,6 +85,11 @@ protected:
   std::set<SkeletonMarkerPtr> mSkeletonMarkers;
   std::set<FrameMarkerPtr> mFrameMarkers;
   std::set<TrajectoryMarkerPtr> mTrajectoryMarkers;
+
+  /// NameManager for name uniqueness of trajectories in the same
+  /// InteractiveMarkerServer.
+  dart::common::NameManager<trajectory::ConstTrajectoryPtr>
+      mTrajectoryNameManager;
 
   std::atomic_bool mRunning;
   std::atomic_bool mUpdating;
--- a/include/aikido/rviz/InteractiveMarkerViewer.hpp
+++ b/include/aikido/rviz/InteractiveMarkerViewer.hpp
@@ -54,17 +54,17 @@ public:
 
   /// Adds trajectory marker to this viewer.
   ///
-  /// \param[in] trajectory C-sapce (or joint-space) trajectory.
+  /// \param[in] trajectory C-space (or joint-space) trajectory.
   /// \param[in] skeleton Target DART meta skeleton that the C-space trajectory
   /// will be applied to compute the visualizing task-space trajectory.
   /// \param[in] frame Target DART frame where the trajectory of its origin
   /// will be visualized.
-  /// \param[in] rgba Color and alpha of the visualizing task-space trajectory.
-  /// Default is [RGBA: 0.75, 0.75, 0.75, 0.75].
-  /// \param[in] thickness Thickness of the visualizing task-space trajectory.
-  /// Default is 0.01.
-  /// \param[in] numLineSegments Number of line-segments of the visualizing
-  /// task-space trajectory. Default is 16.
+  /// \param[in] rgba Color and alpha of the visualized trajectory. Default is
+  /// [RGBA: 0.75, 0.75, 0.75, 0.75].
+  /// \param[in] thickness Thickness of the visualized trajectory. Default is
+  /// 0.01.
+  /// \param[in] numLineSegments Number of line segments in the visualized
+  /// trajectory. Default is 16.
   /// \return Trajectory marker added to this viewer.
   TrajectoryMarkerPtr addTrajectoryMarker(
       trajectory::ConstTrajectoryPtr trajectory,
--- a/include/aikido/rviz/TrajectoryMarker.hpp
+++ b/include/aikido/rviz/TrajectoryMarker.hpp
@@ -17,17 +17,17 @@ public:
   /// Constructor.
   /// \param[in] markerServer RViz marker server.
   /// \param[in] frameId RViz frame ID.
-  /// \param[in] trajectory C-sapce (or joint-space) trajectory.
-  /// \param[in] skeleton Target DART meta skeleton that the C-space trajectory
-  /// will be applied to compute the visualizing task-space trajectory.
-  /// \param[in] frame Target DART frame where the trajectory of its origin
-  /// will be visualized.
-  /// \param[in] rgba Color and alpha of the visualizing task-space trajectory.
-  /// Default is [RGBA: 0.75, 0.75, 0.75, 0.75].
-  /// \param[in] thickness Thickness of the visualizing task-space trajectory.
-  /// Default is 0.01.
-  /// \param[in] numLineSegments Number of line-segments of the visualizing
-  /// task-space trajectory. Default is 16.
+  /// \param[in] trajectory C-space (or joint-space) trajectory.
+  /// \param[in] skeleton DART meta skeleton for visualizing the trajectory in
+  /// task space.
+  /// \param[in] frame DART frame for visualizing the trajectory in task space.
+  /// The trajectory of the origin will be visualized.
+  /// \param[in] rgba Color and alpha of the visualized trajectory. Default is
+  /// [RGBA: 0.75, 0.75, 0.75, 0.75].
+  /// \param[in] thickness Thickness of the visualized trajectory. Default is
+  /// 0.01.
+  /// \param[in] numLineSegments Number of line segments in the visualized
+  /// trajectory. Default is 16.
   TrajectoryMarker(
       interactive_markers::InteractiveMarkerServer* markerServer,
       const std::string& frameId,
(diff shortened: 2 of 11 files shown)


include/aikido/rviz/InteractiveMarkerViewer.hpp
```diff
@@ -51,6 +52,28 @@ class InteractiveMarkerViewer
       int nSamples = 10,
       const std::string& basename = "");
 
+  /// Adds trajectory marker to this viewer.
+  ///
+  /// \param[in] trajectory C-sapce (or joint-space) trajectory.
```
brianhou: Typo: sapce -> space

include/aikido/rviz/InteractiveMarkerViewer.hpp
```diff
@@ -51,6 +52,28 @@ class InteractiveMarkerViewer
...
+  /// \param[in] skeleton Target DART meta skeleton that the C-space trajectory
+  /// will be applied to compute the visualizing task-space trajectory.
+  /// \param[in] frame Target DART frame where the trajectory of its origin
+  /// will be visualized.
+  /// \param[in] rgba Color and alpha of the visualizing task-space trajectory.
```
brianhou: Nit: visualizing -> visualized. I think you can take out task-space here.

include/aikido/rviz/InteractiveMarkerViewer.hpp
```diff
@@ -51,6 +52,28 @@ class InteractiveMarkerViewer
...
+  /// Default is [RGBA: 0.75, 0.75, 0.75, 0.75].
+  /// \param[in] thickness Thickness of the visualizing task-space trajectory.
```
brianhou: Same as above.

include/aikido/rviz/InteractiveMarkerViewer.hpp
```diff
@@ -51,6 +52,28 @@ class InteractiveMarkerViewer
...
+  /// Default is 0.01.
+  /// \param[in] numLineSegments Number of line-segments of the visualizing
```
brianhou: Same as above.

Nit: "line-segments of" -> "line segments in"

include/aikido/rviz/TrajectoryMarker.hpp
```diff
@@ -0,0 +1,137 @@
...
+
+  ///
+  std::string mFrameId;
+
+  /// C-sapce (or joint-space) trajectory.
+  trajectory::ConstTrajectoryPtr mTrajectory;
+
+  /// Number of line segments of the discretized task-space trajectory.
+  std::size_t mNumLineSegments;
+  // TODO(JS): Consider switching to resolution once arc-length calculation is
```
brianhou: What if we use a time resolution (which exists) rather than an arc-length resolution (which I guess doesn't yet exist)? I think the number of line segments is unintuitive.

include/aikido/rviz/TrajectoryMarker.hpp
```diff
@@ -0,0 +1,137 @@
...
+
+/// A wrapper class of RViz InteractiveMarker for visualizing AIKIDO trajectory
+/// in RViz.
+class TrajectoryMarker final
+{
+public:
+  /// Constructor.
+  /// \param[in] markerServer RViz marker server.
+  /// \param[in] frameId RViz frame ID.
+  /// \param[in] trajectory C-sapce (or joint-space) trajectory.
```
brianhou: Typo: sapce -> space

include/aikido/rviz/TrajectoryMarker.hpp
```diff
@@ -0,0 +1,137 @@
...
+  /// \param[in] skeleton Target DART meta skeleton that the C-space trajectory
```
brianhou: I found this description somewhat confusing. What do you think of "DART meta skeleton for visualizing the trajectory in task space"?

include/aikido/rviz/TrajectoryMarker.hpp
```diff
@@ -0,0 +1,137 @@
...
+  /// will be applied to compute the visualizing task-space trajectory.
+  /// \param[in] frame Target DART frame where the trajectory of its origin
```
brianhou: I found this description somewhat confusing. What do you think of "DART frame for visualizing the trajectory in task space. The trajectory of the origin will be visualized."?

include/aikido/rviz/TrajectoryMarker.hpp
```diff
@@ -0,0 +1,137 @@
...
+  /// will be visualized.
+  /// \param[in] rgba Color and alpha of the visualizing task-space trajectory.
```
brianhou: Nit: visualizing -> visualized. I think you can take out task-space here.

include/aikido/rviz/TrajectoryMarker.hpp
```diff
@@ -0,0 +1,137 @@
...
+  /// Default is [RGBA: 0.75, 0.75, 0.75, 0.75].
+  /// \param[in] thickness Thickness of the visualizing task-space trajectory.
```
brianhou: Nit: visualizing -> visualized. I think you can take out task-space here.

include/aikido/rviz/TrajectoryMarker.hpp
```diff
@@ -0,0 +1,137 @@
...
+  /// Default is 0.01.
+  /// \param[in] numLineSegments Number of line-segments of the visualizing
```
brianhou: I think `numLineSegments` is a somewhat unintuitive way to discretize the trajectory. More comments below.

include/aikido/rviz/TrajectoryMarker.hpp
```diff
@@ -0,0 +1,137 @@
...
+  /// invalid_argument exception.
+  void setTrajectory(trajectory::ConstTrajectoryPtr trajectory);
+
+  /// Returns trajectory associated with this marker
+  trajectory::ConstTrajectoryPtr getTrajectory() const;
+
+  /// Sets or updates color (RGB) of visualizing trajectory.
+  void setColor(const Eigen::Vector3d& rgb);
+
+  /// Returns color (RGB) of visualizing trajectory.
```
brianhou: Nit: visualizing -> visualized.

include/aikido/rviz/TrajectoryMarker.hpp
```diff
@@ -0,0 +1,137 @@
...
+  ///
+  /// \param[in] trajectory C-sapce (or joint-space) trajectory. The statespace
+  /// of the trajectory should be MetaSkeletonStateSpace. Otherwise, throw
+  /// invalid_argument exception.
+  void setTrajectory(trajectory::ConstTrajectoryPtr trajectory);
+
+  /// Returns trajectory associated with this marker
+  trajectory::ConstTrajectoryPtr getTrajectory() const;
+
+  /// Sets or updates color (RGB) of visualizing trajectory.
```
brianhou: Nit: visualizing -> visualized.

include/aikido/rviz/TrajectoryMarker.hpp
```diff
@@ -0,0 +1,137 @@
...
+  void setColor(const Eigen::Vector3d& rgb);
+
+  /// Returns color (RGB) of visualizing trajectory.
+  Eigen::Vector3d getColor() const;
+
+  /// Sets or updates alpha of visualizing trajectory.
```
brianhou: Nit: visualizing -> visualized.

include/aikido/rviz/TrajectoryMarker.hpp
```diff
@@ -0,0 +1,137 @@
...
+  void setAlpha(double alpha);
+
+  /// Returns alpha of visualizing trajectory.
```
brianhou: Nit: visualizing -> visualized.

include/aikido/rviz/TrajectoryMarker.hpp
```diff
@@ -0,0 +1,137 @@
...
+  double getAlpha() const;
+
+  /// Sets or updates RGBA of visualizing trajectory.
```
brianhou: Nit: visualizing -> visualized.

include/aikido/rviz/TrajectoryMarker.hpp
```diff
@@ -0,0 +1,137 @@
...
+  void setRGBA(const Eigen::Vector4d& rgb);
+
+  /// Returns RGBA of visualizing trajectory.
```
brianhou: Nit: visualizing -> visualized.

include/aikido/rviz/TrajectoryMarker.hpp
```diff
@@ -0,0 +1,137 @@
...
+  Eigen::Vector4d getRBGA() const;
+
+  /// Sets or updates thickness of visualizing trajectory.
```
brianhou: Nit: visualizing -> visualized.

include/aikido/rviz/TrajectoryMarker.hpp
```diff
@@ -0,0 +1,137 @@
...
+  void setThickness(double thickness);
+
+  /// Returns thickness of visualizing trajectory.
```
brianhou: Nit: visualizing -> visualized.

include/aikido/rviz/TrajectoryMarker.hpp
```diff
@@ -0,0 +1,137 @@
...
+  /// Returns const marker.
+  const visualization_msgs::Marker& getMarker() const;
+
+  /// RViz marker server.
+  interactive_markers::InteractiveMarkerServer* mMarkerServer;
+
+  /// RViz interactive marker.
+  visualization_msgs::InteractiveMarker mInteractiveMarker;
+
+  ///
```
brianhou: Missing documentation?

include/aikido/rviz/TrajectoryMarker.hpp
```diff
@@ -0,0 +1,137 @@
...
+  std::string mFrameId;
+
+  /// C-sapce (or joint-space) trajectory.
```
brianhou: Typo: sapce -> space

src/rviz/TrajectoryMarker.cpp
```diff
@@ -0,0 +1,259 @@
...
+  , mNumLineSegments()
+  , mSkeleton(targetSkeleton)
+  , mFrame(frame)
+  , mNeedUpdate(true)
+  , mNeedPointsUpdate(true)
+{
+  using visualization_msgs::InteractiveMarkerControl;
+  using visualization_msgs::Marker;
+
+  // TODO(JS): Check the compatibility of statespaces of the skeleton and the
```
brianhou: Shouldn't that only need to happen in `setTrajectory`?

src/rviz/TrajectoryMarker.cpp
```diff
@@ -0,0 +1,259 @@
...
+  setTrajectory(std::move(trajectory));
+  setRGBA(rgba);
+  setThickness(thickness);
+  setNumLineSegments(numLineSegments);
+}
+
+//==============================================================================
+TrajectoryMarker::~TrajectoryMarker()
+{
+  mMarkerServer->erase(mInteractiveMarker.name);
```
brianhou: It looks like this isn't going to be unique based on what happens above in the constructor. Will that be a problem?

src/rviz/TrajectoryMarker.cpp
```diff
@@ -0,0 +1,259 @@
...
+
+//==============================================================================
+void TrajectoryMarker::setTrajectory(trajectory::ConstTrajectoryPtr trajectory)
+{
+  using visualization_msgs::Marker;
+
+  if (trajectory)
+  {
+    auto statespace = trajectory->getStateSpace();
+    if (!std::dynamic_pointer_cast<statespace::dart::MetaSkeletonStateSpace>(
```
brianhou: Maybe we can add a method in `MetaSkeletonStateSpace` to actually check whether a `MetaSkeleton` is compatible. Then we can use that check as well.

src/rviz/TrajectoryMarker.cpp
```diff
@@ -0,0 +1,259 @@
...
+  return marker.scale.x;
+}
+
+//==============================================================================
+void TrajectoryMarker::setNumLineSegments(std::size_t numLineSegments)
+{
+  if (numLineSegments < 2)
+  {
+    throw std::invalid_argument(
+        "Number of line segments should be greater than 2");
```
brianhou: This error message doesn't match the check -- did you mean "greater than or equal to 2"?

Either way, I'm not sure if this should be an actual error that we throw since it only means that the visualized trajectory won't look very good.

src/rviz/TrajectoryMarker.cpp
```diff
@@ -0,0 +1,259 @@
...
+    return;
+
+  auto& marker = getMarker();
+  auto& points = marker.points;
+  points.clear();
+
+  if (!mTrajectory)
+    return;
+
+  Eigen::VectorXd savedPositions = mSkeleton.getPositions();
```
brianhou: Is there a reason for using this over `MetaSkeletonSaver`?

include/aikido/rviz/TrajectoryMarker.hpp
```diff
@@ -0,0 +1,137 @@
...
+      const dart::dynamics::MetaSkeleton& skeleton,
+      const dart::dynamics::Frame& frame,
+      const Eigen::Vector4d& rgba = Eigen::Vector4d::Constant(0.75),
+      double thickness = 0.01,
+      std::size_t numLineSegments = 16u);
+
+  /// Destructor
+  ~TrajectoryMarker();
+
+  /// Sets or updates trajectory to visualize
```
brianhou: From reading the code, it seems that passing in `nullptr` will clear the current trajectory. Maybe we can explain that here as well?

src/rviz/TrajectoryMarker.cpp
jslee02: Good catch! I'm thinking of using DART name manager because AIKIDO trajectory currently doesn't have name property (related issue #284).

src/rviz/TrajectoryMarker.cpp
jslee02: That's because `MetaSkeletonSaver` saves more than positions, but I'm open to using `MetaSkeletonSaver`.

src/rviz/TrajectoryMarker.cpp
jslee02: Treating this as not an error is fine to me. In that case, I would like to at least (1) print a warning so that the user doesn't get confused why the trajectory is not visualized and (2) throw exception when `numLineSegments` is negative.

src/rviz/TrajectoryMarker.cpp
jslee02: Sounds good to me. As the check is not available yet, I'm just checking the dimensions and adding a comment suggesting using more comprehensive check once available.

include/aikido/rviz/TrajectoryMarker.hpp
jslee02: I would prefer discretizing a smooth curve based on its length so that we don't see giant straight lines for a long curve (The trajectory length computation will be introduced in implementing TrajOpt).

Using line segments might not be the best alternative for now but I believe this trick is widely used in practice. For example, GLU functions use similar parameters to draw a sphere, a cylinder, etc.

I actually would like to represent time resolution in a different visualization way (e.g., using different colors or alpha per seconds) as a future improvement.

src/rviz/TrajectoryMarker.cpp
brianhou: I think the old OpenRAVE state saver allowed you to pass in flags to that specified what should be saved. Can you add an issue for supporting that in `MetaSkeletonStateSaver`, and use `MetaSkeletonStateSaver` here?

jslee02: Done (https://github.com/personalrobotics/aikido/issues/293).

src/rviz/TrajectoryMarker.cpp
jslee02: https://github.com/personalrobotics/aikido/issues/292

PR #291, opened 2018-01-06
Enable CI tests on macOS

jslee02:


jslee02:
`mumps`, which is a transient dependency of DART, [fails to build on Travis CI](https://travis-ci.org/personalrobotics/aikido/jobs/327332983#L629-L650) (but it builds on my macbook). Will keep investigating on this.

jslee02:
Let me close this and open once all the issues are resolved.

PR #299, opened 2018-01-12
Fix CollisionGroup bugs in Hand executors.

### LINKED ISSUE #271: BarrettHand::execute should permit dynamically setting the collision groups
`BarrettHandKinematicSimulationPositionCommandExecutor` sets the collision group when the executor is constructed. This makes it difficult to use because (1) environment collision groups may not yet exist when the executor is created and (2) we may want to collision check against different groups at different times. One possible fix is to just pass in the collision group to `execute`, in addition to the desired goal position.

brianhou:
`BarrettHandKinematicSimulationPositionCommandExecutor::setCollideWith` didn't call `setCollideWith` on the underlying position or spread executors. In addition, calling `BarrettFingerKinematicSimulationPositionCommandExecutor::setCollideWith` and `BarrettFingerKinematicSimulationSpreadCommandExecutor::setCollideWith` could result in a mismatch if the new `CollisionGroup` came from a different `CollisionDetector`.

This PR fixes the first issue and adds a `setFingerCollisionGroup` method to both finger executors, which updates the appropriate finger collision groups (if necessary).

I think this resolves #271. (I didn't see the `setCollideWith` methods when I created the issue earlier.)

***

### DIFF
--- a/include/aikido/control/BarrettFingerKinematicSimulationPositionCommandExecutor.hpp
+++ b/include/aikido/control/BarrettFingerKinematicSimulationPositionCommandExecutor.hpp
@@ -81,6 +81,9 @@ public:
   bool setCollideWith(::dart::collision::CollisionGroupPtr collideWith);
 
 private:
+  /// Creates a CollisionGroup in the CollisionDetector for the fingers.
+  void setFingerCollisionGroup();
+
   constexpr static double kMimicRatio = 0.333;
   // TODO: read velocity limit from herb_description
   constexpr static double kProximalSpeed = 2.0;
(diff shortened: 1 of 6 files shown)


brianhou:
I re-enabled the `BarrettHand` tests by resolving the API changes from #166. Unfortunately, they all fail.

I'm not sure if this is another instance of https://github.com/personalrobotics/libherb/issues/43, but we should look into this soon.

brianhou:
Bugs fixed, thanks @gilwoolee!

I mis-remembered what the preshapes should actually look like, so this resolves https://github.com/personalrobotics/libherb/issues/43 as well.

Issue #213, opened 2017-05-26
Implement a wrapper for MinTOS trajectory smoothing


mkoval:
I have a mostly-implemented Aikido interface for MinTOS in the [`feature/MinTOSComponent` branch](https://github.com/personalrobotics/aikido/tree/feature/MinTOSComponent). MinTOS is a constrained trajectory smoothing and retiming package developed by Kris Hauser. See [this webpage](http://motion.pratt.duke.edu/software.html) or the [associated paper](http://www.roboticsproceedings.org/rss09/p22.pdf) for more information.

The current implementation in Aikido largely works. Here is [a video](https://youtu.be/if_wEJhjSWI) I made ages ago of executing a few trajectories optimized by MinTOS on HERB.

The next steps are to:

1. Update the `feature/MinTOSComponent` to the latest` master`, which likely has some conflicts.
2. Fix anything unsatisfactory about the current implementation.
3. Finish any missing unit tests.
4. Add a note about the license of the MinTOS Aikido module. (see below)

MinTOS links against GLPK, which is licensed under GPLv3 (ugh). I was careful to keep the Aikido code that links against MinTOS isolated into a single `aikido-planner-mintos` component. As a result, the library we build from `aikido-planner-mintos` will inherit that GPLv3 license. In turn, any other projects that depend on `aikido-planner-mintos` also inherit GPLv3.

In the longer run, we should consider modifying MinTOS to use a different optimization backend that is not as encumbered. However, there is no harm in merging the current version as it stands.

PR #300, opened 2018-01-14
Move Trajectory PostProcessors from Magi into AIKIDO.

sniyaz:
Move the smoothing and timing postprocessing classes into AIKIDO. Some changes from the MAGI versions:

1) The postprocessing classes now take a StatespacePtr in their constructors, instead of a MetaskeletonStatespacePtr. This was done since nothing about the parabolic timing/smoothing methods relies on this.

2) Simplify trajectory method from process.cpp has been removed.

3) Methods in the MAGI version used to post-process a Spline trajectory have been removed. This is because of a planned PR by @dqyi11 that will modify the VFP to return an InterpolatedPtr, making these methods unnecessary. 

4) A new test has been added that tries to smooth a trajectory and ensures the SmoothTrajectoryPostProcessor does something sane.

**TODO:**

1) Clean up docstrings before merging

2) I noticed that both PostProcessor classes don't actually use the statespace argument to the constructor. Is there a reason this is being passed, or is this something that can be safely removed?


include/aikido/planner/postprocessor/SmoothTrajectoryPostProcessor.hpp
```diff
@@ -0,0 +1,60 @@
...
+  /// \param _blendIterations Blend iterations for blending. It is
+  /// ineffective when _enableBlend is false.
+  SmoothTrajectoryPostProcessor(
+      aikido::statespace::StateSpacePtr _space,
+      double _smootherFeasibilityCheckResolution,
+      double _smootherFeasibilityApproxTolerance,
+      const Eigen::VectorXd& _velocityLimits,
+      const Eigen::VectorXd& _accelerationLimits,
+      const aikido::constraint::TestablePtr& _collisionTestable,
+      bool _enableShortcut,
```
gilwoolee: Can we put default values for these?

src/planner/postprocessor/postprocess.cpp
```diff
@@ -0,0 +1,99 @@
+#include <dart/common/StlHelpers.hpp>
+#include <aikido/planner/parabolic/ParabolicSmoother.hpp>
+#include <aikido/planner/parabolic/ParabolicTimer.hpp>
+#include <aikido/planner/postprocessor/postprocess.hpp>
```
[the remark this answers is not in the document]
brianhou: Do you think this is something that we can get clang-format to automatically fix? I know I've been pretty bad at this also...

include/aikido/trajectory.hpp
```diff
@@ -1,3 +1,7 @@
 #include "trajectory/Interpolated.hpp"
+#include "trajectory/RetimeTrajectoryPostProcessor.hpp"
```
brianhou: Do these header files exist? I think everything should be in `planner` now.

brianhou:
I'm not sure if there's a good reason for `timeTrajectory` and `smoothTrajectory` being explicit functions rather than methods in the corresponding class. I think this is just because that's how we were initially using all this code from libherb and it was easier this way.

I also think that those names (and therefore, the names of the classes in this PR) might be misleading.`RetimeTrajectoryPostProcessor` and `SmoothTrajectoryPostProcessor` don't feel descriptive enough to me, especially once when we have other smoothing post-processors in the future (e.g. #213).

@jslee02, what do you think about this proposal:
- Move the `timeTrajectory` and `smoothTrajectory` functions from their own helper file into the appropriate class
- Rename `RetimeTrajectoryPostProcessor` to `ParabolicRetimer`
- Rename `SmoothTrajectoryPostProcessor` to `ParabolicSmoother`

src/planner/postprocessor/postprocess.cpp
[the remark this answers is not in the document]
jslee02: Unfortunately, that option is not available in clang-format 3.8. :disappointed:

jslee02:
:+1: That proposal all sounds good to me. 
> * Move the `timeTrajectory` and `smoothTrajectory` functions from their own helper file into the appropriate class
> * Rename `RetimeTrajectoryPostProcessor` to `ParabolicRetimer`
> * Rename `SmoothTrajectoryPostProcessor` to `ParabolicSmoother`

src/planner/postprocessor/postprocess.cpp
```diff
@@ -0,0 +1,99 @@
...
+  using aikido::planner::parabolic::doShortcutAndBlend;
+
+  // Get timed trajectory for arm
+  auto timedTrajectory = computeParabolicTiming(
+      *_inputTraj, _velocityLimits, _accelerationLimits);
+
+  if (_enableShortcut && _enableBlend)
+  {
+    return doShortcutAndBlend(
+        *timedTrajectory,
```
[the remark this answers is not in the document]
sniyaz: After looking through the code, it appears this isn't possible.

sniyaz:
Thanks for your feedback everybody! 😄  I've addressed the majority of comments. These are the last things to work out:

@jslee02 I was attempting to mirror the structure from the `parabolic` and `vectorfield` subdirectories. I'm happy to move it into the main `planner` folder- I'm just worried things will become a little inconsistent.

@brianhou Regarding this:

> * Rename `RetimeTrajectoryPostProcessor` to `ParabolicRetimer`
> * Rename `SmoothTrajectoryPostProcessor` to `ParabolicSmoother`

It looks like files named `ParabolicSmoother` and `ParabolicTimer` already exist in `planner/parabolic`, so this could be confusing. How about `ParabolicSmootherPostProcessor` and `ParabolicRetimerPostProcessor`?

Also, re-pinging me earlier TODO question:
> * I noticed that both PostProcessor classes don't actually use the statespace argument to the constructor. Is there a reason this is being passed, or is this something that can be safely removed?

brianhou:
> It looks like files named `ParabolicSmoother` and `ParabolicTimer` already exist in `planner/parabolic`, so this could be confusing. How about `ParabolicSmootherPostProcessor` and `ParabolicRetimerPostProcessor`?

This is a bit of a drastic change, but what if we remove `planner/postprocessor`, move the class-based post-processor implementations to `planner/parabolic` (into the `ParabolicSmoother` and `ParabolicTimer` files) and move the `TrajectoryPostProcessor` header file into the root of `planner`?

The `planner::parabolic` namespace only exists for the two post-processing methods. I think the classes don't add that much more to the convenience functions that already exist (aside from the unified `TrajectoryPostProcessor` interface), so it makes sense to me to keep them there as well. This also conveniently fixes our convention of having classes named `X` in `X.cpp` files.

> I noticed that both PostProcessor classes don't actually use the statespace argument to the constructor. Is there a reason this is being passed, or is this something that can be safely removed?

I think we take the state space directly from the trajectory, so it seems like we can remove this. This would match the underlying convenience functions as well.

sniyaz:
@brianhou That proposal sounds fine to me! Getting on that now.

sniyaz:
@brianhou @jslee02 Things should be close to final now. Could you give feedback/rubber stamp when you have a chance?

include/aikido/planner/parabolic/ParabolicSmoother.hpp
```diff
@@ -129,6 +130,60 @@ std::unique_ptr<trajectory::Spline> doShortcutAndBlend(
...
+      int _blendIterations = DEFAULT_BLEND_ITERATIONS,
+      double _smootherFeasibilityCheckResolution = DEFAULT_CHECK_RESOLUTION,
+      double _smootherFeasibilityApproxTolerance = DEFAULT_TOLERANCE);
+
+  // Documentation inherited.
+  std::unique_ptr<aikido::trajectory::Spline> postprocess(
+      const aikido::trajectory::InterpolatedPtr& _inputTraj,
+      const aikido::common::RNG* _rng) override;
+
+private:
```
brianhou: Please document these members. Also, I don't think "smoother" is necessary in `mSmootherFeasibilityCheckResolution` and `mSmootherFeasibilityApproxTolerance` either.

include/aikido/planner/parabolic/ParabolicSmoother.hpp
```diff
@@ -129,6 +130,60 @@ std::unique_ptr<trajectory::Spline> doShortcutAndBlend(
...
+  double mSmootherFeasibilityApproxTolerance;
+  const Eigen::VectorXd mVelocityLimits;
+  const Eigen::VectorXd mAccelerationLimits;
+  aikido::constraint::TestablePtr mCollisionTestable;
+
+  bool mEnableShortcut;
+  bool mEnableBlend;
+  double mShortcutTimelimit;
+  double mBlendRadius;
+  int mBlendIterations;
```
jslee02: Could we have docstring for these members as well?

src/planner/parabolic/ParabolicSmoother.cpp
```diff
@@ -3,6 +3,7 @@
 #include <dart/common/StlHelpers.hpp>
 #include <aikido/common/Spline.hpp>
 #include <aikido/planner/parabolic/ParabolicSmoother.hpp>
+#include <aikido/planner/parabolic/ParabolicTimer.hpp>
 #include "DynamicPath.h"
 #include "HauserParabolicSmootherHelpers.hpp"
 #include "ParabolicUtil.hpp"
```
jslee02: Nit: 
```cpp
#include "aikido/common/Spline.hpp"
#include "aikido/planner/parabolic/ParabolicSmoother.hpp"
#include "aikido/planner/parabolic/ParabolicTimer.hpp"
#include "aikido/planner/parabolic/DynamicPath.h"
#include "aikido/planner/parabolic/HauserParabolicSmootherHelpers.hpp"
#include "aikido/planner/parabolic/ParabolicUtil.hpp"
```

src/planner/parabolic/ParabolicSmoother.cpp
```diff
@@ -111,6 +112,90 @@ std::unique_ptr<trajectory::Spline> doShortcutAndBlend(
   return outputTrajectory;
 }
 
+SmoothTrajectoryPostProcessor::SmoothTrajectoryPostProcessor(
```
jslee02: Could you add function definition separator here?

src/planner/parabolic/ParabolicSmoother.cpp
```diff
@@ -111,6 +112,90 @@ std::unique_ptr<trajectory::Spline> doShortcutAndBlend(
...
+  , mEnableShortcut{_enableShortcut}
+  , mEnableBlend{_enableBlend}
+  , mShortcutTimelimit{_shortcutTimelimit}
+  , mBlendRadius{_blendRadius}
+  , mBlendIterations{_blendIterations}
+{
+  // Do nothing
+}
+
+std::unique_ptr<aikido::trajectory::Spline>
```
jslee02: Could you add function definition separator here?

src/planner/parabolic/ParabolicTimer.cpp
```diff
@@ -113,6 +113,28 @@ std::unique_ptr<aikido::trajectory::Spline> computeParabolicTiming(
   return outputTrajectory;
 }
 
+RetimeTrajectoryPostProcessor::RetimeTrajectoryPostProcessor(
```
jslee02: Could you add function definition separator here?

src/planner/parabolic/ParabolicTimer.cpp
```diff
@@ -113,6 +113,28 @@ std::unique_ptr<aikido::trajectory::Spline> computeParabolicTiming(
...
+    const Eigen::VectorXd& _velocityLimits,
+    const Eigen::VectorXd& _accelerationLimits)
+  : mVelocityLimits{_velocityLimits}, mAccelerationLimits{_accelerationLimits}
+{
+  // Do nothing
+}
+
+std::unique_ptr<aikido::trajectory::Spline>
```
jslee02: Could you add function definition separator here?

include/aikido/planner/parabolic/ParabolicSmoother.hpp
```diff
@@ -129,6 +130,60 @@ std::unique_ptr<trajectory::Spline> doShortcutAndBlend(
     double _checkResolution = DEFAULT_CHECK_RESOLUTION,
     double _tolerance = DEFAULT_TOLERANCE);
 
+/// Class for performing parabolic smoothing on trajectories
+class SmoothTrajectoryPostProcessor
+    : public aikido::planner::TrajectoryPostProcessor
+{
+public:
+  /// \param _velocityLimits maximum velocity for each dimension.
```
brianhou: Capitalize `maximum` for both `_velocityLimits` and `_accelerationLimits`.

include/aikido/planner/parabolic/ParabolicTimer.hpp
```diff
@@ -43,6 +44,28 @@ std::unique_ptr<aikido::trajectory::Spline> computeParabolicTiming(
 std::unique_ptr<aikido::trajectory::Spline> convertToSpline(
     const aikido::trajectory::Interpolated& _inputTrajectory);
 
+/// Class for performing parabolic retiming on trajectories.
+class RetimeTrajectoryPostProcessor
+    : public aikido::planner::TrajectoryPostProcessor
+{
+public:
+  /// \param _velocityLimits maximum velocity for each dimension.
```
brianhou: Capitalize `maximum` for both `_velocityLimits` and `_accelerationLimits`.

include/aikido/planner/parabolic/ParabolicTimer.hpp
```diff
@@ -43,6 +44,28 @@ std::unique_ptr<aikido::trajectory::Spline> computeParabolicTiming(
 std::unique_ptr<aikido::trajectory::Spline> convertToSpline(
     const aikido::trajectory::Interpolated& _inputTrajectory);
 
+/// Class for performing parabolic retiming on trajectories.
+class RetimeTrajectoryPostProcessor
```
brianhou: I'd still like to rename this `ParabolicTimer` (and I think it makes more sense now that it's in this file).

include/aikido/planner/parabolic/ParabolicSmoother.hpp
```diff
@@ -129,6 +130,60 @@ std::unique_ptr<trajectory::Spline> doShortcutAndBlend(
     double _checkResolution = DEFAULT_CHECK_RESOLUTION,
     double _tolerance = DEFAULT_TOLERANCE);
 
+/// Class for performing parabolic smoothing on trajectories
+class SmoothTrajectoryPostProcessor
```
brianhou: I'd still like to rename this `ParabolicSmoother` (and I think it makes more sense now that it's in this file).

include/aikido/planner/parabolic/ParabolicTimer.hpp
```diff
@@ -43,6 +44,28 @@ std::unique_ptr<aikido::trajectory::Spline> computeParabolicTiming(
...
+{
+public:
+  /// \param _velocityLimits maximum velocity for each dimension.
+  /// \param _accelerationLimits maximum acceleration for each dimension.
+  RetimeTrajectoryPostProcessor(
+      const Eigen::VectorXd& _velocityLimits,
+      const Eigen::VectorXd& _accelerationLimits);
+
+  /// Performs parabolic retiming on an input trajectory.
+  /// Documentation inherited.
```
brianhou: I think since you're adding something here, this might need to look like

```
/// Performs parabolic retiming on an input trajectory.
/// \copydoc TrajectoryPostProcessor::postprocess
```

include/aikido/planner/parabolic/ParabolicTimer.hpp
```diff
@@ -43,6 +44,28 @@ std::unique_ptr<aikido::trajectory::Spline> computeParabolicTiming(
...
+  std::unique_ptr<aikido::trajectory::Spline> postprocess(
+      const aikido::trajectory::InterpolatedPtr& _inputTraj,
+      const aikido::common::RNG* _rng) override;
+
+private:
```
brianhou: Please document these members.

include/aikido/planner/parabolic/ParabolicSmoother.hpp
```diff
@@ -129,6 +130,60 @@ std::unique_ptr<trajectory::Spline> doShortcutAndBlend(
...
+  SmoothTrajectoryPostProcessor(
+      const Eigen::VectorXd& _velocityLimits,
+      const Eigen::VectorXd& _accelerationLimits,
+      const aikido::constraint::TestablePtr& _collisionTestable,
+      bool _enableShortcut = true,
+      bool _enableBlend = true,
+      double _shortcutTimelimit = DEFAULT_TIMELIMT,
+      double _blendRadius = DEFAULT_BLEND_RADIUS,
+      int _blendIterations = DEFAULT_BLEND_ITERATIONS,
+      double _smootherFeasibilityCheckResolution = DEFAULT_CHECK_RESOLUTION,
```
brianhou: I don't think "smoother" is necessary in this variable name (and the one below).

sniyaz:
@brianhou Final nits responded to!

jslee02:
:+1: Thanks for all the updates!

brianhou:
Looks good to me! I'm happy to merge this once Travis is satisfied.

PR #302, opened 2018-01-18
Add ParabolicTiming for linear spline

dqyi11:
This pull request adds `computeParabolicTime` for linear spline trajectory.


CHANGELOG.md
```diff
@@ -30,6 +30,7 @@
 
   * Added World class: [#243](https://github.com/personalrobotics/aikido/pull/243), [#252](https://github.com/personalrobotics/aikido/pull/252), [#265](https://github.com/personalrobotics/aikido/pull/265)
   * Added vector field planner [#246](https://github.com/personalrobotics/aikido/pull/246), [#262](https://github.com/personalrobotics/aikido/pull/262), [#268](https://github.com/personalrobotics/aikido/pull/268)
+  * Added parabolic timing for linear spline [#302](https://github.com/personalrobotics/aikido/pull/302)
```
jslee02: Please move this to 0.3.0.

src/planner/parabolic/ParabolicTimer.cpp
```diff
@@ -113,6 +115,43 @@ std::unique_ptr<aikido::trajectory::Spline> computeParabolicTiming(
...
+    if (!std::isfinite(_maxAcceleration[i]))
+      throw std::invalid_argument("Acceleration limits must be finite.");
+  }
+
+  double startTime = _inputTrajectory.getStartTime();
+  auto dynamicPath = detail::convertToDynamicPath(
+      _inputTrajectory, _maxVelocity, _maxAcceleration, true);
+
+  auto outputTrajectory
+      = detail::convertToSpline(*dynamicPath, startTime, stateSpace);
```
jslee02: Is `dynamicPath` always not `nullptr`? If so, please throw an exception if `dynamicPath` is `nullptr`. 

Otherwise, return `nullptr` as output trajectory if `dynamicPath` is `nullptr`.

src/planner/parabolic/ParabolicTimer.cpp
```diff
@@ -113,6 +115,43 @@ std::unique_ptr<aikido::trajectory::Spline> computeParabolicTiming(
...
+
+  for (std::size_t i = 0; i < dimension; ++i)
+  {
+    if (_maxVelocity[i] <= 0.)
+      throw std::invalid_argument("Velocity limits must be positive.");
+    if (!std::isfinite(_maxVelocity[i]))
+      throw std::invalid_argument("Velocity limits must be finite.");
+
+    if (_maxAcceleration[i] <= 0.)
+      throw std::invalid_argument("Acceleration limits must be positive.");
```
jslee02: Shouldn't be fine to allow zero velocity/acceleration limits? Could you explain for me why nonnegative limits aren't enough? :sweat_smile:

dqyi11: This is following the code of `computeParabolicTiming` for an `Interpolated`.
I think the reason is that if the velocity/acceleration limit of one dimension is zero, it should not be considered in timing a trajectory.

src/planner/parabolic/ParabolicTimer.cpp
dqyi11: The current implementation of `convertToDynamicPath` guarantees that `dynamicPath` will never be `nullptr`.
Do you think we shall add the check in case any potential change will happen in the future?

src/planner/parabolic/ParabolicUtil.hpp
```diff
@@ -60,7 +60,8 @@ std::unique_ptr<aikido::trajectory::Spline> convertToSpline(
 std::unique_ptr<ParabolicRamp::DynamicPath>
     convertToDynamicPath(const aikido::trajectory::Spline& _inputTrajectory,
                          const Eigen::VectorXd& _maxVelocity,
-                         const Eigen::VectorXd& _maxAcceleration);
+                         const Eigen::VectorXd& _maxAcceleration,
+                         bool untimed = false);
```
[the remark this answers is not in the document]
psigen: In general, I am a fan of boolean flags to indicate input properties including an `is` or `has` prefix to help indicate their intent e.g. `isTimed`.  But that's neither here nor there.

dqyi11:
@brianhou  I agree with your suggestion on introducing a `WaypointTrajectory` class.
Shall we create an issue for discussion before making the changes?
My minor concern is that a `Spline` is a set of segments. A segment includes a start state, duration and curve coefficients.
https://github.com/personalrobotics/aikido/blob/master/include/aikido/trajectory/Spline.hpp#L132
The "waypoints" of a `Spline` is "indirectly" obtained from segments.
https://github.com/personalrobotics/aikido/blob/master/src/trajectory/Spline.cpp#L206
https://github.com/personalrobotics/aikido/blob/master/src/trajectory/Spline.cpp#L226

There is an alternative approach in conversion.
We could generate a time sequence from start time to end time, and call `evaluate` method of a `Trajectory` class to generate a sequence of "waypoints" for conversion.
But it will be very hard to determine a "good" time step.

brianhou:
I see. Let's just merge this PR as is (to unblock @sniyaz and MAGI cleanup) and then create an issue to track this cleanup. @dqyi11, can you create the issue? I'll merge once Travis is satisfied.

dqyi11:
@brianhou Thanks!
I have created an issue #312

PR #305, opened 2018-01-19
fix step sequence iterator

dqyi11:
In current `aikido`, unexpected exception is thrown in some cases.
`StepSequence seq(0.2, false); for (double v : seq) { DART_UNUSED(v); count++; }`
An exception will be throw
>unknown file: Failure
>C++ exception with description "Indexed maximum intenger." thrown in the test body.
Expected sequence shall be `(0.0, 0.2, 0.4, 0.6, 0.8)`; but `getMaxSteps()` returns 6.
Thus an exception is triggered in `operator []`.

In this PR,
 `StepSequence seq(0.2, true)` returns `(0.0, 0.2, 0.4, 0.6, 0.8, 1.0)` and `getMaxSteps()` returns 6.
 `StepSequence seq(0.2, false)` returns `(0.0, 0.2, 0.4, 0.6, 0.8)` and `getMaxSteps()` returns 5.
Unit tests are updated accordingly.

### DIFF
--- a/src/common/StepSequence.cpp
+++ b/src/common/StepSequence.cpp
@@ -59,27 +59,12 @@ StepSequence::const_iterator StepSequence::end()
 int StepSequence::getMaxSteps() const
 {
   int numSteps = (mEndPoint - mStartPoint) / mStepSize;
-  if (!mIncludeEndpoints)
+
+  if (std::abs(mStartPoint + mStepSize * numSteps - mEndPoint) > 1e-7)
   {
-    if (std::abs(mStartPoint + mStepSize * numSteps - mEndPoint) < 1e-7)
-      return numSteps;
-    else
-      // Return numSteps + 1 for the start
-      return numSteps + 1;
+    numSteps++;
   }
-  else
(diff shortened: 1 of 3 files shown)


src/common/StepSequence.cpp
```diff
@@ -61,8 +61,11 @@ int StepSequence::getMaxSteps() const
   int numSteps = (mEndPoint - mStartPoint) / mStepSize;
   if (!mIncludeEndpoints)
   {
-    // Return numSteps + 1 for the start
-    return numSteps + 1;
+    if (fabs(mStartPoint + mStepSize * numSteps - mEndPoint) < 1e-7)
```
brianhou: I found the current logic somewhat difficult to follow. Can we do something like:

```
int numSteps = (mEndPoint - mStartPoint) / mStepSize;
if (fabs(mStartPoint + mStepSize * numSteps - mEndPoint) > 1e-7)
  numSteps++; // include start
return numSteps + mIncludeEndpoints;
```

PR #308, opened 2018-01-20
Use forward declaration of smart pointers

jslee02:
This PR moves the declaration of smart pointers to separate headers to avoid the circular dependencies.

***

### DIFF
--- a/include/aikido/common/smart_pointer.hpp
+++ b/include/aikido/common/smart_pointer.hpp
@@ -1,0 +1,21 @@
+#ifndef AIKIDO_COMMON_SMARTPOINTER_HPP_
+#define AIKIDO_COMMON_SMARTPOINTER_HPP_
+
+#include <dart/common/SmartPointer.hpp>
+
+// clang-format off
+#define AIKIDO_COMMON_DECLARE_SMART_POINTERS(X)                                \
+  DART_COMMON_MAKE_SHARED_WEAK(X)                                              \
+  using Unique ## X ## Ptr      = std::unique_ptr< X >;                        \
+  using UniqueConst ## X ## Ptr = std::unique_ptr< const X >;
+// clang-format on
+
+namespace aikido {
+namespace common {
+
+// Declare here the smart pointers of aikido::common objects
+
+} // namespace common
+} // namespace aikido
+
+#endif // AIKIDO_COMMON_SMARTPOINTER_HPP_
--- a/include/aikido/constraint/CartesianProductProjectable.hpp
+++ b/include/aikido/constraint/CartesianProductProjectable.hpp
@@ -1,6 +1,7 @@
 #ifndef AIKIDO_CONSTRAINT_CARTESIANPRODUCTPROJECTABLE_HPP_
 #define AIKIDO_CONSTRAINT_CARTESIANPRODUCTPROJECTABLE_HPP_
 
+#include "aikido/constraint/smart_pointer.hpp"
 #include "../statespace/CartesianProduct.hpp"
 #include "Projectable.hpp"
 
--- a/include/aikido/constraint/CartesianProductTestable.hpp
+++ b/include/aikido/constraint/CartesianProductTestable.hpp
@@ -2,6 +2,7 @@
 #define AIKIDO_CONSTRAINT_TESTABLESUBSPACE_HPP_
 
 #include <vector>
+#include "aikido/constraint/smart_pointer.hpp"
 #include "../statespace/CartesianProduct.hpp"
 #include "Testable.hpp"
 
--- a/include/aikido/constraint/CollisionFree.hpp
+++ b/include/aikido/constraint/CollisionFree.hpp
@@ -90,8 +90,6 @@ private:
   std::vector<std::shared_ptr<CollisionGroup>> mGroupsToSelfCheck;
 };
 
-using CollisionFreePtr = std::shared_ptr<CollisionFree>;
-
 } // namespace constraint
 } // namespace aikido
 
--- a/include/aikido/constraint/CyclicSampleable.hpp
+++ b/include/aikido/constraint/CyclicSampleable.hpp
@@ -1,6 +1,7 @@
 #ifndef AIKIDO_CONSTRAINT_CYCLICSAMPLEABLE_HPP_
 #define AIKIDO_CONSTRAINT_CYCLICSAMPLEABLE_HPP_
 
+#include "aikido/constraint/smart_pointer.hpp"
 #include "Sampleable.hpp"
 
 namespace aikido {
--- a/include/aikido/constraint/Differentiable.hpp
+++ b/include/aikido/constraint/Differentiable.hpp
@@ -3,6 +3,7 @@
 
 #include <memory>
 #include <Eigen/Dense>
+#include "aikido/statespace/smart_pointer.hpp"
 #include "../statespace/StateSpace.hpp"
 
 namespace aikido {
@@ -66,8 +67,6 @@ public:
       Eigen::VectorXd& _val,
       Eigen::MatrixXd& _jac) const;
 };
-
-using DifferentiablePtr = std::shared_ptr<Differentiable>;
 
 } // namespace constraint
 } // namespace aikido
--- a/include/aikido/constraint/DifferentiableIntersection.hpp
+++ b/include/aikido/constraint/DifferentiableIntersection.hpp
@@ -3,6 +3,7 @@
 
 #include <memory>
 #include <Eigen/Dense>
+#include "aikido/constraint/smart_pointer.hpp"
 #include "../statespace/StateSpace.hpp"
 #include "Differentiable.hpp"
 
--- a/include/aikido/constraint/DifferentiableSubspace.hpp
+++ b/include/aikido/constraint/DifferentiableSubspace.hpp
@@ -1,6 +1,7 @@
 #ifndef AIKIDO_CONSTRAINT_DIFFERENTIABLESUBSPACE_HPP_
 #define AIKIDO_CONSTRAINT_DIFFERENTIABLESUBSPACE_HPP_
 
+#include "aikido/constraint/smart_pointer.hpp"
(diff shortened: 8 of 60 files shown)


psigen:
I'm not too familiar with the mechanics or reasoning behind this change.  @jslee02 could you help me understand:
1) Why do we need this declaration at the top of every file, including many that do not declare any smart pointers?
2) What is the purpose of the empty namespace declaration in the `aikido/common/smart_pointer.hpp` (https://github.com/personalrobotics/aikido/pull/308/files#diff-aed776db41926218904fdef145673436R16)

jslee02:
1. Unnecessary header inclusion (doesn't use any [smart] pointers) should be removed. Let me double check if there are.
2. That is a placeholder for possible smart pointer declaration for objects in `common` namespace.

psigen:
Thanks for the clarifications!

Here is one example where I couldn't figure out why the include needed to be added.  Perhaps there is some usage that I'm not immediately seeing:

https://github.com/personalrobotics/aikido/blob/30aca0849f305fbf00ac60a8de94a5b2b4124aae/include/aikido/constraint/CyclicSampleable.hpp

jslee02:
That's because `SampleablePtr` ([here](https://github.com/personalrobotics/aikido/blob/30aca0849f305fbf00ac60a8de94a5b2b4124aae/include/aikido/constraint/CyclicSampleable.hpp#L21) and [here](https://github.com/personalrobotics/aikido/blob/30aca0849f305fbf00ac60a8de94a5b2b4124aae/include/aikido/constraint/CyclicSampleable.hpp#L30)) is defined in `aikido/constraint/smart_pointer.hpp`.

psigen:
Ah, I think I understand, the `SampleablePtr` here is coming through this header instead of `Sampleable.hpp`, is that right?

This is going to be a bit confusing to track over time, since it's going to be difficult to tell which declarations are coming from which headers.

Do you have a link to the issue explaining where we were encountering circular dependencies with the smart pointer declaration?  I'd like to learn more about this issue.

jslee02:
> This is going to be a bit confusing to track over time, since it's going to be difficult to tell which declarations are coming from which headers.

The pattern would be like this. All the smart pointers of the objects in <component> will be defined in `aikido/<component>/smart_pointer.hpp`. In `Sampleable` case, since it's in constraint component (or namespace), its smart pointers are defined in `aikido/constraint/smart_pointer.hpp`.

> Do you have a link to the issue explaining where we were encountering circular dependencies with the smart pointer declaration? I'd like to learn more about this issue.

Here is a simple example we might encounter circular dependencies:
```cpp
// A.hpp
#include "B.hpp" // for BPtr
class A {
public:
  void foo();
  BPtr mB; 
};
using APtr = std::shared_ptr<A>;

// B.hpp
class B {
public:
  void bar();
  APtr mA;
};
using BPtr = std::shared_ptr<B>;
```

psigen:
Right, that's true, but we don't actually require a complete or unique declaration for the `shared_ptr`, so would it not be possible to do:

```
// A.hpp

class B;
using BPtr = std::shared_ptr<B>;

class A {
public:
  void foo();
  BPtr mB; 
};
using APtr = std::shared_ptr<A>;

// B.hpp
#include "A.hpp" // for APtr
class B {
public:
  void bar();
  APtr mA;
};
using BPtr = std::shared_ptr<B>;
```

Since we already have macros defined for this, it seems like we would actually only really one macro line:

```
// A.hpp

AIKIDO_COMMON_DECLARE_SMART_POINTERS(B)

class A {
public:
  void foo();
  BPtr mB; 
};
AIKIDO_COMMON_DECLARE_SMART_POINTERS(A)

// B.hpp
#include "A.hpp" // for APtr
class B {
public:
  void bar();
  APtr mA;
};
AIKIDO_COMMON_DECLARE_SMART_POINTERS(B)
```

This means that if there were an inconsistent declaration, we would get a compile error, but it seems like the automated build tests would catch this, and could only occur within files that had close enough coupling to introduce a circular dependency.

jslee02:
Your second code block would work, but I would prefer not including `A.hpp` only to get the forward declaration in the header because it would introduce unnecessary additional header inclusion (that are included in `A.hpp`. That would increase the compile time. The overhead wouldn't be too much for our project size, but why not avoid when we can.

One another potential issue of the second code block is that we might end up many lines of `AIKIDO_COMMON_DECLARE_SMART_POINTERS(...)` as we use more than one smart pointers. Also, if we use the macro anyway then why we need the smart pointer declaration at the end of header files.

psigen:
Those seems like reasonable points.

It definitely make it hard to tell which files contain the declaration of which pointers, and for instances like this:
https://github.com/personalrobotics/aikido/blob/30aca0849f305fbf00ac60a8de94a5b2b4124aae/include/aikido/constraint/CyclicSampleable.hpp#L24

it takes a bit of thinking to figure out that the `StateSpacePtr` is actually coming in through `Sampleable.hpp`, while the `SampleablePtr` is coming in through `smart_pointer.hpp`.

Also, if the primary function of smart_pointer.hpp is to _contain_ pointers, rather than to declare the structure of the pointers, perhaps it should be `smart_pointers.hpp`?  Though interestingly, it will also contain _every_ related class forward declaration as well, which might be useful when declaring occasional other pointer types like: https://github.com/personalrobotics/aikido/blob/30aca0849f305fbf00ac60a8de94a5b2b4124aae/include/aikido/constraint/CyclicSampleable.hpp#L27
so should it just be `declarations.hpp`?

psigen:
Doing some reading, the only examples I could find of this pattern was the C header `iosfwd`, and an example in a C textbook called `cfwd.h`, so I'm not sure if there's any other naming convention that would make this clearer.

jslee02:
> it takes a bit of thinking to figure out that the StateSpacePtr is actually coming in through Sampleable.hpp, while the SampleablePtr in the same file is coming in through smart_pointer.hpp.

That file is just an oversight. It should include `aikido/statespace/smart_pointer.hpp` because `StateSpace` is in `statespace` component.

Here is a summary of my thought on this issue:

#### Baseline

* I assume we would prefer using type aliasing for smart pointers for less typing. Technically, we don't need any type aliasings. (`std::shared_ptr<X>` --> `XPtr`)
* There are many classes that don't have their smart pointer aliasings.
* I assume we wouldn't prefer declaring the same smart pointers here and there.

#### Solution 1

* Declare smart pointers at the bottom of the corresponding header. For example,
```cpp
class A {};
AIKIDO_COMMON_DECLARE_SMART_POINTERS(A) 
```
* Downside 1: Circular dependency can happen (https://github.com/personalrobotics/aikido/pull/308#issuecomment-359212765).

* Downside 2: `A` cannot use the smart pointer alias. For example,
```cpp
class A {
  static APtr create(); // error
};
AIKIDO_COMMON_DECLARE_SMART_POINTERS(A) 
```

#### Solution 2

* Declare smart pointers at the top of the corresponding header. This can resolve the above Downside 2, but circular dependency issue still exists.

#### Solution 3

* Define all the smart pointers of `aikido::<namespace>::X`s in `aikido/<component>/smart_pointer.hpp` (or `smart_pointers.hpp`). 
* Include `aikido::<namespace>::smart_pointer.hpp` to every classe in `include/aikido/<namespace>/ClassName.hpp`.

I would like to propose "Solution 3":

If you want just pointer manipulation (don't need the definition) of a class in `<namespace>`, then include `aikido/<namespace>/smart_pointer.hpp`. This should resolve circular dependencies.

Otherwise, just include `aikido/<namespace>/Class.hpp`. You could use `Class` and its smart pointers as well.

For the classes in the subdirectories of `include/aikido/<namespace>/...`, I'm open to declare their smart pointers in `include/aikido/<namespace>/smart_pointer.hpp` or `include/aikido/<namespace>/sub_directory/smart_pointer.hpp`.

@psigen @mkoval (and others): If you find any confusion in "Solution 3" or better idea, let me know.

psigen:
Thanks for taking the time to think through this with me @jslee02 👍 

I think my preference would still be *Solution 2*, since:
(1) I honestly don't think we have _that_ many circular dependencies and any one can be manually broken with a single extra `AIKIDO_COMMON_DECLARE_SMART_POINTERS()` in one of the headers.
(2) I think the majority of users will _not_ be expecting to do only pointer manipulation of a class, and will generally want both the class and its pointers together.

But I understand your reasoning to do *Solution 3* since it is the most general model and supports the use case of a client that wants to do only pointer-manipulation without introducing dependencies and also not re-calling the `AIKIDO_COMMON_DECLARE_SMART_POINTERS()` macro.

Within *Solution 3*, the naming convention of `include/aikido/<namespace>/smart_pointer.hpp` seems generally fine.  I might suggest we simply name it `.../pointers.hpp` since the fact that they are smart is really more of an implementation detail, and we might include non-smart-pointer aliases in this file in the future.

I would caution that *Solution 3* adds what seems to me like a _lot_ of code bloat.  We are adding many header files, and many places that those headers need to be included accurately, all enforced by convention, not compilation.  I would point to the oversight with `StateSpacePtr` as an example of how it could be difficult to track whether this is being done correctly.

However, I would really welcome some external input from more experienced folks.  Maybe @mkoval or even @mxgrey would be able to provide some better architectural guidance here.

jslee02:
> Within Solution 3, the naming convention of include/aikido/<namespace>/smart_pointer.hpp seems generally fine. I might suggest we simply name it .../pointers.hpp since the fact that they are smart is really more of an implementation detail, and we might include non-smart-pointer aliases in this file in the future.

:+1:

jslee02:
I think we could consider another solution:

#### Solution 4

* Solution 2 + having `pointers.hpp`

We generally include `aikido/component/ClassA.hpp` to use `ClassA` and its pointers, and rarely include `aikido/component/pointers.hpp` when encounting circular dependencies.

jslee02:
@psigen I think Solution 4 can be a compromise here. If it sounds fine, then I will change this PR with it.

jslee02:
Closing in favor of #317

PR #314, opened 2018-01-25
[WIP] Refactor planner API

jslee02:
This is a WIP PR to initiate discussion on introducing planner API.

**Summary**
* Add planner classes
  * `Planner`: Base class
  * `ConcretePlanner : Planner`: Base class for non-meta planners (e.g., `SnapPlanner`). Open to better name.
  * `MetaPlanner`
* Add problem classes (_method_ in `prpy`): This is to encapsulate different sets of parameters per planning problems.

***

### DIFF
--- a/include/aikido/planner/Planner.hpp
+++ b/include/aikido/planner/Planner.hpp
@@ -1,0 +1,33 @@
+#ifndef AIKIDO_PLANNER_PLANNER_HPP_
+#define AIKIDO_PLANNER_PLANNER_HPP_
+
+#include <functional>
+#include <unordered_set>
+#include "aikido/planner/Problem.hpp"
+#include "aikido/trajectory/Trajectory.hpp"
+
+namespace aikido {
+namespace planner {
+
+/// Base class for a meta-planner
+class Planner
+{
+public:
+  /// Returns true if this planner can solve \c problem.
+  virtual bool canSolve(const Problem* problem) const = 0;
+
+  /// Returns all the problems that this planner can solve.
+  virtual std::unordered_set<std::string> getSolvableProblems() const = 0;
+
+  /// Solves \c problem returning the result to \c result.
+  virtual trajectory::TrajectoryPtr solve(
+      const Problem* problem, Problem::Result* result = nullptr)
+      = 0;
+};
+
+using PlannerPtr = std::shared_ptr<Planner>;
+
+} // namespace planner
+} // namespace aikido
+
+#endif // AIKIDO_PLANNER_PLANNER_HPP_
--- a/include/aikido/planner/Problem.hpp
+++ b/include/aikido/planner/Problem.hpp
@@ -1,0 +1,54 @@
+#ifndef AIKIDO_PLANNER_PROBLEM_HPP_
+#define AIKIDO_PLANNER_PROBLEM_HPP_
+
+#include <string>
+#include "aikido/statespace/StateSpace.hpp"
+
+namespace aikido {
+namespace planner {
+
+/// Base class for various planning problems.
+class Problem
+{
+public:
+  class Result;
+
+  /// Constructor.
+  explicit Problem(statespace::StateSpacePtr stateSpace);
+
+  /// Destructor.
+  virtual ~Problem() = default;
+
+  /// Returns name.
+  virtual const std::string& getName() const = 0;
+
+  /// Returns const state space.
+  statespace::StateSpacePtr getStateSpace() const;
+
+protected:
+  /// State space associated with this problem.
+  statespace::StateSpacePtr mStateSpace;
+};
+
+/// Base class for planning result of various planning problems.
+class Problem::Result
+{
+public:
+  /// Destructor.
+  virtual ~Result() = default;
+
+  /// Sets message.
+  void setMessage(const std::string& message);
+
+  /// Returns message.
+  const std::string& getMessage() const;
+
+protected:
+  /// Message.
+  std::string mMessage;
+};
+
+} // namespace planner
+} // namespace aikido
+
+#endif
--- a/include/aikido/planner/RankedMetaPlanner.hpp
+++ b/include/aikido/planner/RankedMetaPlanner.hpp
@@ -1,0 +1,22 @@
+#ifndef AIKIDO_PLANNER_RANKEDMETAPLANNER_HPP_
+#define AIKIDO_PLANNER_RANKEDMETAPLANNER_HPP_
+
+#include "aikido/planner/MetaPlanner.hpp"
+
+namespace aikido {
+namespace planner {
+
+class RankedMetaPlanner : public MetaPlanner
+{
+public:
+  // Documentation inherited.
+  trajectory::TrajectoryPtr solve(
+      const Problem* problem, Problem::Result* result = nullptr) override;
+
+protected:
+};
+
+} // namespace planner
+} // namespace aikido
+
+#endif // AIKIDO_PLANNER_RANKEDMETAPLANNER_HPP_
--- a/include/aikido/planner/SequenceMetaPlanner.hpp
+++ b/include/aikido/planner/SequenceMetaPlanner.hpp
@@ -1,0 +1,20 @@
+#ifndef AIKIDO_PLANNER_SEQUENCEMETAPLANNER_HPP_
+#define AIKIDO_PLANNER_SEQUENCEMETAPLANNER_HPP_
+
+#include "aikido/planner/MetaPlanner.hpp"
+
+namespace aikido {
+namespace planner {
+
+class SequenceMetaPlanner : public MetaPlanner
+{
+public:
+  // Documentation inherited.
+  trajectory::TrajectoryPtr solve(
+      const Problem* problem, Problem::Result* result = nullptr) override;
+};
+
+} // namespace planner
+} // namespace aikido
+
+#endif // AIKIDO_PLANNER_SEQUENCEMETAPLANNER_HPP_
--- a/src/planner/FirstSupportedMetaPlanner.cpp
+++ b/src/planner/FirstSupportedMetaPlanner.cpp
@@ -1,0 +1,24 @@
+#include "aikido/planner/FirstSupportedMetaPlanner.hpp"
+
+namespace aikido {
+namespace planner {
+
+//==============================================================================
+trajectory::TrajectoryPtr FirstSupportedMetaPlanner::solve(
+    const Problem* problem, Problem::Result* result)
+{
+  for (const auto& planner : mPlanners)
+  {
+    if (!planner->canSolve(problem))
+      continue;
+
+    return planner->solve(problem, result);
+  }
+
+  // TODO: Set result "no supported planner".
+
+  return nullptr;
+}
+
+} // namespace planner
+} // namespace aikido
--- a/src/planner/Problem.cpp
+++ b/src/planner/Problem.cpp
@@ -1,0 +1,32 @@
+#include "aikido/planner/Problem.hpp"
+
+namespace aikido {
+namespace planner {
+
+//==============================================================================
+Problem::Problem(statespace::StateSpacePtr stateSpace)
+  : mStateSpace(std::move(stateSpace))
+{
+  // Do nothing
+}
+
+//==============================================================================
+statespace::StateSpacePtr Problem::getStateSpace() const
+{
+  return mStateSpace;
+}
+
+//==============================================================================
+void Problem::Result::setMessage(const std::string& message)
+{
+  mMessage = message;
+}
+
+//==============================================================================
+const std::string& Problem::Result::getMessage() const
+{
+  return mMessage;
+}
+
+} // namespace planner
+} // namespace aikido
--- a/tests/planner/test_SnapPlanner.cpp
+++ b/tests/planner/test_SnapPlanner.cpp
@@ -4,6 +4,7 @@
 #include <aikido/constraint/CollisionFree.hpp>
 #include <aikido/constraint/Testable.hpp>
 #include <aikido/distance/defaults.hpp>
+#include <aikido/planner/PlanToConfiguration.hpp>
 #include <aikido/planner/PlanningResult.hpp>
 #include <aikido/planner/SnapPlanner.hpp>
 #include <aikido/statespace/GeodesicInterpolator.hpp>
@@ -14,6 +15,7 @@
 using std::shared_ptr;
 using std::make_shared;
 
+//==============================================================================
 class SnapPlannerTest : public ::testing::Test
 {
 public:
@@ -53,42 +55,76 @@ public:
   shared_ptr<PassingConstraint> passingConstraint;
   shared_ptr<FailingConstraint> failingConstraint;
   shared_ptr<GeodesicInterpolator> interpolator;
-  aikido::planner::PlanningResult planningResult;
+  aikido::planner::PlanToConfiguration::Result planningResult;
 };
 
+//==============================================================================
+class UnknownProblem : public aikido::planner::Problem
+{
+public:
+  UnknownProblem() : aikido::planner::Problem(nullptr)
+  {
+    // Do nothing
+  }
+
+  const std::string& getName() const override
+  {
+    return getStaticName();
+  }
+
+  static const std::string& getStaticName()
+  {
+    static std::string name("UnknownProblem");
+    return name;
+  }
+};
+
+//==============================================================================
+TEST_F(SnapPlannerTest, CanSolveProblems)
+{
+  auto planner = std::make_shared<aikido::planner::SnapPlanner>();
+
+  auto problem = aikido::planner::PlanToConfiguration(
+      stateSpace, *startState, *goalState, interpolator, failingConstraint);
+  auto unknownProblem = UnknownProblem();
+
+  EXPECT_TRUE(planner->canSolve(&problem));
+  EXPECT_FALSE(planner->canSolve(&unknownProblem));
+}
+
+//==============================================================================
 TEST_F(SnapPlannerTest, ThrowsOnStateSpaceMismatch)
 {
-  SkeletonPtr empty_skel = dart::dynamics::Skeleton::create("skel");
+  SkeletonPtr emptySkel = dart::dynamics::Skeleton::create("skel");
   auto differentStateSpace
-      = make_shared<MetaSkeletonStateSpace>(empty_skel.get());
+      = make_shared<MetaSkeletonStateSpace>(emptySkel.get());
   EXPECT_THROW(
       {
-        planSnap(
+        auto problem = aikido::planner::PlanToConfiguration(
             differentStateSpace,
             *startState,
             *goalState,
             interpolator,
-            passingConstraint,
-            planningResult);
+            passingConstraint);
+        DART_UNUSED(problem);
       },
       std::invalid_argument);
 }
 
+//==============================================================================
 TEST_F(SnapPlannerTest, ReturnsStartToGoalTrajOnSuccess)
 {
   stateSpace->getState(skel.get(), *startState);
   skel->setPosition(0, 2.0);
   stateSpace->setState(skel.get(), *goalState);
 
-  auto traj = planSnap(
-      stateSpace,
-      *startState,
-      *goalState,
-      interpolator,
-      passingConstraint,
-      planningResult);
+  auto problem = aikido::planner::PlanToConfiguration(
+      stateSpace, *startState, *goalState, interpolator, passingConstraint);
+  auto planner = std::make_shared<aikido::planner::SnapPlanner>();
+  auto traj = planner->planToConfiguration(&problem, &planningResult);
 
   auto subSpace = stateSpace->getSubspace<SO2>(0);
+  DART_UNUSED(subSpace);
   EXPECT_EQ(2, traj->getNumWaypoints());
 
   auto startValue = startState->getSubStateHandle<SO2>(0).getRotation();
@@ -108,14 +144,12 @@ TEST_F(SnapPlannerTest, ReturnsStartToGoalTrajOnSuccess)
       << "on success final element of trajectory should be goal state.";
 }
 
+//==============================================================================
 TEST_F(SnapPlannerTest, FailIfConstraintNotSatisfied)
 {
-  auto traj = planSnap(
-      stateSpace,
-      *startState,
-      *goalState,
-      interpolator,
-      failingConstraint,
-      planningResult);
+  auto problem = aikido::planner::PlanToConfiguration(
+      stateSpace, *startState, *goalState, interpolator, failingConstraint);
+  auto planner = std::make_shared<aikido::planner::SnapPlanner>();
+  auto traj = planner->planToConfiguration(&problem, &planningResult);
   EXPECT_EQ(nullptr, traj);
 }
--- a/include/aikido/constraint/Testable.hpp
+++ b/include/aikido/constraint/Testable.hpp
@@ -37,6 +37,7 @@ public:
 };
 
 using TestablePtr = std::shared_ptr<Testable>;
+using ConstTestablePtr = std::shared_ptr<const Testable>;
 
 } // namespace constraint
 } // namespace aikido
--- a/include/aikido/planner/ConcretePlanner.hpp
+++ b/include/aikido/planner/ConcretePlanner.hpp
@@ -1,0 +1,56 @@
+#ifndef AIKIDO_PLANNER_CONCRETEPLANNER_HPP_
+#define AIKIDO_PLANNER_CONCRETEPLANNER_HPP_
+
+#include <functional>
+#include <unordered_map>
+#include "aikido/planner/Planner.hpp"
+#include "aikido/planner/Problem.hpp"
+#include "aikido/trajectory/Trajectory.hpp"
+
+namespace aikido {
+namespace planner {
+
+class ConcretePlanner : public Planner
+{
+public:
+  /// Constructor
+  ConcretePlanner() = default;
+
+  /// Destructor
+  virtual ~ConcretePlanner() = default;
+
+  // Documentation inherited.
+  bool canSolve(const Problem* problem) const override;
+
+  // Documentation inherited.
+  std::unordered_set<std::string> getSolvableProblems() const override;
+
+  // Documentation inherited.
+  trajectory::TrajectoryPtr solve(
+      const Problem* problem, Problem::Result* result = nullptr) override;
+
+protected:
+  using PlanningFunction = std::function<trajectory::TrajectoryPtr(
+      const Problem*, Problem::Result*)>;
+  using PlanningFunctionMap = std::unordered_map<std::string, PlanningFunction>;
+
+  /// Returns planning function map.
+  ///
+  /// The concrete planner classes should contain a static map and return it by
+  /// overriding this function.
+  virtual PlanningFunctionMap& getPlanningFunctionMap() = 0;
+
+  /// Returns planning function map.
+  const PlanningFunctionMap& getPlanningFunctionMap() const;
+
+  /// Registers planning function to this planner.
+  template <class ProblemT, typename T, typename R, typename... Args>
+  void registerPlanningFunction(R (T::*func)(Args...));
+};
+
+} // namespace planner
+} // namespace aikido
+
+#include "aikido/planner/detail/ConcretePlanner-impl.hpp"
+
+#endif // AIKIDO_PLANNER_CONCRETEPLANNER_HPP_
--- a/include/aikido/planner/FirstSupportedMetaPlanner.hpp
+++ b/include/aikido/planner/FirstSupportedMetaPlanner.hpp
@@ -1,0 +1,20 @@
+#ifndef AIKIDO_PLANNER_FIRSTSUPPORTEDMETAPLANNER_HPP_
+#define AIKIDO_PLANNER_FIRSTSUPPORTEDMETAPLANNER_HPP_
+
+#include "aikido/planner/MetaPlanner.hpp"
+
+namespace aikido {
+namespace planner {
+
+class FirstSupportedMetaPlanner : public MetaPlanner
+{
+public:
+  // Documentation inherited.
+  trajectory::TrajectoryPtr solve(
+      const Problem* problem, Problem::Result* result = nullptr) override;
+};
+
+} // namespace planner
+} // namespace aikido
+
+#endif // AIKIDO_PLANNER_FIRSTSUPPORTEDMETAPLANNER_HPP_
--- a/include/aikido/planner/MetaPlanner.hpp
+++ b/include/aikido/planner/MetaPlanner.hpp
@@ -1,0 +1,50 @@
+#ifndef AIKIDO_PLANNER_METAPLANNER_HPP_
+#define AIKIDO_PLANNER_METAPLANNER_HPP_
+
+#include <unordered_set>
+#include <vector>
+#include "aikido/planner/Planner.hpp"
+
+namespace aikido {
+namespace planner {
+
+class MetaPlanner : public Planner
+{
+public:
+  /// Constructor.
+  explicit MetaPlanner(const std::vector<PlannerPtr>& planners);
+
+  /// Adds planner.
+  void addPlanner(PlannerPtr planner);
+
+  /// Returns true if this MetaPlanner contains \c planner.
+  bool hasPlanner(const Planner* planner);
+
+  /// Returns all the planners in this MetaPlanner.
+  const std::vector<PlannerPtr>& getPlanners() const;
+
+  /// Returns planner given \c index.
+  ///
+  /// \throw If index is equal to or greater than the number of planners in this
+  /// MetaPlanner.
+  PlannerPtr getPlanner(std::size_t index);
+
+  // Documentation inherited.
+  bool canSolve(const Problem* problem) const override;
+
+  // Documentation inherited.
+  std::unordered_set<std::string> getSolvableProblems() const override;
+
+  /// Returns set of planners that can solve \c problem.
+  std::unordered_set<PlannerPtr> getPlannersCanSolve(
+      const Problem* problem) const;
+
+protected:
+  /// Planners
+  std::vector<PlannerPtr> mPlanners;
+};
+
+} // namespace planner
+} // namespace aikido
+
+#endif // AIKIDO_PLANNER_METAPLANNER_HPP_
--- a/include/aikido/planner/PlanToConfiguration.hpp
+++ b/include/aikido/planner/PlanToConfiguration.hpp
@@ -1,0 +1,85 @@
+#ifndef AIKIDO_PLANNER_PLANTOCONFIGURATION_HPP_
+#define AIKIDO_PLANNER_PLANTOCONFIGURATION_HPP_
+
+#include "aikido/constraint/Testable.hpp"
+#include "aikido/planner/Problem.hpp"
+#include "aikido/statespace/StateSpace.hpp"
+#include "aikido/trajectory/Interpolated.hpp"
+
+namespace aikido {
+namespace planner {
+
+/// Planning problem to plan to a single goal configuration.
+///
+/// Plan a trajectory from start state to goal state by using an interpolator to
+/// interpolate between them.
+class PlanToConfiguration : public Problem
+{
+public:
+  class Result;
+
+  /// Constructor.
+  ///
+  /// \param stateSpace State space.
+  /// \param startState Start state.
+  /// \param goalState Goal state.
+  /// \param interpolator Interpolator used to produce the output trajectory.
+  /// \param constraint Trajectory-wide constraint that must be satisfied.
+  /// \throw If \c stateSpace is not compatible to \c constraint's state space.
+  PlanToConfiguration(
+      statespace::StateSpacePtr stateSpace,
+      const statespace::StateSpace::State* startState,
+      const statespace::StateSpace::State* goalState,
+      statespace::InterpolatorPtr interpolator,
+      constraint::TestablePtr constraint);
+
+  /// Returns the name of the planner problem.
+  const std::string& getName() const override;
+  static const std::string& getStaticName();
+
+  /// Returns the start state.
+  const statespace::StateSpace::State* getStartState() const;
+
+  /// Returns the goal state.
+  const statespace::StateSpace::State* getGoalState() const;
+
+  /// Returns the interpolator used to produce the output trajectory.
+  statespace::InterpolatorPtr getInterpolator();
+
+  /// Returns the interpolator used to produce the output trajectory.
+  statespace::InterpolatorPtr getInterpolator() const;
+  // TODO: Should return ConstInterpolatorPtr when resolving const correctness
+  // issues.
+
+  /// Returns the constraint that must be satisfied throughout the trajectory.
+  constraint::TestablePtr getConstraint();
+
+  /// Returns the constraint that must be satisfied throughout the trajectory.
+  constraint::TestablePtr getConstraint() const;
+  // TODO: Should return ConstInterpolatorPtr when resolving const correctness
+  // issues.
+
+protected:
+  /// Start state.
+  const statespace::StateSpace::State* mStartState;
+
+  /// Goal state.
+  const statespace::StateSpace::State* mGoalState;
+
+  /// Interpolator used to produce the output trajectory.
+  statespace::InterpolatorPtr mInterpolator;
+
+  /// Trajectory-wide constraint that must be satisfied.
+  constraint::TestablePtr mConstraint;
+};
+
+class PlanToConfiguration::Result : public Problem::Result
+{
+public:
+protected:
+};
+
+} // namespace planner
+} // namespace aikido
+
+#endif
--- a/include/aikido/planner/PlanToConfigurations.hpp
+++ b/include/aikido/planner/PlanToConfigurations.hpp
@@ -1,0 +1,81 @@
+#ifndef AIKIDO_PLANNER_PLANTOCONFIGURATIONS_HPP_
+#define AIKIDO_PLANNER_PLANTOCONFIGURATIONS_HPP_
+
+#include "aikido/constraint/Testable.hpp"
+#include "aikido/planner/Problem.hpp"
+#include "aikido/statespace/StateSpace.hpp"
+#include "aikido/statespace/Interpolator.hpp"
+
+namespace aikido {
+namespace planner {
+
+/// Planning problem to plan to a multiple goal configurations.
+///
+/// Plan a trajectory from start state to any of the goal states using an
+/// interpolator to interpolate between the states.
+class PlanToConfigurations : public Problem
+{
+public:
+  class Result;
+
+  /// Constructor.
+  ///
+  /// \param stateSpace State space.
+  /// \param startState Start state.
+  /// \param goalStates Goal states.
+  /// \param interpolator Interpolator used to produce the output trajectory.
+  /// \param constraint Trajectory-wide constraint that must be satisfied.
+  /// \throw If \c stateSpace is not compatible to \c constraint's state space.
+  PlanToConfigurations(
+      statespace::StateSpacePtr stateSpace,
+      const statespace::StateSpace::State* startState,
+      const std::vector<statespace::StateSpace::State*> goalStates,
+      statespace::InterpolatorPtr interpolator,
+      constraint::TestablePtr constraint);
+
+  /// Returns the name of the planner problem.
+  const std::string& getName() const override;
+  static const std::string& getStaticName();
+
+  /// Returns the start state.
+  const statespace::StateSpace::State* getStartState() const;
+
+  /// Returns the vector of goal states.
+  const std::vector<statespace::StateSpace::State*> getGoalStates() const;
+
+  /// Returns the interpolator used to produce the output trajectory.
+  statespace::InterpolatorPtr getInterpolator();
+
+  /// Returns the interpolator used to produce the output trajectory.
+  statespace::ConstInterpolatorPtr getInterpolator() const;
+
+  /// Returns the constraint that must be satisfied throughout the trajectory.
+  constraint::TestablePtr getConstraint();
+
+  /// Returns the constraint that must be satisfied throughout the trajectory.
+  constraint::ConstTestablePtr getConstraint() const;
+
+protected:
+  /// Start state.
+  const statespace::StateSpace::State* mStartState;
+
+  /// Goal States.
+  const std::vector<statespace::StateSpace::State*> mGoalStates;
+
+  /// Interpolator used to produce the output trajectory.
+  statespace::InterpolatorPtr mInterpolator;
+
+  /// Trajectory-wide constrained that must be satisfied.
+  constraint::TestablePtr mConstraint;
+};
+
+class PlanToConfigurations::Result : public Problem::Result
+{
+public:
+protected:
+};
+
+} // namespace planner
+} // namespace aikido
+
+#endif
--- a/include/aikido/planner/PlanToEndEffectorOffset.hpp
+++ b/include/aikido/planner/PlanToEndEffectorOffset.hpp
@@ -1,0 +1,93 @@
+#ifndef AIKIDO_PLANNER_PLANTOENDEFFECTOROFFSET_HPP_
+#define AIKIDO_PLANNER_PLANTOENDEFFECTOROFFSET_HPP_
+
+#include <dart/dart.hpp>
+#include "aikido/planner/Problem.hpp"
+#include "aikido/statespace/StateSpace.hpp"
+#include "aikido/trajectory/Interpolated.hpp"
+#include "aikido/constraint/Testable.hpp"
+
+namespace aikido {
+namespace planner {
+
+/// Planning problem to plan to desired en-effector offset with fixed
+/// orientation
+class PlanToEndEffectorOffset : public Problem
+{
+public:
+  class Result;
+
+  /// Constructor.
+  ///
+  /// \param stateSpace State space.
+  /// \param bodyNode Body Node or robot for which the path is to be planned.
+  /// \param startState Start state.
+  /// \param goalState Goal state.
+  /// \param direction Direction of motion [unit vector in the world frame].
+  /// \param distance Distance to move, in meters.
+  /// \param interpolator Interpolator used to produce the output trajectory.
+  /// \param constraint Trajectory-wide constraint that must be satisfied.
+  /// \throw If \c stateSpace is not compatible to \c constraint's state space.
+  PlanToEndEffectorOffset(
+      statespace::StateSpacePtr stateSpace,
+      dart::dynamics::BodyNodePtr bodyNode,
+      const statespace::StateSpace::State* startState,
+      const Eigen::Vector3d& direction,
+      const double& distance,
+      statespace::InterpolatorPtr interpolator,
+      constraint::TestablePtr constraint);
+
+  /// Returns the name of the planner problem.
+  const std::string& getName() const override;
+  static const std::string& getStaticName();
+
+  /// Returns the body node or robot for which the path is to be planned.
+  dart::dynamics::BodyNodePtr getBodyNode();
+
+  /// Returns the start state.
+  const statespace::StateSpace::State* getStartState() const;
+
+  /// Returns the direction of motion specified in the world frame.
+  const Eigen::Vector3d& getDirection() const;
+
+  /// Returns the distance to move in the specified direction [meters]
+  const double& getDistance() const;
+
+  /// Returns the interpolator used to produce the output trajectory.
+  statespace::InterpolatorPtr getInterpolator();
+  statespace::ConstInterpolatorPtr getInterpolator() const;
+
+  /// Returns the constraint that must be satisfied throughout the trajectory.
+  constraint::TestablePtr getConstraint();
+  constraint::ConstTestablePtr getConstraint() const;
+
+protected:
+  /// Body Node or Robot.
+  dart::dynamics::BodyNodePtr mBodyNode;
+
+  /// Start state.
+  const statespace::StateSpace::State* mStartState;
+
+  /// Direction of motion.
+  const Eigen::Vector3d mDirection;
+
+  /// Distance to move in the direction specified.
+  const double mDistance;
+
+  /// Interpolator used to produce the output trajectory.
+  statespace::InterpolatorPtr mInterpolator;
+
+  /// Trajectory-wide constraint that must be satisfied.
+  constraint::TestablePtr mConstraint;
+};
+
+class PlanToEndEffectorOffset::Result : public Problem::Result
+{
+public:
+protected:
+};
+
+} // namespace planner
+} // namespace aikido
+
+#endif
--- a/include/aikido/planner/PlanToEndEffectorPose.hpp
+++ b/include/aikido/planner/PlanToEndEffectorPose.hpp
@@ -1,0 +1,86 @@
+#ifndef AIKIDO_PLANNER_PLANTOENDEFFECTORPOSE_HPP_
+#define AIKIDO_PLANNER_PLANTOENDEFFECTORPOSE_HPP_
+
+#include <dart/dart.hpp>
+#include "aikido/planner/Problem.hpp"
+#include "aikido/statespace/StateSpace.hpp"
+#include "aikido/trajectory/Interpolated.hpp"
+#include "aikido/constraint/Testable.hpp"
+
+namespace aikido {
+namespace planner {
+
+/// Planning problem to plan to a given end effector pose.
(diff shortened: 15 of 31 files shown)


mkoval:
Thanks to @jslee02 for putting together this draft. It's a great start and is very helpful for driving the discussion.

This is actually quite a bit more sophisticated than what I had in mind. I wrote up my original thoughts from back when we started Aikido below. I intend this to be the starting point of further discussion - I am not trying to prescribe this as the only solution. I am very interested in hearing what drove the design decisions in this pull request, since the devil is often in the details of the implementation.

---

I was originally envisioning four "layers" to the planning interface:

1. **The planner itself,** which may comply with no standard interface. The goal of this API is to expose the _full functionality_ of the planner while (a) a providing a clean API, (b) using standard Aikido types wherever possible, and (c) introducing minimal overhead. Often, this will be as simple as a free function. All of the planning functions in Aikido today fall into this layer, e.g. `planOMPL()`, `planCRRTConnect()`, and `computeParabolicTiming()`.

2. **A planner interface** for each type of query. Instead of having a singular `Planner` interface, there would be a separate interface for each type of query. We would define a new interface if: (a) the query is common in our application code (e.g. MAGI actions) and (b) more than one planner can solve the query, i.e. it is possible to implement more than one adapter class as outlined in (3) below.

3. **An adapter class,** that uses (1) to solve queries specified in a standard format. Any arguments that are not part of the planner interface, e.g. planner-specific resolution parameters, should be specified by the adapter's constructor. Often, this involves converting the query into the type of constraints that the planner accepts. For example, an adapter class could solve a "plan to a goal configuration" query by calling `planOMPL()` with a `FiniteSampleable` goal constraint.

4. **A meta-planner** implements an adapter class using another planner interface, i.e. it converts from one planner interface to another. For example, a meta-planner may convert an arbitrary `PlanToConfigurations` planner to a `PlanToConfiguration` planner by calling `PlanToConfigurations` on a list that contains the singular goal configuration.

If we need a meta planner that operates on multiple types of planning queries, e.g. by calling multiple planners in sequence, then we can implement that as a template class. The details here are are a little hairy, so we need to think carefully about how we define the planner interfaces to make this possible.

---

These are the general principles that drove this design:

- It should always be possible to access the full power of a planner. Implementing an interface should never reduce a planner's functionality.
- An interface only has value if all implementations of the interface can be used interchangeably (sometimes called the [Liskov substitution principle](https://en.wikipedia.org/wiki/Liskov_substitution_principle)).
- Every planning query has a different interface, so it's not possible to use them interchangeably.
- It's useful to enforce the type on arguments and return values of planning functions at compile time.
- Trying to implement a generic `Planner` base class requires bypassing most of the C++ type system. If that is necessary, you should consider writing that code in Python instead. :winking:

Looking at the current implementation in this pull request, I think there are a few great ideas that we could add to the mix:

- **`Problem` classes** provide a nice mechanism for defining arguments. Inheritance could come in handy here once we have a wider variety of planner interfaces. For example, we could write helper code for a lot of boilerplate `StateSpace`, `DistanceMetric`, etc initialization. It also is a nice workaround for C++'s lack of named initialization lists.
- **`Result` classes** provide a nice mechanism for annotating the output of planners with planner-specific information. This has many of the same advantages as `Outcome` did in https://github.com/personalrobotics/aikido/pull/266.

---

Here is a more concrete example in C++-like pseudocode:

```c++
// 1. Planner
trajectory::InterpolatedUniquePtr planOMPL(
  const statespace::StateSpace::State* _start,
  const statespace::StateSpace::State* _goal,
  statespace::StateSpacePtr _stateSpace,
  statespace::InterpolatorPtr _interpolator,
  distance::DistanceMetricPtr _dmetric,
  constraint::SampleablePtr _sampler,
  constraint::TestablePtr _validityConstraint,
  constraint::TestablePtr _boundsConstraint,
  constraint::ProjectablePtr _boundsProjector,
  double _maxPlanTime,
  double _maxDistanceBtwValidityChecks);

// 2. Planner Interfaces
class FromConfigurationToConfigurationPlanner
{
public:
  virtual trajectory::TrajectoryUniquePtr planFromConfigurationToConfiguration(
    const statespace::StateSpacePtr& stateSpace,
    const statespace::StateSpace::State* startState,
    const statespace::StateSpace::State* goalState) = 0;

};

class FromConfigurationToConfigurationsPlanner
{
public:
  virtual trajectory::TrajectoryUniquePtr planFromConfigurationToConfigurations(
    const statespace::StateSpacePtr& stateSpace,
    const statespace::StateSpace::State* startState,
    std::vector<const statespace::StateSpace::State*> goalStates) = 0;
};


// 3. Adapter Interfaces
class OMPLFromConfigurationsToConfigurationsPlanner
  : public virtual FromConfigurationToConfigurationsPlanner
{
public:
  virtual trajectory::TrajectoryUniquePtr planFromConfigurationToConfigurations(
    const statespace::StateSpacePtr& stateSpace,
    const statespace::StateSpace::State* startState,
    std::vector<const statespace::StateSpace::State*> goalStates)
  {
    return planOMPL(...);
  }
};


// 4. Meta-Planner
class FromConfigurationsToConfigurations_to_FromConfigurationToConfiguration
  : public virtual FromConfigurationToConfigurationPlanner
{
public:
  explicit FromConfigurationsToConfigurations_to_FromConfigurationToConfiguration(
      std::unique_ptr<OMPLFromConfigurationsToConfigurationsPlanner> delegatePlanner)
    : mDelegatePlanner{std::move(delegatePlanner)}
  {
  }

  virtual trajectory::TrajectoryUniquePtr planFromConfigurationToConfiguration(
    const statespace::StateSpacePtr& stateSpace,
    const statespace::StateSpace::State* startState,
    const statespace::StateSpace::State* goalState)
  {
    return mDelegatePlanner->planFromConfigurationToConfigurations(
      stateSpace, startState, {goalState});
  }
};

```

jslee02:
@mkoval Thanks for the suggestion. I think the major differences between our designs are:
(1) Having base planner class (i.e., `Planner`) vs. not
(2) Having all the solvable planning methods in a planner vs. having them as separate classes

More specifically, I propose having a planner class and putting all the functions of the solvable planning problems (methods) into the planner class. So we might only have like `SnapPlanner`, `VectorFieldPlanner`, `OmplPlanner`, `TrajOptPlanner`, and so on; and multiple planning functions will be in each class. In your design, we would have more classes which are the permutation of planners and planning problems, and single planning function in each class.

Here are some thoughts on possible advantages of this PR's implementation that I believe:

Having planning functions in a specific planner class would be able to share reusable functionality and data in the same class. I believe those planning functions in a planner class have a similar fashion of solving the problems. For example, `TrajOptPlanner` would solve all the problems solving optimization problems. It would be still possible with your design by having an additional class(es) and composing it into the necessary adaptor classes, but I think this could introduce unnecessary complex (I could be biased).

Another possible advantage is that the base planner class can provide useful interface such as `canSolve(problem)`, `solve(problem)`, and `getSolvableProblems()`. These functions allow you to get the capability of the planner. This functionality would be able to be achieved with your design by adding them to meta-planner, but my proposal would enable to implement meta-planners straightforwardly with the virtual functions of the base planner class.

I'm not sure how to incorporate `Problem` classes into your design because it seems the planner interfaces are the problem classes to me. Probably we have different definitions of `Problem` classes in each mind? :sweat_smile:

mkoval:
I think your summary of the design differences is accurate. 👍 

---

> In your design, we would have more classes which are the permutation of planners and planning problems, **and [a] single planning function in each class.**

This is not strictly required. If the implementation of multiple planning interfaces for a planner are similar, then you can implement them in the same class. However, I do think that this will end up being relatively rare in practice - I think an adapter class is often the better way to go.

> So we might only have like `SnapPlanner`, `VectorFieldPlanner`, `OmplPlanner`, `TrajOptPlanner`, and so on; and multiple planning functions will be in each class.

This is the approach that we used in PrPy. We quickly discovered that it is difficult to handle planner-specific parameters. Multiple planners may implement the same planning method (e.g. `PlanToConfiguration`), but take different parameters. For example: some planners require time limits, others require heuristics, etc.

With my proposed API, planner-specific parameters (e.g. time limits) are passed into the adapter class' constructor. The planning method itself always take only the arguments defined by the planner interface(s) that it implements.

This is critical when planners are used via dependency injection. For example, a MAGI action can take an `std::unique_ptr<PlanToEndEffectorOffset>` without caring which specific planner it is using or what extra `**kwargs` it needs to pass to it.

> Having planning functions in a specific planner class would be able to share reusable functionality and data in the same class. [...] **It would be still possible with your design by having an additional class(es)** and composing it into the necessary adaptor classes, but I think this could introduce unnecessary complex (I could be biased).

This is exactly how I propose to solve the re-use problem. Nothing about defining a variety of interfaces prohibits us from defining helper functions and/or classes. I suspect that the resulting code would end up being more reusable and easier to unit test, since it separates the helper code from the implementation of the planner. (This is largely my gut reaction, so I don't want to start a holy war. 😬)

> Another possible advantage is that the base planner class can provide useful interface such as `canSolve(problem)`, `solve(problem)`, and `getSolvableProblems()`. These functions allow you to get the capability of the planner. 

This is a neat idea. This seems most useful for allowing a `MetaPlanner` to have the upmost flexibility in deciding which planner to apply to a problem. The disadvantage is that you lose a lot of nice type safety enforced by the C++ type system. My approach is much more rigid because the list of queries is specified as C++ types at compile time, but has that type safety.

While it has a lot of potential, I am not sure how useful this level of generality will be in practice today. We have relatively small sets of planners and types of planning queries. However, I do entirely understand the appeal of this - I believe this is actually pretty similar to the design that @cdellin and the NREC folks used for motion planning on CHIMP in the DRC.

---

In general, I was trying to follow the philosophy that Aikido is a toolbox of small, reusable pieces of code that can stitched together to solve a variety of planning problems. I really like the idea of adapter classes because it is so [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself). Uou can write (and test!) a piece of code once, rather than repeating it once per planner.

For example, any planner that implements `PlanFromConfigurationsToConfigurations` would automatically get `PlanToConfigurationFromConfigurations`, `PlanFromConfigurationsToConfiguration`, and `PlanToConfigurationToConfiguration` for free. You could also convert any planner that implements `PlanToConfiguration` to `PlanToTSR` by sampling from the TSR and calling `PlanToConfiguration` (this is what `TSRPlanner` does in PrPy).

This is similar to how we defined `NewtonsMethodProjectable` once, rather than re-implementing a rootfinder every time we need to do constraint projection.

---

I am interested in hearing your thoughts on this @jslee02 - I'm open to be convinced otherwise. 😈

siddhss5:
At a meta-level, we seem to be discussing the relative merits of composition over inheritance. @mkoval and @jslee02 are way better at discussing this than I am, so I'll defer to them, except point out that composition often enables more flexibility whereas inheritance often enables more readability.

I want to emphasize one comment that @mkoval made that I think is critical for any path we pursue:

> This is the approach that we used in PrPy. We quickly discovered that it is difficult to handle planner-specific parameters. Multiple planners may implement the same planning method (e.g. PlanToConfiguration), but take different parameters. For example: some planners require time limits, others require heuristics, etc.

This has annoyed me [maybe us all] no end. Although I am a fan of having some structure in what a planner can do [and I think the `PlanTo*` structure we came up with in `prpy` is really fantastic], I always felt like the way it was implemented in `prpy` really really straightjacketed us in accessing the planners. I suppose one way to address this is how `scipy` and much of `MATLAB` address this is via an open-ended `options` dictionary. But then that demands a _lot_ of documentation and can easily go very wrong.

gilwoolee:
Hi all, thanks for the great suggestions. I think both designs are flexible and capable to handle the variety of problems we have. In terms of readability, I slightly prefer @jslee02 's proposal. However, I think there's one major advantage of @mkoval 's proposal that may be a bit complicated to handle in @jslee02 's current design:

> With my proposed API, planner-specific parameters (e.g. time limits) are passed into the adapter class' constructor. The planning method itself always take only the arguments defined by the planner interface(s) that it implements. This is critical when planners are used via dependency injection. For example, a MAGI action can take an std::unique_ptr<PlanToEndEffectorOffset> without caring which specific planner it is using or what extra **kwargs it needs to pass to it.

These are in @jslee02 's version `problem x planner`-specific parameters, so it cannot be passed in at the constructor of @jslee02 's planner (e.g. `OMPLPlanner`). @jslee02 and I had a brief discussion on how to handle this, e.g. via DART's Aspect functionality, which (to my best understanding) is a sophisticated version of an open-ended `options`-dictionary.  The amount of documentation may not be too different from what we may end up doing with @mkoval 's design, but given that we have to take such an approach, I think being able to rely on C++'s type safety is more appealing.

jslee02:
> This is the approach that we used in PrPy. We quickly discovered that it is difficult to handle planner-specific parameters. Multiple planners may implement the same planning method (e.g. PlanToConfiguration), but take different parameters. For example: some planners require time limits, others require heuristics, etc.

As @gilwoolee mentioned, we could use DART's aspect mechanism (of @mxgrey) to achieve that. In our case, `[~]Problem` will be `Composite`s, and each additional planner specific parameters will be in `Aspect`s, which will be defined in each planner. As so, the planner can use the aspect when it's added to the problem; otherwise, use the defaults. I believe `Aspect`-`Component` is designed to handle this kind of problems.

> (@mkoval's comment) In general, I was trying to follow the philosophy that Aikido is a toolbox of small, reusable pieces of code that can stitched together to solve a variety of planning problems. I really like the idea of adapter classes because it is so DRY. Uou can write (and test!) a piece of code once, rather than repeating it once per planner.
> (@siddhss5's comment) At a meta-level, we seem to be discussing the relative merits of composition over inheritance. @mkoval and @jslee02 are way better at discussing this than I am, so I'll defer to them, except point out that composition often enables more flexibility whereas inheritance often enables more readability.

I also prefer DRY, and it can be achieved in my design as well. `[X]Problem` class could have reusable common functionalities, which are the X planning problem specific. I believe the problem classes are the same with the planner interface in your pseudo code in that sense; except (1) the planning functions are defined in planner classes in my design, and (2) problem classes are derived as adapter planners for each planning methods in your design.

Overall, I believe both designs don't have any major technical issue to achieve certain flexibility we want to obtain, but I think my design would be more straightforward to implement the planner capability checking functionality (e.g., `canSolve(problem)`, `solve(problem)`, and `getSolvableProblems()`).

I think it would be helpful to compare and improve the design ideas here (or in gist). Let me write down sample code soon. :)

jslee02:
@brianhou, @gilwoolee, and I discussed further, and I think we can reach a compromise. Essentially, it would be based on your _Planner Interfaces_ and incorporate `Problem` classes and `canSolve`/`getSolveableProblems` into that.

Here is a high-level picture I have:

**Summary**
* Define planner interfaces per planning problems
* Define adapter interfaces
* Pass `Problem` classes to the planner interfaces
* Add `canSolve` and `getSolveableProblems` to `Robot` class
* Make a concrete robot class (e.g., `Herb`) inherit `Robot` class and planner interfaces that the robot want to support

Please let me know what you guys think or what is not clear. If this sounds good, I will make the changes in this PR.

jslee02:
I've revised the planner API according to the discussion (+ ideas came up with while implementing it). Let me know what do you think. Here is a rough summary of the changes.

**Note on the Changes**
- Kept `Planner` class to provide `canPlan`, `getPlannableProblems`, and `plan`, which takes the base problem class
- Added `PlannerForSingleProblem : Planner` which is a base class of _the planner interface_ classes. It is templatized for a planning problem and implements all the virtual function of `Planner` class.
- Added `PlannerForMultiProblem : Planner` which is the base class of concrete planner class that can solve more than one planning problem
- Revised snap planner (i.e., `SnapConfigurationToConfigurationPlanner`) to inherit `PlannerForSingleProblem`
- Renamed planning problem classes in this fashion: <start_type>To<goal_type>. For example, `ConfigurationToConfiguration`
- Renamed planner classes in the same fashion of planning problem but with '-Planner` appendix.

jslee02:
@mkoval @psigen Could you guys take a look the latest changes if it makes sense when you have a chance? :smile: Thanks!

include/aikido/planner/SnapConfigurationToConfigurationPlanner.hpp
```diff
@@ -0,0 +1,44 @@
+#ifndef AIKIDO_PLANNER_SNAPCONFIGURATIONTOCONFIGURATIONPLANNER_HPP_
+#define AIKIDO_PLANNER_SNAPCONFIGURATIONTOCONFIGURATIONPLANNER_HPP_
+
+#include "aikido/planner/ConfigurationToConfiguration.hpp"
+#include "aikido/planner/ConfigurationToConfigurationPlanner.hpp"
+
+namespace aikido {
+namespace planner {
+
+/// Planner that checks the straight-line trajectory to the goal.
```
gilwoolee: `checks` -> `plans`

include/aikido/planner/ConfigurationToEndEffectorOffset.hpp
```diff
@@ -0,0 +1,85 @@
...
+
+/// Planning problem to plan to desired en-effector offset with fixed
+/// orientation
+class ConfigurationToEndEffectorOffset : public Problem
+{
+public:
+  /// Constructor.
+  ///
+  /// \param stateSpace State space.
+  /// \param bodyNode Body Node or robot for which the path is to be planned.
```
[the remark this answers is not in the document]
brianhou: Is this code written this way because we think it'd be valuable for a robot (e.g. RACECAR) that doesn't have an end-effector to be able to plan to an offset in its current direction?

include/aikido/planner/ConfigurationToConfiguration.hpp
```diff
@@ -0,0 +1,85 @@
...
+      statespace::InterpolatorPtr interpolator,
+      constraint::TestablePtr constraint);
+
+  // Documentation inherited.
+  const std::string& getName() const override;
+
+  /// Returns the name of the planner problem.
+  static const std::string& getStaticName();
+
+  /// Sets the start state.
```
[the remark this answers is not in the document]
aditya-vk: For Single Source Single Target, I agree that we _could_ create new problems. I don't see any problem with reusing by changing the start and goal though. For instance, at least the current implementation of `planToConfigurations` under `robot/util.cpp` is just iterating through each goal state and calling `planToConfiguration(goal)`. If we are doing that here as well, we could just `setGoalState` instead of creating a new instance everytime.

include/aikido/planner/ConfigurationToConfiguration.hpp
```diff
@@ -0,0 +1,85 @@
...
+  void setStartState(const statespace::StateSpace::State* state);
+
+  /// Returns the start state.
+  const statespace::StateSpace::State* getStartState() const;
+
+  /// Sets the goal state.
+  void setGoalState(const statespace::StateSpace::State* state);
```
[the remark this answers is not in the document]
aditya-vk: Same as above.

include/aikido/planner/ConfigurationToConfigurationPlanner.hpp
```diff
@@ -0,0 +1,28 @@
...
+
+class ConfigurationToConfigurationPlanner
+    : public PlannerForSingleProblem<ConfigurationToConfigurationPlanner,
+                                     ConfigurationToConfiguration>
+{
+public:
+  using TheProblem = ConfigurationToConfiguration;
+  using TheResult = Result;
+
+  virtual trajectory::TrajectoryPtr plan(
```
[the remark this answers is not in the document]
aditya-vk: Should we have `SingleSourceSingleTargetPlanner`, `SingleSourceMultipleTargetsPlanner` etc.?
Similarly for problems. The fact that the planning problem/planner names differ by a single letter is super confusing. No?

include/aikido/planner/ConfigurationToConfigurations.hpp
```diff
@@ -0,0 +1,73 @@
...
+{
+public:
+  /// Constructor.
+  ///
+  /// \param stateSpace State space.
+  /// \param startState Start state.
+  /// \param goalStates Goal states.
+  /// \param interpolator Interpolator used to produce the output trajectory.
+  /// \param constraint Trajectory-wide constraint that must be satisfied.
+  /// \throw If \c stateSpace is not compatible to \c constraint's state space.
```
aditya-vk: nit: compatible with?

include/aikido/planner/ConfigurationToConfigurations.hpp
```diff
@@ -0,0 +1,73 @@
...
+      statespace::StateSpacePtr stateSpace,
+      const statespace::StateSpace::State* startState,
+      const std::vector<statespace::StateSpace::State*>& goalStates,
+      statespace::InterpolatorPtr interpolator,
+      constraint::TestablePtr constraint);
+
+  /// Returns the name of the planner problem.
+  const std::string& getName() const override;
+  static const std::string& getStaticName();
+
```
aditya-vk: What about `addGoal()` method?

include/aikido/planner/ConfigurationToConfigurations.hpp
```diff
@@ -0,0 +1,73 @@
...
+  /// \param stateSpace State space.
+  /// \param startState Start state.
+  /// \param goalStates Goal states.
+  /// \param interpolator Interpolator used to produce the output trajectory.
+  /// \param constraint Trajectory-wide constraint that must be satisfied.
+  /// \throw If \c stateSpace is not compatible to \c constraint's state space.
+  ConfigurationToConfigurations(
+      statespace::StateSpacePtr stateSpace,
+      const statespace::StateSpace::State* startState,
+      const std::vector<statespace::StateSpace::State*>& goalStates,
```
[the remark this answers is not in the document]
aditya-vk: I agree with @brianhou that we would need a set for fast lookup. From a performance perspective, the whole point of having this feature, to my understanding, is to parallelly search for multiple goals and terminate when path to one of them is found. This implies, we would ideally not want to _iterate_ through each goal and plan. I think if done right, this would need a set and not a vector. Correct me if I am wrong.

include/aikido/planner/ConfigurationToConfigurations.hpp
```diff
@@ -0,0 +1,73 @@
...
+  statespace::InterpolatorPtr getInterpolator();
+
+  /// Returns the interpolator used to produce the output trajectory.
+  statespace::ConstInterpolatorPtr getInterpolator() const;
+
+  /// Returns the constraint that must be satisfied throughout the trajectory.
+  constraint::TestablePtr getConstraint();
+
+  /// Returns the constraint that must be satisfied throughout the trajectory.
+  constraint::ConstTestablePtr getConstraint() const;
```
aditya-vk: Can we have this in `PlanConfigurationToConfiguration` problem as well.

include/aikido/planner/ConfigurationToConfigurations.hpp
```diff
@@ -0,0 +1,73 @@
...
+  const statespace::StateSpace::State* mStartState;
+
+  /// Goal States.
+  const std::vector<statespace::StateSpace::State*> mGoalStates;
+
+  /// Interpolator used to produce the output trajectory.
+  statespace::InterpolatorPtr mInterpolator;
+
+  /// Trajectory-wide constrained that must be satisfied.
+  constraint::TestablePtr mConstraint;
```
aditya-vk: constraint should be `const` during planning.

include/aikido/planner/ConfigurationToConfigurations.hpp
```diff
@@ -0,0 +1,73 @@
...
+
+  /// Returns the constraint that must be satisfied throughout the trajectory.
+  constraint::ConstTestablePtr getConstraint() const;
+
+protected:
+  /// Start state.
+  const statespace::StateSpace::State* mStartState;
+
+  /// Goal States.
+  const std::vector<statespace::StateSpace::State*> mGoalStates;
```
aditya-vk: If we have an `addGoal()` this should not be `const`. Instead it should be `std::vector<const --> mGoalStates` which should be the case nevertheless.

include/aikido/planner/ConfigurationToEndEffectorOffset.hpp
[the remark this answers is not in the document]
aditya-vk: Maybe change the name of the problem also in that case? End Effector is very much manipulator specific. We'd like the planners and problems to be generic no? Maybe `SingleSourceSingleTargetFixedOrientationProblem`. Took me a while to type that name. :P [Still not a good enough name though]

include/aikido/planner/ConfigurationToEndEffectorOffset.hpp
```diff
@@ -0,0 +1,85 @@
...
+  /// Returns the name of the planner problem.
+  const std::string& getName() const override;
+  static const std::string& getStaticName();
+
+  /// Returns the body node or robot for which the path is to be planned.
+  dart::dynamics::BodyNodePtr getBodyNode();
+
+  /// Returns the start state.
+  const statespace::StateSpace::State* getStartState() const;
+
```
aditya-vk: Like @brianhou mentioned, none of the problems except for `ConfigurationToConfiguration` have setters for start and goal.

include/aikido/planner/ConfigurationToEndEffectorOffset.hpp
```diff
@@ -0,0 +1,85 @@
...
+  const Eigen::Vector3d mDirection;
+
+  /// Distance to move in the direction specified.
+  const double mDistance;
+
+  /// Interpolator used to produce the output trajectory.
+  statespace::InterpolatorPtr mInterpolator;
+
+  /// Trajectory-wide constraint that must be satisfied.
+  constraint::TestablePtr mConstraint;
```
aditya-vk: Constraint should be `const`

include/aikido/planner/ConfigurationToEndEffectorPose.hpp
```diff
@@ -0,0 +1,84 @@
...
+  const statespace::StateSpace::State* mStartState;
+
+  /// Goal pose.
+  const Eigen::Isometry3d mGoalPose;
+
+  /// Interpolator used to produce the output trajectory.
+  statespace::InterpolatorPtr mInterpolator;
+
+  /// Trajectory-wide constraint that must be satisfied.
+  constraint::TestablePtr mConstraint;
```
aditya-vk: As before.

include/aikido/planner/ConfigurationToTSR.hpp
```diff
@@ -0,0 +1,72 @@
...
+  const statespace::StateSpace::State* mStartState;
+
+  /// Goal TSR
+  const constraint::dart::TSRPtr mGoalTSR;
+
+  /// Interpolator used to produce the output trajectory.
+  statespace::InterpolatorPtr mInterpolator;
+
+  /// Trajectory-wide constraint that must be satisfied.
+  constraint::TestablePtr mConstraint;
```
aditya-vk: As before.

include/aikido/planner/Planner.hpp
```diff
@@ -0,0 +1,55 @@
...
+AIKIDO_DECLARE_POINTERS(Planner)
+
+/// Base class for a meta-planner
+class Planner
+{
+public:
+  class Result;
+
+  /// Returns true if this planner can solve \c problem.
+  virtual bool canPlan(const Problem* problem) const = 0;
```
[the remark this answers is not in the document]
aditya-vk: I agree.

src/planner/ConfigurationToConfiguration.cpp
```diff
@@ -0,0 +1,95 @@
...
+}
+
+//==============================================================================
+constraint::TestablePtr ConfigurationToConfiguration::getConstraint()
+{
+  return mConstraint;
+}
+
+//==============================================================================
+constraint::TestablePtr ConfigurationToConfiguration::getConstraint() const
```
aditya-vk: Repeated code. Can we have this in the base problem class instead? All problems seem to be sharing the same implementations. At least the const versions can be implemented in the base class.

src/planner/ConfigurationToConfigurations.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+//==============================================================================
+const statespace::StateSpace::State*
+ConfigurationToConfigurations::getStartState() const
+{
+  return mStartState;
+}
+
+//==============================================================================
+const std::vector<statespace::StateSpace::State*>
```
aditya-vk: `const std::vector<const statespace::StateSpace::State*>`

Although like I mentioned, we would want an `addGoal()` so the container cannot be `const` in that case.

src/planner/ConfigurationToConfigurations.cpp
```diff
@@ -0,0 +1,82 @@
...
+  return mConstraint;
+}
+
+//==============================================================================
+constraint::ConstTestablePtr ConfigurationToConfigurations::getConstraint()
+    const
+{
+  return mConstraint;
+}
+
```
aditya-vk: DRY.

include/aikido/planner/CompositePlanner.hpp
```diff
@@ -0,0 +1,62 @@
...
+  bool hasPlanner(const Planner* planner);
+
+  /// Returns all the planners in this CompositePlanner.
+  const std::vector<PlannerPtr>& getPlanners() const;
+
+  /// Returns planner given \c index.
+  ///
+  /// \throw If index is equal to or greater than the number of planners in this
+  /// CompositePlanner.
+  PlannerPtr getPlanner(std::size_t index);
```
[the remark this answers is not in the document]
jslee02: Your inference is correct. Let me add `getNumPlanners()`.

One additional rationale behind this is for const-correctness (again :laughing:). Since this class already has `getPlanners()`, it can be seemed `getPlanner()` is not necessary. But, `getPlanners()` should be non-const because it returns `std::vector<PlannerPtr>`, which is a list of non-const instances. We can't call this function from const instance of `CompositePlanner` unless we additionally maintain `std::vector<ConstPlannerPtr>`. This is a considerable approach, but for now, let's have const/non-const `getPlanner()` for simplicity.

include/aikido/planner/CompositePlanner.hpp
```diff
@@ -0,0 +1,62 @@
...
+
+  // Documentation inherited.
+  virtual bool canPlan(const Problem* problem) const override;
+
+  // Documentation inherited.
+  std::unordered_set<std::string> getPlannableProblems() const override;
```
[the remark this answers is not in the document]
jslee02: Removed this function entirely.

include/aikido/planner/ConfigurationToConfigurations.hpp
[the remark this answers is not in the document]
jslee02: Agreed. Changed to `std::unordered_set` instead of `std::set` because the order doesn't matter here.

src/planner/CompositePlanner.cpp
```diff
@@ -0,0 +1,118 @@
...
+  // Do nothing
+}
+
+//==============================================================================
+void CompositePlanner::addPlanner(PlannerPtr planner)
+{
+  if (std::find(mPlanners.begin(), mPlanners.end(), planner) != mPlanners.end())
+  {
+    dtwarn << "[CompositePlanner::addPlanner] Attempts to add a planner that "
+              "already exists in this CompositePlanner.\n";
```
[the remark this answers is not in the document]
jslee02: [`std::endl` is `\n` plus flushing](), which an unnecessary overhead I think.

src/planner/PlanningResult.cpp
```diff
@@ -0,0 +1,21 @@
+#include "aikido/planner/PlanningResult.hpp"
```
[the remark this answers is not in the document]
jslee02: That's true. This class will be removed once all the old planning functions that use this class are removed or converted to the new API.

I'll add a note to this class about this.

include/aikido/planner/ConfigurationToEndEffectorOffset.hpp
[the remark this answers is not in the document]
jslee02: I think this is just about confusing description. I renamed it to `endEffectorBodyNode` (and getters and setters accordingly) and improved the descriptions.

include/aikido/planner/CompositePlanner.hpp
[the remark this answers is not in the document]
brianhou: Sorry, I think I'm still missing something. Are you saying that `getPlanner` and `getNumPlanners` are necessary in order for a meta-planner to work? (And we aren't able to just iterate over the vector of planners because of this const-correctness issue? Why can't we just always have (one) vector of `ConstPlanner`s? Maybe a related question: when would we want to change a `Planner`?)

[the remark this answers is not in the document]
aditya-vk: There _can_ be instances when the planner is changing its parameters as it plans. If these params are modified as member variables, we might need `<non-const>Planner`. :thinking:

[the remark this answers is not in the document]
jslee02: @aditya-vk I also considered that case at first, but now I think the instances being changed as it plans should be mutable. This is because I think planner logically shouldn't need to change its internal states but just the returning trajectory and the result.

So I'm thinking of having const planners, and then having only `getPlanners()` should be okay.

src/planner/ConfigurationToConfiguration.cpp
jslee02: It sounds good. I'll move this function to the base then.

src/planner/ConfigurationToConfiguration.cpp
```diff
@@ -0,0 +1,95 @@
...
+//==============================================================================
+const std::string& ConfigurationToConfiguration::getName() const
+{
+  return getStaticName();
+}
+
+//==============================================================================
+const std::string& ConfigurationToConfiguration::getStaticName()
+{
+  static std::string name("ConfigurationToConfiguration");
```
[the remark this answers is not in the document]
jslee02: Since this is a static function, it cannot return a local variable.

include/aikido/planner/ConfigurationToConfigurationPlanner.hpp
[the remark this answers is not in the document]
jslee02: It was to differentiate the concrete problem the planner can solver from `Problem` (base class). Renamed to `SolverbleProblem`. Open to here better name. :smile:

include/aikido/planner/PlanningResult.hpp
```diff
@@ -6,9 +6,20 @@
 namespace aikido {
 namespace planner {
 
-struct PlanningResult
+class PlanningResult
```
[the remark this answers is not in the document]
jslee02: Yes. Let's remove once all the old planning API uses this class is removed.

include/aikido/planner/ConfigurationToConfigurationPlanner.hpp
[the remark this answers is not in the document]
brianhou: What about `ProblemType`?

include/aikido/planner/ConfigurationToConfiguration.hpp
```diff
@@ -0,0 +1,85 @@
...
+  statespace::InterpolatorPtr getInterpolator() const;
+  // TODO: Should return ConstInterpolatorPtr when resolving const correctness
+  // issues.
+
+  /// Returns the constraint that must be satisfied throughout the trajectory.
+  constraint::TestablePtr getConstraint();
+
+  /// Returns the constraint that must be satisfied throughout the trajectory.
+  constraint::TestablePtr getConstraint() const;
+  // TODO: Should return ConstInterpolatorPtr when resolving const correctness
```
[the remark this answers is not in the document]
jslee02: Removed this TODO as the issue is resolved.

jslee02:
I have revised this PR based on the online and offline discussions. @brianhou, @gilwoolee, @aditya-vk Second round of review, please. :smile:

include/aikido/planner/ConfigurationToEndEffectorOffset.hpp
```diff
@@ -0,0 +1,75 @@
...
+  /// trajectory.
+  /// \param[in] constraint Trajectory-wide constraint that must be satisfied.
+  /// \throw If \c stateSpace is not compatible with \c constraint's state
+  /// space.
+  ConfigurationToEndEffectorOffset(
+      statespace::StateSpacePtr stateSpace,
+      dart::dynamics::ConstBodyNodePtr endEffectorBodyNode,
+      const statespace::StateSpace::State* startState,
+      const Eigen::Vector3d& direction,
+      double distance,
```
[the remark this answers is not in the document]
jslee02: Passing const value is not preferred for [this reason](https://stackoverflow.com/a/46292715):

include/aikido/planner/RankedMetaPlanner.hpp
```diff
@@ -0,0 +1,20 @@
...
+
+namespace aikido {
+namespace planner {
+
+class RankedMetaPlanner : public CompositePlanner
+{
+public:
+  // Documentation inherited.
+  trajectory::TrajectoryPtr plan(
+      const Problem& problem, Result* result = nullptr) override;
```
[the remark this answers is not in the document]
jslee02: We can mark `final` specifier here when we're sure any derived class don't override the virtual function. I just leave it open because we haven't implemented this class yet.

include/aikido/statespace/GeodesicInterpolator.hpp
```diff
@@ -15,12 +15,12 @@ class GeodesicInterpolator : public Interpolator
   /// Constructs a \c GeodesicInterpolator for \c _stateSpace.
   ///
   /// \param _stateSpace used for interpolation
-  explicit GeodesicInterpolator(statespace::StateSpacePtr _stateSpace);
+  explicit GeodesicInterpolator(statespace::ConstStateSpacePtr _stateSpace);
```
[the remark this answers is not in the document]
jslee02: That's true. I just wouldn't prefer including those fixes in this PR to make this PR as small as possible.

include/aikido/planner/ConfigurationToConfigurationPlanner.hpp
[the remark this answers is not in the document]
jslee02: @brianhou I prefer `SolvableProblem` because it's more descriptive, but I'm also open to changing it to `ProblemType`.

include/aikido/planner/CompositePlanner.hpp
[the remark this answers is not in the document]
jslee02: Since we decided not to make `plan()` const, this function shouldn't be removed for these reasons:
* `getPlanner(index) const` for const iteration
* `getPlanner(index)` to provide consistent API with `getPlanner(index) const`.

But I'm open to removing `getPlanner(index)` because we could still do non-const iteration using `getPlanners()`.

include/aikido/planner/Problem.hpp
```diff
@@ -0,0 +1,50 @@
...
+  virtual ~Problem() = default;
+
+  /// Returns type of this planning problem.
+  virtual const std::string& getType() const = 0;
+
+  /// Returns const state space.
+  statespace::ConstStateSpacePtr getStateSpace() const;
+
+  /// Returns the constraint that must be satisfied throughout the trajectory.
+  constraint::ConstTestablePtr getConstraint() const;
```
gilwoolee: If the problem was created with nullptr, does it return a `SatisfiedPtr` or a nullptr?

include/aikido/planner/SequenceMetaPlanner.hpp
```diff
@@ -0,0 +1,20 @@
+#ifndef AIKIDO_PLANNER_SEQUENCEMETAPLANNER_HPP_
+#define AIKIDO_PLANNER_SEQUENCEMETAPLANNER_HPP_
+
+#include "aikido/planner/CompositePlanner.hpp"
+
+namespace aikido {
+namespace planner {
+
+class SequenceMetaPlanner : public CompositePlanner
```
gilwoolee: Class docstring? Is this for using a list of planners to solve one problem?

include/aikido/planner/ConfigurationToConfigurationPlanner.hpp
```diff
@@ -0,0 +1,32 @@
...
+{
+public:
+  /// Constructor
+  ///
+  /// \param[in] stateSpace State space that this planner associated with.
+  explicit ConfigurationToConfigurationPlanner(
+      statespace::ConstStateSpacePtr stateSpace);
+
+  /// Solves \c problem returning the result to \c result.
+  virtual trajectory::TrajectoryPtr plan(
```
gilwoolee: Can we have docstring for the params?

include/aikido/planner/ConfigurationToEndEffectorOffset.hpp
```diff
@@ -0,0 +1,75 @@
...
+{
+public:
+  /// Constructor.
+  ///
+  /// \param[in] stateSpace State space.
+  /// \param[in] endEffectorBodyNode BodyNode to be planned to move to a desired
+  /// offest while maintaining the current orientation.
+  /// \param[in] startState Start state.
+  /// \param[in] direction Direction of motion [unit vector in the world frame].
+  /// \param[in] distance Distance to move, in meters.
```
gilwoolee: Let's also mention that this must be nonnegative.

include/aikido/planner/Problem.hpp
jslee02: It returns `SatisfiedPtr`. Added note about this to the constructor.

include/aikido/planner/SequenceMetaPlanner.hpp
jslee02: This class is to solve multiple problems. If we want to have a composite planner to solve only one problem, then we could have
* `SingleProblemCompositePlanner<ProblemType>`
* `MultiProblemCompositePlanner` that is renamed from `CompositePlanner`.

include/aikido/planner/CompositePlanner.hpp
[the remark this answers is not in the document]
brianhou: If I remember our discussion correctly, I think we concluded that external users might not need to be able to iterate over the individual planners at all? I'm in favor of removing `getPlanner(index)` and `getNumPlanners`, and potentially adding them back if we find they're useful or necessary.

include/aikido/planner/ConfigurationToConfigurations.hpp
```diff
@@ -0,0 +1,62 @@
+#ifndef AIKIDO_PLANNER_PLANTOCONFIGURATIONS_HPP_
```
brianhou: Nit: `CONFIGURATIONTOCONFIGURATIONS`

include/aikido/planner/ConfigurationToEndEffectorOffset.hpp
```diff
@@ -0,0 +1,77 @@
+#ifndef AIKIDO_PLANNER_PLANTOENDEFFECTOROFFSET_HPP_
```
brianhou: Nit: `CONFIGURATIONTOENDEFFECTOROFFSET`.

include/aikido/planner/ConfigurationToEndEffectorPose.hpp
```diff
@@ -0,0 +1,64 @@
+#ifndef AIKIDO_PLANNER_PLANTOENDEFFECTORPOSE_HPP_
```
brianhou: Nit: `CONFIGURATIONTOENDEFFECTORPOSE`.

include/aikido/planner/ConfigurationToTSR.hpp
```diff
@@ -0,0 +1,63 @@
+#ifndef AIKIDO_PLANNER_PLANTOTSR_HPP_
```
brianhou: Nit: `CONFIGURATIONTOTSR`.

src/planner/ConfigurationToEndEffectorOffset.cpp
```diff
@@ -0,0 +1,69 @@
...
+  : Problem(std::move(stateSpace), std::move(constraint))
+  , mEndEffectorBodyNode(std::move(endEffectorBodyNode))
+  , mStartState(startState)
+  , mDirection(direction.normalized())
+  , mDistance(distance)
+{
+  if (direction.norm() < 1e-6)
+    throw std::invalid_argument("Length of direction shouldn't be zero.");
+
+  if (distance < 0.0)
```
brianhou: I personally prefer to be allowed to pass in a negative distance. We can flip the direction and distance in here, if our underlying planners would be unhappy with that.

src/planner/ConfigurationToEndEffectorOffset.cpp
```diff
@@ -0,0 +1,69 @@
...
+    const Eigen::Vector3d& direction,
+    const double distance,
+    constraint::ConstTestablePtr constraint)
+  : Problem(std::move(stateSpace), std::move(constraint))
+  , mEndEffectorBodyNode(std::move(endEffectorBodyNode))
+  , mStartState(startState)
+  , mDirection(direction.normalized())
+  , mDistance(distance)
+{
+  if (direction.norm() < 1e-6)
```
brianhou: What happens in line 19 if the norm is zero? If it can be successfully normalized, I think we should just allow it.

jslee02: According to the [Eigen documentation](https://eigen.tuxfamily.org/dox/classEigen_1_1MatrixBase.html#ad16303c47ba36f7a41ea264cb26bceb6), zero vector will be unchanged. Otherwise, it'll be successfully normalized. So I just changed this to compare to zero instead of `1e-6`.

If we allow zero vector, then I think we should document this as well.

jslee02:
The build failures on macOS should be fixed once #395 is merged.

include/aikido/planner/ConfigurationToEndEffectorOffset.hpp
```diff
@@ -0,0 +1,79 @@
+#ifndef AIKIDO_PLANNER_CONFIGURATIONTOCONFIGURATIONS_HPP_
```
gilwoolee: SHould be `AIKIDO_PLANNER_CONFIGURATIONTOENDEFFECTOROFFSET_HPP_`

include/aikido/planner/ConfigurationToEndEffectorOffset.hpp
```diff
@@ -0,0 +1,79 @@
...
+  /// offest while maintaining the current orientation.
+  /// \param[in] startState Start state.
+  /// \param[in] direction Unit vector that represents the direction of motion
+  /// [unit vector in the world frame]. The direction is allowed to be zero
+  /// vector.
+  /// \param[in] signedDistance Signed distance to move, in meters.
+  /// \param[in] interpolator Interpolator used to produce the output
+  /// trajectory.
+  /// \param[in] constraint Trajectory-wide constraint that must be satisfied.
+  /// \throw (1) If \c stateSpace is not compatible with \c constraint's state
```
gilwoolee: Why do we need (1)?

include/aikido/planner/ConfigurationToEndEffectorOffset.hpp
```diff
@@ -0,0 +1,79 @@
...
+  /// space.
+  ConfigurationToEndEffectorOffset(
+      statespace::ConstStateSpacePtr stateSpace,
+      dart::dynamics::ConstBodyNodePtr endEffectorBodyNode,
+      const statespace::StateSpace::State* startState,
+      const Eigen::Vector3d& direction,
+      double signedDistance,
```
gilwoolee: Does this mean that it can have both positive and negative? (We'll have to make sure this is supported in underlying planners.)

src/planner/CompositePlanner.cpp
```diff
@@ -0,0 +1,43 @@
...
+CompositePlanner::CompositePlanner(
+    statespace::ConstStateSpacePtr stateSpace,
+    const std::vector<PlannerPtr>& planners)
+  : Planner(std::move(stateSpace)), mPlanners(planners)
+{
+  // Do nothing
+}
+
+//==============================================================================
+bool CompositePlanner::hasPlanner(const Planner* planner) const
```
gilwoolee: Could it be more useful to do a type check instead of reference check?

jslee02: This class is general to handle the case of having different planner instances of the same type. In that case, type checking will return the first planner instance of the type. Or we could have another function that returns set of planners of the type.

As a side note, it'd be better to take type for type checking.

include/aikido/planner/ConfigurationToEndEffectorOffset.hpp
jslee02: Yes. If the underlying planners don't support negative distance, they could flip the direction and sign or throw an exception if flipping doesn't make sense to the planner.

tests/planner/test_SnapPlanner.cpp
```diff
@@ -51,43 +57,80 @@ class SnapPlannerTest : public ::testing::Test
   shared_ptr<PassingConstraint> passingConstraint;
   shared_ptr<FailingConstraint> failingConstraint;
   shared_ptr<GeodesicInterpolator> interpolator;
-  aikido::planner::PlanningResult planningResult;
+  SnapConfigurationToConfigurationPlanner::Result planningResult;
```
brianhou: What's the difference between `SnapConfigurationToConfigurationPlanner::Result` and `Planner::Result`? Maybe it'd be easier if we can always use `Planner::Result` so we don't have to change our result type if we decide we want to use a different planner? What would happen for a meta-planner?

jslee02: They are identical. Note that `SnapConfigurationToConfigurationPlanner` doesn't actually typedefs `Result`, but it just `Planner::Result`. Using `[X]Planner::Result` would be a safe way because it will always use the right result type.

Planners could have different types of information so it would be more flexible to define different result type when necessary.

Maybe we need to introduce `Planner::createResult()` to be able to create the right result from a base planner pointer like we did for `Outcome`.

include/aikido/planner/ConfigurationToConfigurations.hpp
```diff
@@ -0,0 +1,62 @@
+#ifndef AIKIDO_PLANNER_CONFIGURATIONTOCONFIGURATIONS_HPP_
+#define AIKIDO_PLANNER_CONFIGURATIONTOCONFIGURATIONS_HPP_
+
+#include <unordered_set>
+#include "aikido/constraint/Testable.hpp"
```
aditya-vk: Not: There needs to be an empty line between aikido header and standard headers.

brianhou: Why? `clang-format` won't order them correctly if we add empty lines.

include/aikido/planner/ConfigurationToEndEffectorOffset.hpp
```diff
@@ -0,0 +1,77 @@
...
+public:
+  /// Constructor.
+  ///
+  /// \param[in] stateSpace State space.
+  /// \param[in] endEffectorBodyNode BodyNode to be planned to move to a desired
+  /// offest while maintaining the current orientation.
+  /// \param[in] startState Start state.
+  /// \param[in] direction Unit vector that represents the direction of motion
+  /// [unit vector in the world frame]. The direction is allowed to be zero
+  /// vector.
```
aditya-vk: Should this docstring mention what direction being zero-vector means?

tests/planner/test_SnapPlanner.cpp
brianhou: Sounds good. Let's merge this PR first and add that change later.

include/aikido/planner/ConfigurationToConfigurations.hpp
aditya-vk: From a recent comment by @psigen , AIKIDO follows this convention:
https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes

include/aikido/planner/ConfigurationToEndEffectorOffset.hpp
brianhou: Maybe let's just throw an error if the direction is zero since we shouldn't ever actually want to do that.

aditya-vk: Yeah.
@jslee02 why is it allowed to be zero vector?

include/aikido/planner/Planner.hpp
```diff
@@ -0,0 +1,75 @@
+#ifndef AIKIDO_PLANNER_PLANNER_HPP_
+#define AIKIDO_PLANNER_PLANNER_HPP_
+
+#include <string>
+
```
aditya-vk: Uh oh :P

include/aikido/planner/PlanningResult.hpp
```diff
@@ -6,9 +6,22 @@
...
+  void setMessage(const std::string& message);
+
+  /// Returns message.
+  const std::string& getMessage() const;
+
+public: // TODO: Change this to protected
+  /// Message
+  std::string mMessage;
+
+  // TODO: Remove
```
aditya-vk: Requires some work?

jslee02: This class will be removed anyway. :)

src/planner/ConfigurationToEndEffectorOffset.cpp
```diff
@@ -19,7 +19,8 @@ ConfigurationToEndEffectorOffset::ConfigurationToEndEffectorOffset(
   , mDirection(direction.normalized())
   , mDistance(signedDistance)
 {
-  // Do nothing
+  if (direction.squaredNorm() == 0)
```
brianhou: Should this be `direction.isZero()` instead?

include/aikido/planner/SnapConfigurationToConfigurationPlanner.hpp
```diff
@@ -0,0 +1,65 @@
...
+  /// The planner returns success if the resulting trajectory satisfies
+  /// constraint at some resolution and failure (returning \c nullptr)
+  /// otherwise. The reason for the failure is stored in the \c result output
+  /// parameter.
+  ///
+  /// \param[in] problem Planning problem.
+  /// \param[out] result Information about success or failure.
+  /// \return Trajectory or \c nullptr if planning failed.
+  /// \throw If \c problem is not ConfigurationToConfiguration.
+  /// \throw If \c result is not ConfigurationToConfiguration::Result.
```
aditya-vk: If this splitting of `\throw` is the convention, can you please follow this even in `include/aikido/planner/Problem.hpp`. If there _is_ no convention, let us be consistent within the PR :)

src/planner/FirstSupportedMetaPlanner.cpp
```diff
@@ -0,0 +1,22 @@
...
+trajectory::TrajectoryPtr FirstSupportedMetaPlanner::plan(
+    const Problem& problem, Result* result)
+{
+  for (const auto& planner : mPlanners)
+  {
+    if (planner->canSolve(problem))
+      return planner->plan(problem, result);
+  }
+
+  // TODO: Set result "no supported planner".
```
aditya-vk: (bringing to your addition)

jslee02: I believe we will go over all the planner implementation including meta planners after this PR is merged.

src/planner/Problem.cpp
```diff
@@ -0,0 +1,45 @@
...
+  : mStateSpace(std::move(stateSpace)), mConstraint(std::move(constraint))
+{
+  if (!mStateSpace)
+    throw std::invalid_argument("StateSpace shouldn't be nullptr.");
+
+  if (!mConstraint)
+  {
+    mConstraint = std::make_shared<constraint::Satisfied>(
+        std::const_pointer_cast<statespace::StateSpace>(mStateSpace));
+    // TODO(JS): Remove this const-cast once we fix const-correctness of
```
aditya-vk: Can you reference this in #369, I'll take it up soon? Thanks!

src/planner/SnapConfigurationToConfigurationPlanner.cpp
```diff
@@ -0,0 +1,76 @@
...
+        statespace::ConstInterpolatorPtr interpolator)
+  : ConfigurationToConfigurationPlanner(std::move(stateSpace))
+  , mInterpolator(std::move(interpolator))
+{
+  if (!mInterpolator)
+  {
+    mInterpolator
+        = std::make_shared<statespace::GeodesicInterpolator>(mStateSpace);
+    // TODO(JS): Not sure if GeodesicInterpolator is a good choice as the
+    // default.
```
aditya-vk: Can you make an issue stating why? It can get lost here and seems important if there's valid reasons. Thanks!

jslee02: I put this comment without looking at other interpolators. Now I think it's a good choice. Let me delete this comments.

brianhou:
Thanks for the changes! Merging this now 🎉

PR #316, opened 2018-02-02
Enable faster-than-realtime simulation

brianhou:
This should be generally useful, but is especially important for MAGI so that we can implement `ExecutePositionCommandSolution::jump`.

This breaks (almost) all previous `Executor` APIs by adding a `timepoint` argument to `step()`.

I've also attempted to reorganize the code and documentation to be more consistent across executors. Some parts of the code are near-identical and should probably be refactored more in the future.

***


psigen:
There is something interesting about "timestep" needing to be a fundamental property here.  I think it's because of the assumption that something needs to be evaluated as a discrete time check as a fundamental part of any action...

You could imagine that any action whose results could be evaluated analytically e.g. a swept volume check or a kinematic trajectory without collisions (like a cart on a rail) would not actually need a timestep, but might need some other fundamental configuration option.

include/aikido/control/PositionCommandExecutor.hpp
```diff
@@ -1,29 +1,48 @@
...
   virtual std::future<void> execute(const Eigen::VectorXd& goalPositions) = 0;
 
   // Step once.
   virtual void step() = 0;
+
+  /// Get the current timestep.
+  virtual std::chrono::milliseconds getTimestep() const;
+
+  /// Set the current timestep.
+  virtual void setTimestep(std::chrono::milliseconds timestep);
```
gilwoolee: Does this mean that we can arbitrarily change the timestep during execution? Isn't it too confusing? Why not just set it as a const member variable defined at the constructor?

include/aikido/control/InstantaneousTrajectoryExecutor.hpp
```diff
@@ -17,20 +16,24 @@ class InstantaneousTrajectoryExecutor : public TrajectoryExecutor
...
 public:
   /// Constructor.
+  ///
   /// \param skeleton Skeleton to execute trajectories on.
   ///        All trajectories must have dofs only in this skeleton.
+  /// \param timestep The time period that each call to step() should simulate
   explicit InstantaneousTrajectoryExecutor(
-      ::dart::dynamics::SkeletonPtr skeleton);
+      ::dart::dynamics::SkeletonPtr skeleton,
+      std::chrono::milliseconds timestep);
```
gilwoolee: Doesn't `InstantenousTrajectoryExecutor` steps immediately to the end? Why do we need to set the timestep?

gilwoolee:
@psigen I think the main reason why we need `timestep` here is that we need it to simulate execution. I agree, for examples you mentioned, we shouldn't really need it. I don't think we use these executors in such applications though.

include/aikido/control/InstantaneousTrajectoryExecutor.hpp
brianhou: You're right, I don't think we need to.

mkoval:
Whoops. I missed @psigen's comment - he raises a great point.

All of these classes have `timestep` parameters because they are intended to act as a simulation that can stand-in for execution on the real robot. E.g. when sending a "close hand" command, we wanted to be able to visualize the hand closing just like it does on the real robot.

You raise a great point that this is not strictly necessary in MAGI because we only care about the end state. We could potentially implement a more efficient versions of these simulations there, e.g. by using swept volume collision detection or a bisection method. I haven't seen a strong need to implement those yet, since: (a) it would take a lot of work, (b) could end up being inconsistent with the higher-fidelity simulation, and (c) may not be significantly faster than running the full simulation in a faster-than-realtime mode.

@psigen Does that make sense?

brianhou:
Thanks for the comments! I definitely agree with your first point -- that was my main concern with this implementation. I didn't think to add an argument to `step`, though.

I think having `step` take a `timepoint` to simulate until sounds reasonable, but the name seems potentially misleading. What about `stepUntil` or something similar?

gilwoolee:
Looks mostly good! I left some Nit comments. :)

include/aikido/control/BarrettFingerKinematicSimulationPositionCommandExecutor.hpp
```diff
@@ -52,88 +52,102 @@ class BarrettFingerKinematicSimulationPositionCommandExecutor
...
-  /// \return false if fails to change collideWith (during execution).
+  /// \copydoc BarrettHandKinematicSimulationPositionCommandExecutor::step()
+  ///
+  /// Moves the finger joint positions by dofVelocity * timeSincePreviousCall
+  /// until either the proximal dof reaches goalPosition, a joint limit is
+  /// reached, or collision is detected.
+  /// When collision is detected on the distal link, the finger stops.
+  /// When collision is detected on the proximal link, the distal link continues
+  /// to move until it either reaches mimicRatio * goalPosition, a joint limit
+  /// is reached, or collision is detected.
```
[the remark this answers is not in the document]
brianhou: This should already be copied from `BarrettHandKinematicSimulationPositionCommandExecutor::step`.

include/aikido/control/ros/RosPositionCommandExecutor.hpp
```diff
@@ -48,30 +48,30 @@ class RosPositionCommandExecutor
   ///
   /// To be executed on a separate thread.
   /// Regularly checks for the completion of a sent trajectory.
-  void step() override;
+  void step(
+      const std::chrono::system_clock::time_point& /*timepoint*/) override;
```
jslee02: Nit: We don't need to comment out for argument name in a header file; actually, we might want to have here so that Doxygen uses it. The unused variable warning happens only in the source file.

include/aikido/control/ros/RosTrajectoryExecutor.hpp
```diff
@@ -56,7 +55,8 @@ class RosTrajectoryExecutor : public aikido::control::TrajectoryExecutor
   ///
   /// To be executed on a separate thread.
   /// Regularly checks for the completion of a sent trajectory.
-  void step() override;
+  void step(
+      const std::chrono::system_clock::time_point& /*timepoint*/) override;
```
jslee02: Nit: We don't need to comment out for argument name in a header file; actually, we might want to have here so that Doxygen uses it. The unused variable warning happens only in the source file.

src/control/BarrettHandKinematicSimulationPositionCommandExecutor.cpp
```diff
@@ -121,42 +133,45 @@ std::future<void>
...
             << goalPositions.size() << "] given.";
     throw std::invalid_argument(message.str());
   }
 
   mPromise.reset(new std::promise<void>());
-  mProximalGoalPositions = goalPositions.head<3>();
-  mSpreadGoalPosition = goalPositions.row(3);
-  mInExecution = true;
+
+  Eigen::VectorXd proximalGoalPositions = goalPositions.head<3>();
```
jslee02: Nit: It would be good to use fixed-size vector when we know the size in compile time. So `Eigen::Vector3d proximalGoalPositions = goalPositions.head<3>();`

src/control/BarrettHandKinematicSimulationPositionCommandExecutor.cpp
```diff
@@ -121,42 +133,45 @@ std::future<void>
...
+  Eigen::VectorXd spreadGoalPosition = goalPositions.row(3);
```
jslee02: Nit: There is a fixed-size version (for vector) of this: `.tail<3>()`. Also `Eigen::Vector3d`.

src/control/BarrettHandKinematicSimulationPositionCommandExecutor.cpp
```diff
@@ -121,42 +133,45 @@ std::future<void>
...
   mFingerFutures.clear();
 
-  mFingerFutures.reserve(kNumPositionExecutor + kNumSpreadExecutor);
-  for (std::size_t i = 0; i < kNumPositionExecutor; ++i)
+  mFingerFutures.reserve(kNumPositionExecutors + kNumSpreadExecutor);
+  for (std::size_t i = 0; i < kNumPositionExecutors; ++i)
     mFingerFutures.emplace_back(
-        mPositionCommandExecutors[i]->execute(mProximalGoalPositions.row(i)));
+        mPositionCommandExecutors[i]->execute(proximalGoalPositions.row(i)));
```
jslee02: Nit: Since `proximalGoalPositions` is a vector, we might want to use `[i]` instead of `row(i)`.

brianhou: The `PositionCommandExecutor::execute` interface requires this to be an `Eigen::VectorXd`, not a double.

jslee02: If so, this line makes sense. :smile:

PR #317, opened 2018-02-02
Macro for smart pointer declarations

jslee02:
This is a revision PR of #308 based on the discussion in it (Solution 4).

**Summary**
* Define a macro for declaring smart pointers in `aikido/common/pointers.hpp`
* Use the macro for the forward declaration of pointers before the class definition (so that the class can use the pointer types as well)

***


include/aikido/common/pointers.hpp
```diff
@@ -0,0 +1,13 @@
+#ifndef AIKIDO_COMMON_POINTERS_HPP_
+#define AIKIDO_COMMON_POINTERS_HPP_
+
+#include <dart/common/SmartPointer.hpp>
+
+// clang-format off
+#define AIKIDO_DECLARE_POINTERS(X)                                             \
+  DART_COMMON_MAKE_SHARED_WEAK(X)                                              \
```
mkoval: Do we need to forward-declare `X` here, e.g. `class X`? Does `DART_COMMON_MAKE_SHARED_WEAK` do that internally?

I also suggest adding a comment that this macro _must be used in the class' namespace_. You'd end up with some pretty gross C++ code if you tried to use a fully qualified name, e.g. `Uniquedart::dynamics::BodyNodePtr`. 😉

jslee02:
> what is the concern about circular dependencies? Is there a concrete example where this is an issue in Aikido today?

I was concerned that it could happen, but haven't found any existing circular dependency problem from the current AIKIDO codebase. The other side effect of aggressively using forward declaration instead of including the full definition that I expect is reducing the number of total header inclusions; so the compilation time as well. But I'm open to not doing that.

include/aikido/common/pointers.hpp
jslee02: > Does DART_COMMON_MAKE_SHARED_WEAK do that internally?

Yes. https://github.com/dartsim/dart/blob/master/dart/common/SmartPointer.hpp#L42

> I also suggest adding a comment that this macro must be used in the class' namespace.

:+1:

include/aikido/rviz/pointers.hpp
```diff
@@ -0,0 +1,18 @@
...
+AIKIDO_DECLARE_POINTERS(FrameMarker)
+AIKIDO_DECLARE_POINTERS(BodyNodeMarker)
+AIKIDO_DECLARE_POINTERS(SkeletonMarker)
+AIKIDO_DECLARE_POINTERS(TrajectoryMarker)
+AIKIDO_DECLARE_POINTERS(TSRMarker)
+
+} // namespace rviz
+} // namespace aikido
+
+#endif // ifndef AIKIDO_RVIZ_SMARTPOINTERS_HPP_
```
psigen: Might need to update this comment.

jslee02:
@psigen @mkoval Thanks for the review! 

@mkoval Does this look good to go?

jslee02:
@mkoval Merging this PR as I believe you're happy with this PR now. :stuck_out_tongue:

PR #319, opened 2018-02-03
Clean up Doxygen output

brianhou:
Our Doxygen output has a lot of absolute paths `/home/bhou1/.../aikido`, which (a) clutter up the page / HTML title and (b) expose unnecessary information. This PR attempts to remove some of it.

When no .tag files are used, this PR successfully removes all the unnecessary absolute paths. For some reason, using .tag files results in a ton of HTML tags like
```
<a class="elRef" doxygen="/home/bhou1/Development/aikido-ws/src/aikido/build/doxygen_working/dart.tag:http://dartsim.github.io/dart/v6.3.0/" href="http://dartsim.github.io/dart/v6.3.0/de/df0/CollisionOption_8hpp.html">dart/collision/CollisionOption.hpp</a>
```

***


brianhou:
I still want to figure out how to fix the issue with .tag files before merging.

brianhou:
@jslee02 I think this is good to go! Can you try it out locally as well, just to double check?

(TIL CMake does something special with semicolons and that you should use `$<SEMICOLON>` for semicolon literals instead.)

PR #323, opened 2018-02-04
Add missing explicit specifiers

jslee02:
This PR fixes two categories of issues reported by Codacy:
1. Missing `explicit` for single parameter class functions
1. ~~[Use of an invalid iterator](https://app.codacy.com/app/jslee02/aikido/file/14402016515/issues/source?bid=6457804&fileBranchId=6457804#l138)~~

~~I believe the second issue is misleading but fixed it anyway because changing a protected/private member variable from outside of the class isn't preferable practice. I'm open to hearing better way to solve the issue.~~

***

### DIFF
--- a/include/aikido/common/VanDerCorput.hpp
+++ b/include/aikido/common/VanDerCorput.hpp
@@ -88,7 +88,10 @@ private:
 
   /// Private constructor that should always be constructed from
   /// VanDerCorput::begin()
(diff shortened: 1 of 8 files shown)


jslee02:
I reverted changes to `VanDerCorput` because the fix introduces other problems. Let me fix this in a separate PR.

PR #324, opened 2018-02-06
Update ParabolicSmoother/ParabolicTimer to support Splines

sniyaz:
Magi requires that we have post processors support both Interpolated and Spline trajectories. This is a small PR that makes the parabolic post processor classes support the new methods added in #302.


include/aikido/planner/TrajectoryPostProcessor.hpp
```diff
@@ -17,6 +17,13 @@ class TrajectoryPostProcessor
       const aikido::trajectory::InterpolatedPtr& _inputTraj,
       const aikido::common::RNG* _rng)
       = 0;
+
+  /// \param _inputTraj The untimed *spline* trajectory for the arm to process.
+  /// \param _rng Random number generator.
+  virtual std::unique_ptr<aikido::trajectory::Spline> postprocess(
+      const std::unique_ptr<trajectory::Spline>& _inputTraj,
```
jslee02: I don't believe `postprocess` wants to take the ownership of trajectories. Could we change this to just const reference type as well as the above `Interpolated` version? If this introduces too much work and out of this PR's scope, we could do this as a separate PR.

src/planner/parabolic/ParabolicSmoother.cpp
```diff
@@ -156,10 +156,49 @@ std::unique_ptr<aikido::trajectory::Spline> ParabolicSmoother::postprocess(
...
+
+  return timedTrajectory;
+}
+
+//==============================================================================
+std::unique_ptr<aikido::trajectory::Spline> ParabolicSmoother::postprocess(
+    const std::unique_ptr<trajectory::Spline>& _inputTraj,
+    const aikido::common::RNG* _rng)
+{
+  if (!_rng)
```
jslee02: Could we change the type of `_rng` to const reference if we throw an exception for `nullptr` anyway?

include/aikido/planner/TrajectoryPostProcessor.hpp
sniyaz: Sure! These classes are pretty new, so changing the API shouldn’t break things too much.

sniyaz:
@jslee02, responded to your code review comments! 😄

jslee02:
@sniyaz Thanks for the change! 

It would be great to have tests for this new functions.

sniyaz:
@jslee02 I've added new tests for ParabolicTimer that also test Spline trajectories.

PR #334, opened 2018-02-13
[WIP] Add style guide for template class

jslee02:
The major motivation of this PR is to make using `\tparam` for template argument as an explicit convention of AIKIDO, but it would be also worth to have guide for template classes.

### DIFF
--- a/STYLE.md
+++ b/STYLE.md
@@ -102,7 +102,7 @@ class TemplatedExampleClass
 {
 public:
   /// Brief description.
-  void exampleMethod();
+  T exampleMethod();
 };
 
 } // namespace example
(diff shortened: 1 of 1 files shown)


STYLE.md
```diff
@@ -95,6 +95,16 @@ private:
...
 };
 
+/// This is a template class description.
+/// \tparam T This is an example template parameter description.
+template <typename T>
+class TemplatedExampleClass
+{
+public:
+  /// Brief description.
+  void exampleMethod();
```
brianhou: Maybe have this return type be `T`?

STYLE.md
```diff
@@ -107,6 +117,28 @@ private:
...
+#define AIKIDO_EXAMPLE_DETAIL_EXAMPLECLASS_IMPL_HPP_
+
+#include "aikido/example/ExampleClass.hpp"
+
+namespace aikido {
+namespace example {
+
+//==============================================================================
+template <typename T>
+TemplatedExampleClass<T>::exampleMethod()
```
brianhou: Missing return type?

PR #336, opened 2018-02-16
clean up and integrate pose estimators

yskim041:
In this PR, I integrated pose estimator modules for AprilTags and RcnnPose into one PoseEstimator module. I updated ObjectDatabase and corresponding YAML files in pr-ordata (https://github.com/personalrobotics/pr-ordata/pull/32) as well.

***


include/aikido/perception/ObjectDatabase.hpp
```diff
@@ -44,7 +44,8 @@ class ObjectDatabase
   void getObjectByKey(
```
brianhou: Can we document this method? I'm not entirely clear what `obj_offset` is.

jslee02:
This PR looks like removing `AprilTags` and adding offset to `RcnnPose` (and changing the name). Having offset was the only difference of `AprilTags` from `RcnnPose`? :thinking:

brianhou:
Let's try to update this PR and merge this soon!

include/aikido/perception/ObjectDatabase.hpp
yskim041: I added the documentation for this. `objectName`, `objectResource`, and `objectOffset` are what should be retrieved from ObjectDatabase. For each key, those values are stored in `tag_data_xx.json` in `pr_ordata`.

yskim041: `objectOffset` is the offset between AprilTags and actual origin of objects.

yskim041:
Tested with the can detector by using personalrobotics/herb3_demos#13, and this works well. Since personalrobotics/pr_ordata#32 is merged, this branch works with the master branch of `pr_ordata`.

include/aikido/perception/ObjectDatabase.hpp
```diff
@@ -41,10 +47,17 @@ class ObjectDatabase
 
   virtual ~ObjectDatabase() = default;
 
+  /// Get the object name, resource, and offset from database by objectKey
+  /// \param[in]  objectKey: The key(string) of an object in ObjectDatabase
```
brianhou: I believe we don't want colons in Doxygen comments. Also, minor typo: `key(string)` -> `key (string)`.

include/aikido/perception/ObjectDatabase.hpp
```diff
@@ -41,10 +47,17 @@ class ObjectDatabase
...
+  /// \param[out] objectName: The retrieved object name from ObjectDatabase
+  /// \param[out] objectResource: The retrieved uri of the object
+  /// \param[out] objectOffset: The retrieved offset matrix of the object
+  ///     e.g. the offset between a tag and the actual origin of an object
```
brianhou: I think this should be `i.e.` rather than `e.g.`.  Also, I think we should be calling this a transform rather than an offset, if we're using `Isometry3d`.

include/aikido/perception/PoseEstimatorModule.hpp
```diff
@@ -12,17 +12,18 @@
...
-/// by subscribing to a \c visualization_msg::MarkerArray ROS topic published by
-/// the RcnnPose node. It uses a \c ObjectDatabase to resolve each marker to a
+/// It receives input from the PoseEstimator running in a separate process,
+/// by subscribing to a \c visualization_msg::MarkerArray ROS topic
+/// published by the PoseEstimator node.
+/// It uses a \c ObjectDatabase to resolve each marker to a
 /// \c dart::common::Uri, and updates the environment which is an
 /// \c aikido::planner::World.
-class RcnnPoseModule : public PerceptionModule
+class PoseEstimatorModule : public PerceptionModule
```
brianhou: What do you think about renaming this to `RosPoseEstimationModule`?

brianhou:
Looks good! I had a few minor nits, but I think this should be otherwise good to go.

src/perception/PoseEstimatorModule.cpp
```diff
@@ -1,4 +1,4 @@
-#include <aikido/perception/RcnnPoseModule.hpp>
+#include <aikido/perception/PoseEstimatorModule.hpp>
```
jslee02: Nit: Our convention is to use `"..."` for AIKIDO headers instead of `<...>`.

src/perception/PoseEstimatorModule.cpp
```diff
@@ -45,14 +47,14 @@ bool RcnnPoseModule::detectObjects(
   // Making sure the Message from RcnnPose is non empty
   if (marker_message == nullptr)
   {
-    dtwarn << "[RcnnPoseModule::detectObjects] nullptr Marker Message "
+    dtwarn << "[PoseEstimatorModule::detectObjects] nullptr Marker Message "
            << mMarkerTopic << std::endl;
     return false;
   }
 
   if (marker_message->markers.size() == 0)
```
jslee02: Nit: Using `empty()` is more preferred way to check its emptiness rather than `size()`.

src/perception/PoseEstimatorModule.cpp
yskim041: changed.

yskim041:
I addressed all the issues above.

include/aikido/perception/PoseEstimatorModule.hpp
yskim041: If we want to rename it with a more ros related name, I think `RosMarkerArrayClient` would be the most clear option. If not, I personally prefer `PoseEstimatorModule`.

include/aikido/perception/ObjectDatabase.hpp
brianhou: Should this be an offset or a transform? Since it's an `Isometry3d` it seems like it should be the latter?

brianhou: Some of them in the apriltag tag_data (https://github.com/personalrobotics/pr_ordata/blob/master/data/objects/tag_data_apriltags.json) have rotation components as well.

Issue #281, opened 2017-12-08
Consider changing Trajectory to take const StateSpace


jslee02:
I believe there are no cases that Trajectory changes the StateSpace.

Issue #282, opened 2017-12-08
Fix const correctness in multiple places


jslee02:
At a glance, there are several places violating const correctness like a const function returning a non-const pointer to the member variable.

For example:
* https://github.com/personalrobotics/aikido/blob/master/include/aikido/statespace/StateSpace.hpp#L107
* https://github.com/personalrobotics/aikido/blob/master/include/aikido/trajectory/Trajectory.hpp#L24

* https://github.com/personalrobotics/aikido/blob/master/include/aikido/control/KinematicSimulationTrajectoryExecutor.hpp#L56

* Many functions take `const` pointers and not pointers to `const` objects. For example `const trajectory::TrajectoryPtr` when the intention is `trejctory::ConsTrajectoryPtr`.

However, this issue cannot be fixed now because it will cause other problems that should be fixed by upcoming changes like #278. So this should be fixed once those changes are made.

PR #337, opened 2018-02-18
const stateSpace (and interpolator) for Trajectories

aditya-vk:
This PR addresses Issue #281 

***

Most of the following do not apply.

### DIFF
--- a/src/control/KinematicSimulationTrajectoryExecutor.cpp
+++ b/src/control/KinematicSimulationTrajectoryExecutor.cpp
@@ -47,7 +47,7 @@ void KinematicSimulationTrajectoryExecutor::validate(
   if (traj->metadata.executorValidated)
     return;
 
-  const auto space = std::dynamic_pointer_cast<MetaSkeletonStateSpace>(
+  const auto space = std::dynamic_pointer_cast<const MetaSkeletonStateSpace>(
       traj->getStateSpace());
 
   if (!space)
@@ -65,6 +65,8 @@ void KinematicSimulationTrajectoryExecutor::validate(
 std::future<void> KinematicSimulationTrajectoryExecutor::execute(
     trajectory::TrajectoryPtr traj)
 {
+  using aikido::statespace::dart::MetaSkeletonStateSpacePtr;
+
   validate(traj);
 
   {
@@ -77,8 +79,13 @@ std::future<void> KinematicSimulationTrajectoryExecutor::execute(
     mPromise.reset(new std::promise<void>());
 
     mTraj = traj;
-    mStateSpace = std::dynamic_pointer_cast<MetaSkeletonStateSpace>(
+
+    auto nonConstStateSpace = 
+      std::const_pointer_cast<aikido::statespace::StateSpace>(
         traj->getStateSpace());
+    mStateSpace = std::dynamic_pointer_cast<MetaSkeletonStateSpace>(
+        nonConstStateSpace);
+
     mInProgress = true;
     mExecutionStartTime = std::chrono::system_clock::now();
   }
--- a/src/control/KinematicSimulationTrajectoryExecutor.cpp
+++ b/src/control/KinematicSimulationTrajectoryExecutor.cpp
@@ -78,13 +78,11 @@ std::future<void> KinematicSimulationTrajectoryExecutor::execute(
 
     mPromise.reset(new std::promise<void>());
 
-    mTraj = traj;
+    mTraj = std::move(traj);
 
-    auto nonConstStateSpace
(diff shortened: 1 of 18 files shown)


src/control/KinematicSimulationTrajectoryExecutor.cpp
```diff
@@ -77,8 +79,13 @@ std::future<void> KinematicSimulationTrajectoryExecutor::execute(
...
 
     mTraj = traj;
-    mStateSpace = std::dynamic_pointer_cast<MetaSkeletonStateSpace>(
-        traj->getStateSpace());
+
+    auto nonConstStateSpace
+        = std::const_pointer_cast<aikido::statespace::StateSpace>(
+            traj->getStateSpace());
+    mStateSpace
+        = std::dynamic_pointer_cast<MetaSkeletonStateSpace>(nonConstStateSpace);
```
jslee02: I believe we could make `mStateSpace` const so that `const_pointer_cast` is not required.

aditya-vk: I wanted to get rid of `const_pointer_cast` as well but we initialize the executor dependent only on the skeleton (mStateSpace{nullptr}) and tie it to a statespace` only after the trajectory is received in `execute`. 
1. We could pass the statespace when the executor is instantiated.
2. We could get rid of `mStateSpace` entirely but this would mean extracting it out of the trajectory everytime `step` is called.

jslee02: I think we could achieve this by changing `traj` and `mTraj` to const. I believe a trajectory executor doesn't need to change the trajectory so it would make sense to take trajectory as const.

src/control/KinematicSimulationTrajectoryExecutor.cpp
```diff
@@ -77,8 +79,13 @@ std::future<void> KinematicSimulationTrajectoryExecutor::execute(
     mPromise.reset(new std::promise<void>());
 
     mTraj = traj;
```
jslee02: Nit: `mTraj = std::move(traj);`

src/control/KinematicSimulationTrajectoryExecutor.cpp
aditya-vk: I thought an executor for a skeleton can `execute()` on multiple trajectories.

jslee02: It would be fine. You can take a new _const_ trajectory and set it to `mTraj`, which is also a const trajectory. Please note that the constness applies to the object but not the pointer itself so we can change the pointer to another pointer (but can't change the object the pointer points to; a trajectory executor executes multiple trajectories but doesn't change the parameters of the trajectories). Let me know if I'm missing something here.

aditya-vk: OOOOH! I got confused again! Dammit. :disappointed:
Yeah that makes sense.

aditya-vk: Done. :)
Didn't have to change const-ness of the trajectories but `mTraj` and `traj` should actually be pointers to const objects as well. I will address that while I work on #282

src/planner/parabolic/ParabolicTimer.cpp
```diff
@@ -41,7 +41,7 @@ std::unique_ptr<aikido::trajectory::Spline> convertToSpline(
         "computeParabolicTiming only supports Rn, "
         "SO2, and CartesianProducts consisting of those types.");
 
-  const auto interpolator = _inputTrajectory.getInterpolator();
+  auto interpolator = _inputTrajectory.getInterpolator();
```
brianhou: I think the following type check is redundant because of the one above.

src/planner/parabolic/ParabolicTimer.cpp
```diff
@@ -24,7 +24,7 @@ std::unique_ptr<aikido::trajectory::Spline> convertToSpline(
 {
   using aikido::statespace::GeodesicInterpolator;
 
-  if (nullptr == dynamic_cast<GeodesicInterpolator*>(
+  if (nullptr == dynamic_cast<const GeodesicInterpolator*>(
```
brianhou: I don't think an explicit comparison to `nullptr` is necessary. Instead, let's do something like
`if (!dynamic_pointer_cast<const GeodesicInterpolator>(interpolator))`?

src/planner/parabolic/ParabolicTimer.cpp
aditya-vk: Good catch :+1:

PR #339, opened 2018-02-18
Add flags to Saver classes to specify what to save

### LINKED ISSUE #276: Selective state save/restore for World
During planning we want to save the current state of the world and restore after planning is complete. It is not necessary to save the whole state, and is generally more efficient to save only the components that may be modified during planning, so we may want to support selective state saving/restoring.

### LINKED ISSUE #293: Consider adding flags to MetaSkeletonStateSpaceSaver to specify what to save
https://github.com/personalrobotics/aikido/pull/288#discussion_r160215047

This would be beneficial to avoid unnecessarily setting/getting the skeleton states/properties.

brianhou:
Resolves #276. Resolves #293.

This implementation might not be ideal because the "options" can be arbitrary ints, rather than combinations of the options that we actually enumerate. If that's a problem, maybe something like [this](http://blog.bitwigglers.org/using-enum-classes-as-type-safe-bitmasks/) would be better?

***

### DIFF
--- a/include/aikido/planner/WorldStateSaver.hpp
+++ b/include/aikido/planner/WorldStateSaver.hpp
@@ -21,7 +21,7 @@ public:
   ///
   /// \param[in] world World to save state from and restore to.
   /// \param[in] options Options to specify what should be saved
-  explicit WorldStateSaver(const World* world, int options = CONFIGURATIONS);
+  explicit WorldStateSaver(World* world, int options = CONFIGURATIONS);
 
   virtual ~WorldStateSaver();
 
--- a/include/aikido/planner/WorldStateSaver.hpp
+++ b/include/aikido/planner/WorldStateSaver.hpp
@@ -6,16 +6,16 @@
 namespace aikido {
 namespace planner {
 
-/// Options to specify what WorldStateSaver should save.
-enum WorldStateSaverOptions
-{
-  CONFIGURATIONS = 1 << 0,
-};
-
 /// RAII class to save and restore a World's state.
 class WorldStateSaver
 {
 public:
+  /// Options to specify what WorldStateSaver should save.
+  enum Options
+  {
+    CONFIGURATIONS = 1 << 0,
+  };
+
   /// Construct a WorldStateSaver and save the current state of the \c World.
   /// This state will be restored when WorldStateSaver is destructed.
   ///
--- a/include/aikido/planner/WorldStateSaver.hpp
+++ b/include/aikido/planner/WorldStateSaver.hpp
@@ -1,25 +1,38 @@
 #ifndef AIKIDO_PLANNER_WORLDSTATESAVER_HPP_
 #define AIKIDO_PLANNER_WORLDSTATESAVER_HPP_
-#include "World.hpp"
+
+#include "aikido/planner/World.hpp"
 
 namespace aikido {
 namespace planner {
+
+/// Options to specify what WorldStateSaver should save.
+enum WorldStateSaverOptions
+{
+  CONFIGURATIONS = 1 << 0,
+};
 
 /// RAII class to save and restore a World's state.
 class WorldStateSaver
 {
 public:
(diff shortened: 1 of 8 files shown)


include/aikido/planner/WorldStateSaver.hpp
```diff
@@ -9,17 +10,29 @@ namespace planner {
...
+    CONFIGURATIONS = 1 << 0,
+  };
+
+  /// Construct a WorldStateSaver and save the current state of the \c World.
+  /// This state will be restored when WorldStateSaver is destructed.
   ///
-  /// \param _world World to save state from and restore to.
-  explicit WorldStateSaver(World* const _world);
+  /// \param world World to save state from and restore to.
+  /// \param options Options to specify what should be saved
```
jslee02: Nit: Let's use `\param[in]`!

include/aikido/planner/WorldStateSaver.hpp
```diff
@@ -9,17 +10,29 @@ namespace planner {
...
+  explicit WorldStateSaver(World* const world, int options = CONFIGURATIONS);
```
jslee02: Nit: Please change to `const World* world`. They are identical but different style. I believe our convention is `const World* world`.

include/aikido/statespace/dart/MetaSkeletonStateSaver.hpp
```diff
@@ -12,13 +12,22 @@ namespace dart {
...
+    POSITION_LIMITS = 1 << 1,
+  };
+
   /// Construct a MetaSkeletonStateSaver and save the current state of the \c
   /// MetaSkeleton. This state will be restored when MetaSkeletonStateSaver is
   /// destructed.
   ///
-  /// \param _metaskeleton MetaSkeleton to save/restore
+  /// \param metaskeleton MetaSkeleton to save/restore
+  /// \param options Options to specify what should be saved
```
jslee02: Nit: `\param[in]`

tests/statespace/dart/test_MetaSkeletonStateSaver.cpp
```diff
@@ -1,37 +1,122 @@
...
 using aikido::statespace::dart::MetaSkeletonStateSaver;
-using aikido::statespace::SO2;
 
-TEST(MetaSkeletonStateSaver, MetaSkeletonStateSpaceReturnsToOriginal)
+class MetaSkeletonStateSaverTest : public testing::Test
 {
-  auto skeleton = Skeleton::create();
-  skeleton->createJointAndBodyNodePair<RevoluteJoint>();
+public:
+  void SetUp()
```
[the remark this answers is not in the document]
brianhou: What do you mean? Isn't the correct name [`SetUp`](https://github.com/google/googletest/blob/master/googletest/docs/Primer.md#test-fixtures-using-the-same-data-configuration-for-multiple-tests)?

include/aikido/planner/WorldStateSaver.hpp
brianhou: I think it was actually incorrect to be `const` since we restore the `World` state in the destructor.

tests/statespace/dart/test_MetaSkeletonStateSaver.cpp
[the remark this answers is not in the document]
jslee02: I believe function naming convention is to start with a lower case (note that `SetUp` begins with an upper latter "S").

include/aikido/planner/WorldStateSaver.hpp
jslee02: @brianhou :+1:

tests/statespace/dart/test_MetaSkeletonStateSaver.cpp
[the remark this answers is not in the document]
jslee02: Oh, this is required by gtest. Sorry for the noise. :disappointed:

PR #340, opened 2018-02-19
Use strongly-typed enum flags.

brianhou:
> #339 might not be ideal because the "options" can be arbitrary ints, rather than combinations of the options that we actually enumerate. If that's a problem, maybe something like [this](http://blog.bitwigglers.org/using-enum-classes-as-type-safe-bitmasks/) would be better?

This PR follows the blog post above. I think the biggest annoyance with this implementation is that there's no contextual conversion to `bool`, so you can't write something like `if (mOptions & Options::POSITIONS)`. My intermediate workaround is to require `Options::NONE = 0` and compare directly to that.

I'm not entirely convinced whether this complication is worth it. @jslee02, what do you think?

***


include/aikido/common/EnumFlags.hpp
```diff
@@ -0,0 +1,44 @@
+#ifndef AIKIDO_COMMON_ENUMFLAGS_HPP_
+#define AIKIDO_COMMON_ENUMFLAGS_HPP_
+
+#include <type_traits>
+
+/// Enable bitwise operators for strongly-typed enums.
+///
+/// Adapted from
+/// http://blog.bitwigglers.org/using-enum-classes-as-type-safe-bitmasks/
```
jslee02: Nit: Maybe we want to make this as non-doxygen style comment because this comment is not specifically tied to a code but just a note?

psigen:
Just curious, can you implement something like the following to make that explicit boolean cast work?
https://stackoverflow.com/a/9883421

egordon:
Outdated, to consider for later

PR #349, opened 2018-02-25
const correctness

aditya-vk:
This PR tries to ensure const correctness throughout AIKIDO.

The PR also enables the executors to keep track of trajectories that have been validated (as against trajectories being marked to be validated - see discussion on this in relevant files).

***


src/control/InstantaneousTrajectoryExecutor.cpp
```diff
@@ -23,7 +23,8 @@ InstantaneousTrajectoryExecutor::~InstantaneousTrajectoryExecutor()
 }
 
 //==============================================================================
-void InstantaneousTrajectoryExecutor::validate(trajectory::TrajectoryPtr traj)
+void InstantaneousTrajectoryExecutor::validate(
+    const trajectory::ConstTrajectoryPtr& traj)
```
jslee02: If we throw an exception for nullptr `traj`, there is no point to take `traj` as a pointer. Also, this function doesn't take the ownership so taking it as `shared_ptr` is unnecessary. Taking this as const reference would make more sense because (1) it can enforce to pass not a `nullptr` here, and (2) it allows to pass a raw pointer (a trajectory doesn't need to be shared_ptr to call this function).

One possibility we still want this as `shared_ptr` would be the case that we foresee a derived class of `TrajectoryExecutor` would want to take the ownership to validate a trajectory, which I don't think so. Let me know if I'm missing something here.

If we change `traj` to const reference, please remove `nullptr` check.

src/control/KinematicSimulationTrajectoryExecutor.cpp
```diff
@@ -63,7 +63,7 @@ void KinematicSimulationTrajectoryExecutor::validate(
 
 //==============================================================================
 std::future<void> KinematicSimulationTrajectoryExecutor::execute(
-    trajectory::TrajectoryPtr traj)
+    trajectory::ConstTrajectoryPtr traj)
```
jslee02: Since we don't allow `nullptr`, let's throw an exception for `nullptr`.

Please add this to other `[~]::execute()` functions.

include/aikido/control/TrajectoryExecutor.hpp
```diff
@@ -20,7 +20,7 @@ class TrajectoryExecutor
   /// Validate the traj in preparation for execution.
   ///
   /// \param traj Trajectory to be validated
-  virtual void validate(trajectory::TrajectoryPtr traj) = 0;
+  virtual void validate(const trajectory::ConstTrajectoryPtr& traj) = 0;
```
jslee02: Does `validate` change the internal states of a trajectory executor? If not, let's make this a const function.

include/aikido/trajectory/Trajectory.hpp
```diff
@@ -72,7 +72,7 @@ class Trajectory
       double _t, int _derivative, Eigen::VectorXd& _tangentVector) const = 0;
 
   /// Trajectory metadata
-  TrajectoryMetadata metadata;
+  mutable TrajectoryMetadata metadata;
```
jslee02: At a glance, I think this isn't necessary to be mutable, but let me learn more about `TrajectoryMetadata` not to make stupid suggestions.

I learned that `TrajectoryMetadata` was introduced by #529. What's the purpose of this class? For now, it only includes `executorValidated` and it's set by `TrajectoryExecutor::validate()` (probably this is not to re-validate once it's validated?). However, isn't this vulnerable to store the result in a trajectory to the case that we validate a trajectory using an executor when the trajectory is already validated by a different executor? We can workaround this by setting the flag to false before whenever we validate a trajectory, but it also has an issue that executors will revalidate the same trajectory every time.

Also, I thought the validation logic is different per executers, but it isn't as of now. I just assume some derived trajectory executor class would require different logic. If not, it would be better to have the validity check code in a single function.

src/control/KinematicSimulationTrajectoryExecutor.cpp
aditya-vk: `execute` calls validate(), so the current implementation has execute take `const trajectory::ConstTrajectoryPtr` and pass the stored pointer to `validate` which checks for nullity. I think this considers both of your suggestions made on `validate` and `execute`.

include/aikido/common/ExecutorMultiplexer.hpp
```diff
@@ -47,7 +47,7 @@ class ExecutorMultiplexer final
 private:
   /// Mutex for the list of callbacks. The array of callbacks will be locked
   /// during it's modified and the callbacks are called.
-  std::mutex mMutex;
+  mutable std::mutex mMutex;
```
aditya-vk: This is part of a proposal for style convention, to have class member mutex to always be explicitly `mutable` irrespective of whether the class has `const` member functions. 

Open to removing the keyword if the style convention is not favorable.

include/aikido/control/TrajectoryExecutor.hpp
```diff
@@ -29,7 +29,8 @@ class TrajectoryExecutor
   /// \param _traj Trajectory to be executed.
   /// \return future<void> for trajectory execution. If trajectory terminates
   ///        before completion, future will be set to a runtime_error.
-  virtual std::future<void> execute(trajectory::TrajectoryPtr _traj) = 0;
+  virtual std::future<void> execute(const trajectory::ConstTrajectoryPtr& traj)
```
jslee02: Nit: Could we change this to const raw pointer so that shared pointer is not required to call this function? In some concrete implementation requires the ownership?

include/aikido/control/TrajectoryExecutor.hpp
```diff
@@ -43,6 +44,8 @@ class TrajectoryExecutor
   virtual void abort() = 0;
 
 protected:
+  /// Set of trajectories validated by executor
+  std::set<const trajectory::Trajectory*> mValidatedTrajectories;
```
jslee02: Nit: Please add an empty line between member variables.

include/aikido/control/ros/Conversions.hpp
```diff
@@ -45,7 +45,7 @@ std::unique_ptr<aikido::trajectory::Spline> toSplineJointTrajectory(
 /// \param[in] trajectory Aikido trajectory to be converted.
 /// \param[in] timestep Timestep between two consecutive waypoints.
 trajectory_msgs::JointTrajectory toRosJointTrajectory(
-    const aikido::trajectory::TrajectoryPtr& trajectory, double timestep);
+    const aikido::trajectory::ConstTrajectoryPtr& trajectory, double timestep);
```
jslee02: Nit: Could we change this to const raw pointer for the same reason above?

include/aikido/perception/ObjectDatabase.hpp
```diff
@@ -37,14 +37,14 @@ class ObjectDatabase
   /// \param[in] configDataURI The URI for the configuration information file
   ObjectDatabase(
       const dart::common::ResourceRetrieverPtr& resourceRetriever,
-      dart::common::Uri configDataURI);
+      const dart::common::Uri configDataURI);
```
jslee02: Nit: `const dart::common::Uri& configDataURI`

include/aikido/perception/YamlAprilTagsDatabase.hpp
```diff
@@ -66,7 +66,7 @@ class YamlAprilTagsDatabase : public AprilTagsDatabase
   /// \param[in] configDataURI The URI for the configuration information file
   YamlAprilTagsDatabase(
       const dart::common::ResourceRetrieverPtr& resourceRetriever,
-      dart::common::Uri configDataURI);
+      const dart::common::Uri configDataURI);
```
jslee02: Nit: `const dart::common::Uri& configDataURI`

src/control/ros/RosTrajectoryExecutor.cpp
```diff
@@ -80,21 +81,20 @@ RosTrajectoryExecutor::~RosTrajectoryExecutor()
...
 
 //==============================================================================
-void RosTrajectoryExecutor::validate(trajectory::TrajectoryPtr traj)
+void RosTrajectoryExecutor::validate(const trajectory::Trajectory* traj)
 {
-  using aikido::statespace::dart::MetaSkeletonStateSpace;
-
   if (!traj)
-    throw std::invalid_argument("Trajectory is null.");
+    throw std::invalid_argument("Traj is null.");
```
jslee02: Nit: I think this should be either `traj` or `Trajectory`.

CMakeLists.txt.user
```diff
@@ -0,0 +1,371 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
aditya-vk: oops. My bad.

include/aikido/control/ros/Conversions.hpp
aditya-vk: Same as above

src/robot/ConcreteRobot.cpp
```diff
@@ -235,7 +235,8 @@ Eigen::VectorXd ConcreteRobot::getAccelerationLimits(
 
 // ==============================================================================
 CollisionFreePtr ConcreteRobot::getSelfCollisionConstraint(
-    const MetaSkeletonStateSpacePtr& space, const MetaSkeletonPtr& metaSkeleton)
+    const MetaSkeletonStateSpacePtr& space,
+    const MetaSkeletonPtr& metaSkeleton) const
```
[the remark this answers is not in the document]
aditya-vk: So as discussed with @gilwoolee, it is best if we have the function just return `const` [example, `ConstTestablePtr` but the constraint is being passed into the planning methods and constructors which take `TestablePtr`. Since we are changing the planning API anyway, I think it is best if we wait for the corrections to happen there and make the changes here post that. I'll make an issue if that's agreeable [ :+1: ].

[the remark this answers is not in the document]
jslee02: Sounds good to me! :+1:

include/aikido/control/ros/RosTrajectoryExecutor.hpp
```diff
@@ -39,17 +39,18 @@ class RosTrajectoryExecutor : public aikido::control::TrajectoryExecutor
...
-  std::future<void> execute(trajectory::TrajectoryPtr traj) override;
+  std::future<void> execute(
+      const trajectory::ConstTrajectoryPtr& traj) override;
 
   /// Sends trajectory to ROS server for execution.
   /// \param[in] traj Trajectory to be executed.
   /// \param[in] startTime Start time for the trajectory.
   std::future<void> execute(
-      trajectory::TrajectoryPtr traj, const ::ros::Time& startTime);
+      const trajectory::ConstTrajectoryPtr traj, const ::ros::Time& startTime);
```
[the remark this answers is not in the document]
aditya-vk: This was addressed, not sure why it hasn't been updated here. On the `Files Changed` tab, I see the change reflected. Is this a GitHub 'bug' or a 'feature'? :P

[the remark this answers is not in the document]
jslee02: I think it's a bug. :disappointed:

PR #360, opened 2018-03-07
Add AppVeyor for testing on Ubuntu 16.04 LTS

jslee02:
AppVeyor recently started to support Ubuntu where the version is 16.04 LTS! 

AIKIDO works with ROS Lunar except that [`test_VectorFieldPlanner` is failing](https://ci.appveyor.com/project/jslee02/aikido/build/7/job/pjoibinflx745n7d#L2398).

`pr-control-msgs` installation is disabled since we don't have the package for Xenial yet.


jslee02:
`BarrettFingerKinematicSimulationSpreadCommandExecutorTest::execute_WaitOnFuture_CommandExecuted` [fails on AppVeyor test](https://ci.appveyor.com/project/jslee02/aikido/build/11/job/g2xox2is58frrnny#L2337).

It seems this is because `BarrettFingerKinematicSimulationSpreadCommandExecutor::step()` sometimes [throws an exception](https://github.com/personalrobotics/aikido/blob/master/src/control/BarrettFingerKinematicSimulationSpreadCommandExecutor.cpp#L158) when it's called at [here](https://github.com/personalrobotics/aikido/blob/master/tests/control/test_BarrettFingerKinematicSimulationSpreadCommandExecutor.cpp#L209). Anyone has an idea on this failure?

brianhou:
Looks fine, but I have some quick questions.

- Why is it necessary to write out all these commands versus reusing the existing Travis CI scripts? Should we update the scripts to be more flexible?
- Is there a reason that we're not using Catkin?

jslee02:
No compelling reason. I wanted to quickly test if AppVeyor works for AIKIDO, and later though it would be good to have non-Catkin build test alongside Catkin build. However, I'm open to changing to using Catkin so that we can reuse the Travis CI scripts.

jslee02:
It turned out that `pr-cleanroom` isn't compatible with Ubuntu 16.04 yet (adding unnecessary PPAs such as `ppa:libccd-debs` and `ppa:fcl-debs`; different ROS version; etc). So I just reverted the last a few changes. Let's revisit this once `pr-cleanroom` become compatible with Ubuntu 16.04.

brianhou:
Sure. Can you create an issue to do that? Thanks!

jslee02:
@brianhou Created!
* https://github.com/personalrobotics/pr-cleanroom/issues/10
* https://github.com/personalrobotics/aikido/pull/361

PR #362, opened 2018-03-09
Use catkin build on AppVeyor

jslee02:
Depends on https://github.com/personalrobotics/pr-cleanroom/pull/11


jslee02:
@gilwoolee Could you please confirm if https://github.com/personalrobotics/aikido/pull/362/commits/3ad3c3b101795377f62fc12903cb0194588c43d0 makes sense?

brianhou:
@dqyi11 might know as well, I think it might have at some point been used to generate goal and constraint TSRs for CRRT?

jslee02:
This declaration duplicates since we have this function in [`util.hpp`](https://github.com/personalrobotics/aikido/blob/master/include/aikido/robot/util.hpp#L237) as well. I think the function in the anonymous namespace was accidentally not removed or vice versa.

.ci/script_linux.sh
```diff
@@ -12,7 +12,11 @@ set -e
 ./scripts/internal-run.sh catkin build --no-status --no-deps -p 1 -i --cmake-args -DCMAKE_BUILD_TYPE=$BUILD_TYPE -DTREAT_WARNINGS_AS_ERRORS=ON -DCODECOV=$CODECOV --make-args tests -- aikido
 
 # Run tests and measure test coverage if CodeCov is on.
-if [ $CODECOV = ON ]; then ./scripts/internal-run.sh make -C build/aikido aikido_coverage; else ./scripts/internal-run.sh make -C build/aikido test; fi
+if [ $CODECOV = ON ]; then
+  ./scripts/internal-run.sh make -C build/aikido aikido_coverage
+else 
+  ./scripts/internal-run.sh cd build/aikido && ctest --output-on-failure && cd ../..
```
brianhou: Maybe instead of manually calling `ctest --output-on-failure`, we can set the `CTEST_OUTPUT_ON_FAILURE` environment variable?

jslee02: I think either way is possible. The actual problem here is that we should run `ctest` in the directory where the cmake is configured (i.e., `build/aikido`), but I don't know how to do that using `internal-run.sh`.

brianhou: Doesn't the `-C` flag to `make` already do that?

```
       -C dir, --directory=dir
            Change to directory dir before reading the makefiles or doing any-
            thing else.  If multiple -C options are specified, each is  inter-
            preted  relative to the previous one: -C / -C etc is equivalent to
            -C /etc.  This is typically used  with  recursive  invocations  of
            make.
```

jslee02: Yes, but `make` doesn't have `--output-on-failure` option. It seems none of `make test` and `ctest` has both of `-C` and `--output-on-failure` options.

brianhou: Right, that's why I was suggesting to set the environment variable instead. I think `CTEST_OUTPUT_ON_FAILURE=true make -C build/aikido test` should work.

jslee02: Hm, I presumed `CTEST_OUTPUT_ON_FAILURE` only works with `ctest` but not `make` since the environment begin with `CTEST_`. Let me try this though.

jslee02: `CTEST_OUTPUT_ON_FAILURE` doesn't work. :disappointed: Some tests are failing but the error messages are not printed.

brianhou: Hmm, this is what I see when I set `CTEST_OUTPUT_ON_FAILURE`: https://gist.github.com/brianhou/2207f29212ebceb06922d4f4cfbe08b1

I broke a test in `test_InstantaneousTrajectoryExecutor`, then ran `make -C build/aikido test` and `CTEST_OUTPUT_ON_FAILURE=true make -C build/aikido test`.

```
bhou1@r2:aikido-ws $ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 14.04.5 LTS
Release:	14.04
Codename:	trusty
bhou1@r2:aikido-ws $ ctest --version
ctest version 2.8.12.2
```

jslee02: Your test seems to show `CTEST_OUTPUT_ON_FAILURE=true` works. Could you please break a test so that it fails with an exception segfault like [this](https://ci.appveyor.com/project/jslee02/aikido/build/77/job/tjmni250aj1pkrkv#L5854)?

brianhou: [Yep](https://gist.github.com/brianhou/2207f29212ebceb06922d4f4cfbe08b1#file-gistfile2-txt).

brianhou: I think I figured it out: setting `CTEST_OUTPUT_ON_FAILURE` before calling `make -C build/aikido test` is different than `CTEST_OUTPUT_ON_FAILURE=true make -C build/aikido test`: https://gist.github.com/brianhou/2207f29212ebceb06922d4f4cfbe08b1#file-gistfile3-txt

jslee02:
AppVeyor fails due to #363. I think we could merge this PR (removing AppVeyor from required tests) since the failure is unrelated to this change.

PR #370, opened 2018-04-11
expose base class functions of same name in derived classes

aditya-vk:
Non-virtual functions in base classes have been hidden and therefore were inaccessible through the derived classes. This PR exposes them with an explicit mention.

[1] https://stackoverflow.com/questions/1628768/why-does-an-overridden-function-in-the-derived-class-hide-other-overloads-of-the
[2] https://www.ibm.com/support/knowledgecenter/en/SSLTBW_2.3.0/com.ibm.zos.v2r3.cbclx01/overload_member_fn_base_derived.htm

***

### DIFF
--- a/include/aikido/robot/ConcreteManipulator.hpp
+++ b/include/aikido/robot/ConcreteManipulator.hpp
@@ -24,7 +24,7 @@ public:
   virtual ~ConcreteManipulator() = default;
 
   // Documentation inherited.
-  virtual HandPtr getHand() override;
+  virtual ConstHandPtr getHand() const override;
 
   // Documentation inherited.
   virtual std::unique_ptr<aikido::trajectory::Spline> smoothPath(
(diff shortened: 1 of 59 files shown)


include/aikido/robot/Hand.hpp
```diff
@@ -37,7 +37,8 @@ class Hand
   virtual void step(const std::chrono::system_clock::time_point& timepoint) = 0;
 
   /// Returns the metaskeleton corresponding to this hand.
-  virtual dart::dynamics::MetaSkeletonPtr getMetaSkeleton() const = 0;
+  virtual dart::dynamics::ConstMetaSkeletonPtr getMetaSkeleton() const = 0;
+  virtual dart::dynamics::MetaSkeletonPtr getMetaSkeleton() = 0;
```
jslee02: Nit: We need docstring for an individual function. The second function wouldn't have a description. :sob:

include/aikido/robot/ConcreteManipulator.hpp
```diff
@@ -16,6 +16,11 @@ AIKIDO_DECLARE_POINTERS(ConcreteManipulator)
 class ConcreteManipulator : public Manipulator
 {
 public:
+  /// Expose base class functions
```
jslee02: [It seems Doxygen doesn't support description for the `using` declaration](https://stackoverflow.com/questions/36421785/doxygen-c-document-privately-inherited-members-exposed-by-the-using-declar). Let's make this non-doxygen comment (i.e., `// Expose base class functions`).

include/aikido/robot/ConcreteRobot.hpp
```diff
@@ -26,6 +26,10 @@ AIKIDO_DECLARE_POINTERS(ConcreteRobot)
 class ConcreteRobot : public Robot
 {
 public:
+  /// Expose base class functions
```
jslee02: Ditto with above

jslee02:
CI fails due to whitespaces in Hand.cpp ln 41.

PR #374, opened 2018-04-13
Allow KinematicSimulationTrajectoryExecutor to execute trajectories containing subsets of the skeleton.

brianhou:
This was introduced by #349.

***


src/control/KinematicSimulationTrajectoryExecutor.cpp
```diff
@@ -58,7 +58,7 @@ void KinematicSimulationTrajectoryExecutor::validate(
 
   // TODO: Delete this line once the skeleton is locked by isCompatible
   std::lock_guard<std::mutex> lock(mSkeleton->getMutex());
-  space->checkCompatibility(mSkeleton.get());
+  space->checkIfContained(mSkeleton.get());
```
aditya-vk: :+1: 
Do we need this in the `validate()` of other executors as well?

brianhou: Yeah, I think so. I'll add this soon.

PR #376, opened 2018-04-17
Mark optional dependencies in package.xml.

brianhou:
Depends on https://github.com/personalrobotics/pr-cleanroom/pull/14.

In the CI scripts, we should now run something like
```
./scripts/internal-distro.py ... -o OPTIONAL
```
where `OPTIONAL` is the name of the dependency group below.

***

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@
 * Build & Testing & ETC
 
   * Fixed Eigen memory alignment issues on 32bit Ubuntu: [#368](https://github.com/personalrobotics/aikido/pull/368)
+  * Defined optional dependencies: [#376](https://github.com/personalrobotics/aikido/pull/376)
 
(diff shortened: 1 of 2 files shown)


aditya-vk:
Looks like `libherb` and `herb3_demos` won't pass any CI tests until the optional dependencies issue isn't fixed. We should probably close this with a neat solution [force all optional dependencies to be installed?] soon before merging any PRs in the other downstream packages?

brianhou:
Here's another proposal for managing optional dependencies:
- By default, all dependencies (including optional dependencies) are required.
- Introduce `<optional>` tags (inside `<export>`) to mark which of the `<depend>`s are optional. We should be able to programmatically access those in the `pr-cleanroom/internal-distro.py` script and remove them from the set of dependencies (based on some CLI flag).

I've updated `package.xml` to match this proposal, and will update the `internal-distro.py` script next.

PR #379, opened 2018-04-21
Planning calls with MetaSkeleton explicitly set dofs for InverseKinematics solver

gilwoolee:
This PR fixes the following issues.

1. Planning calls which create `dart::dynamics::InverseKinematics` explicitly sets the IK solver to use `metaSkeleton`'s dof values.
2. `InverseKinematicsSampleable` checks for the number of dofs and throws an error if the value is zero.

***


src/constraint/dart/InverseKinematicsSampleable.cpp
```diff
@@ -83,6 +83,13 @@ InverseKinematicsSampleable::InverseKinematicsSampleable(
     throw std::invalid_argument("InverseKinematics is nullptr.");
 
   const auto ikSkeleton = mInverseKinematics->getNode()->getSkeleton();
+
+  if (mInverseKinematics->getDofs().size() == 0)
```
brianhou: Nit: use `mInverseKinematics->getDofs().empty()` instead?

src/robot/util.cpp
```diff
@@ -180,13 +180,17 @@ InterpolatedPtr planToTSR(
     double timelimit,
     std::size_t maxNumTrials)
 {
+  // Create an IK solver with metaSkeleton dofs.
+  auto ik = InverseKinematics::create(bn);
+  ik->setDofs(metaSkeleton->getDofs());
```
jslee02: I haven't checked whether this would work even for the case that `metaSkeleton` contains bodies from difference skeletons, which I don't think so. Could wait for me to ensure this?

If so, DART should inform this, but let's have a validity check in AIKIDO or at least a comment here.

gilwoolee: Ah, good point. Let me add a quick check here with a comment. :)

mxgrey: Currently `InverseKinematics::setDofs(~)` does assume that you're only passing it DoF objects that belong to the current Skeleton, and it naively grabs whatever their indices are. I think that can be considered a bug (or at least unintuitive behavior) in DART.

I agree with JS that we should change DART to have the `IK::setDofs(~)` skip over any `DegreeOfFreedom*` objects that don't belong to the IK's parent Skeleton.

src/robot/util.cpp
```diff
@@ -180,13 +180,29 @@ InterpolatedPtr planToTSR(
...
+  // TODO: DART may be updated to check for single skeleton
+  if (metaSkeleton->getNumDofs() == 0)
+    throw std::invalid_argument("MetaSkeleton has 0 degree of freedon.");
+
+  auto skeleton = metaSkeleton->getDof(0)->getSkeleton();
+  for (size_t i = 1; i < metaSkeleton->getNumDofs(); ++i)
+  {
+    if (metaSkeleton->getDof(i)->getSkeleton() != skeleton)
+      throw std::invalid_argument("MetaSkeleton has more than 1 skeleton.");
+  }
```
jslee02: This check looks good to me. I think we need this check in `planToTSRwithTrajectoryConstraint()` as well.

src/robot/util.cpp
```diff
@@ -180,13 +180,29 @@ InterpolatedPtr planToTSR(
     double timelimit,
     std::size_t maxNumTrials)
 {
+  // Create an IK solver with metaSkeleton dofs.
+  auto ik = InverseKinematics::create(bn);
+
+  // TODO: DART may be updated to check for single skeleton
+  if (metaSkeleton->getNumDofs() == 0)
+    throw std::invalid_argument("MetaSkeleton has 0 degree of freedon.");
```
aditya-vk: Typo: freedon -> freedom; degree -> degrees

src/robot/util.cpp
```diff
@@ -303,9 +315,22 @@ InterpolatedPtr planToTSRwithTrajectoryConstraint(
   std::shared_ptr<Sampleable> seedConstraint
       = std::move(createSampleableBounds(space, crrtParameters.rng->clone()));
 
-  // crate IK
+  // TODO: DART may be updated to check for single skeleton
+  if (metaSkeleton->getNumDofs() == 0)
+    throw std::invalid_argument("MetaSkeleton has 0 degree of freedon.");
```
aditya-vk: Same as above

brianhou:
Great! Let's merge once appveyor is happy.

PR #382, opened 2018-04-25
Vector field planner parameters fix

Tonkel:
Default values for VectorFieldPlannerParameters constructor (as talked about with Gilwoo).

### DIFF
--- a/include/aikido/robot/util.hpp
+++ b/include/aikido/robot/util.hpp
@@ -24,14 +24,13 @@ namespace util {
 
 struct VectorFieldPlannerParameters
 {
-  VectorFieldPlannerParameters() = default;
   VectorFieldPlannerParameters(
-      double linearVelocity,
-      double negativeDistanceTolerance,
-      double positiveDistanceTolerance,
-      double initialStepSize,
-      double jointLimitTolerance,
-      double constraintCheckResolution,
+      double linearVelocity = 0.2,
+      double negativeDistanceTolerance = 0.001,
+      double positiveDistanceTolerance = 0.001,
+      double initialStepSize = 0.001,
+      double jointLimitTolerance = 1e-3,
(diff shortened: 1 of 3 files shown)


Tonkel:
this branch is based off of bugfix/IK. Should it merge back to bugfix/IK or master?

brianhou:
I think these parameters can dramatically affect the runtime and success rates of the VectorField planner, so it makes more sense to me to set these in each `librobot`. Why is this change necessary?

gilwoolee:
In the current librobots none of these are set -- just the default constructors are being used, which leads to planning failures or segfaults. We should either explicitly remove the default constructor and force the librobots to set them to some values, or provide some default values. I prefer having a set of good default values that have been benchmarked. @dqyi11 , I asked @Tonkel to copy these values from unit tests; could you verify that these are reasonable parameter values to use?

dqyi11:
Those parameters have not been benchmarked. They were selected for a simple "robot" created in unit tests. Some of the parameters can vary significantly in problems. For example, `linearVelocity` defines how fast we would hope an end-effector moving forward (which usually depends on the movement distance and time). 

Also planning for different functionalities using different vector fields takes different parameters. This one seems for `planToEndEffectorOffset`. Those for `planToEndEffectorPose` would be slightly different. 
If we are going to define default parameters in this way, will it be better to define by functionalities, like `planToTSR`, `planToEndEffectorPose`, etc?

gilwoolee:
Thanks for the comments @dqyi11 ! That makes sense. These were meant originally as a convenient methods until we have planner API, so how about we close this PR and as we work on the planners, set benchmarked planner x problem-specific params? 

@Tonkel Let's keep these only to your local works for ADA for now (you can directly set these params in libada), and revisit after the planner APIs are done.

PR #385, opened 2018-04-26
Bugfix/planning locks

Tonkel:
Fixed locking issue where planner locked the robot and then called another planner which tried to lock it again.

Exposed the pointer to the ConcreteRobot's RNG, so we can use it for the CRRTPlannerParameters RNG.

PR #386, opened 2018-04-27
Fixed Planning Locks

Tonkel:
This is a cleaned up version of #385, it fixes a locking issue where one planning function locks the robot and then may call another planning function which attempts to lock it again.


brianhou:
I agree, at least until we figure out a better way to handle locking with meta-planners. Thanks for starting this @Tonkel!

src/planner/vectorfield/VectorFieldPlanner.cpp
```diff
@@ -141,6 +141,11 @@ std::unique_ptr<aikido::trajectory::Spline> planToEndEffectorOffset(
     std::chrono::duration<double> timelimit,
     planner::PlanningResult* planningResult)
 {
+
+  // ensure that no two planners run at the same time
+  auto robot = metaskeleton->getBodyNode(0)->getSkeleton();
```
jslee02: It should return `nullptr` (with warning) or throw an exception if the number of bodies is zero.

src/planner/vectorfield/VectorFieldPlanner.cpp
```diff
@@ -141,6 +141,11 @@ std::unique_ptr<aikido::trajectory::Spline> planToEndEffectorOffset(
     std::chrono::duration<double> timelimit,
     planner::PlanningResult* planningResult)
 {
+
```
jslee02: Nit: Remove whitespace.

src/planner/vectorfield/VectorFieldPlanner.cpp
Tonkel: good idea :)
done

src/planner/vectorfield/VectorFieldPlanner.cpp
Tonkel: done.

src/planner/vectorfield/VectorFieldPlanner.cpp
```diff
@@ -141,6 +141,14 @@ std::unique_ptr<aikido::trajectory::Spline> planToEndEffectorOffset(
     std::chrono::duration<double> timelimit,
     planner::PlanningResult* planningResult)
 {
+  // ensure that no two planners run at the same time
+  if (metaskeleton->getNumBodyNodes() <= 0)
```
jslee02: Nit: This is technically fine, but negativity check is unnecessary here because the return type is unsigned.

src/planner/vectorfield/VectorFieldPlanner.cpp
```diff
@@ -141,6 +141,14 @@ std::unique_ptr<aikido::trajectory::Spline> planToEndEffectorOffset(
...
+  {
+      throw std::runtime_error("MetaSkeleton doesn't have any body nodes.");
```
jslee02: The code formatting check wouldn't be happy with 4-space indentation. :smile: Please run `make format` to format the code (make sure `clang-format-3.8` is installed).

gilwoolee:
LGTM!

PR #387, opened 2018-04-27
exposed RNG pointer of ConcreteRobot

Tonkel:
This is a cleaned up version of #385, regarding ConcreteRobot's RNG and how it should be exposed. There are also some comments about this in #385.


gilwoolee:
Closing this since `librobot` typically has its own `mRNG` to be cloned and managed directly.

PR #388, opened 2018-04-28
Herbathon: add makeBodyFromUrdf convenience function

Tonkel:
Implemented for the Herbathon. I thought a small util file would be appropriate for stuff like this. What do you think?

I tested the function with the food manipulation demo.


include/aikido/io/util.hpp
```diff
@@ -0,0 +1,22 @@
...
+#include <dart/dart.hpp>
+#include <dart/utils/urdf/DartLoader.hpp>
+
+using dart::dynamics::SkeletonPtr;
+
+namespace aikido {
+namespace io {
+
+const SkeletonPtr makeBodyFromURDF(
+    const std::shared_ptr<aikido::io::CatkinResourceRetriever>
```
brianhou: Maybe this can be a `dart::common::ResourceRetrieverPtr`, since that's what the DART URDF loader is expecting.

include/aikido/io/util.hpp
```diff
@@ -0,0 +1,22 @@
...
+#include <aikido/io/CatkinResourceRetriever.hpp>
+#include <dart/dart.hpp>
+#include <dart/utils/urdf/DartLoader.hpp>
+
+using dart::dynamics::SkeletonPtr;
+
+namespace aikido {
+namespace io {
+
+const SkeletonPtr makeBodyFromURDF(
```
brianhou: Let's rename this to `loadSkeletonFromURDF`.

include/aikido/io/util.hpp
```diff
@@ -0,0 +1,22 @@
...
+    const std::shared_ptr<aikido::io::CatkinResourceRetriever>
+        resourceRetriever,
+    const std::string& uri,
+    const Eigen::Isometry3d& transform);
```
brianhou: Maybe we can make this an optional argument by setting its default value to identity.

include/aikido/io/util.hpp
jslee02: Nit: Returning by const value is generally unnecessary. Let's just return `SkeletonPtr`.

include/aikido/io/util.hpp
```diff
@@ -0,0 +1,22 @@
+#ifndef AIKIDO_IO_UTIL_HPP_
+#define AIKIDO_IO_UTIL_HPP_
+
+#include <aikido/io/CatkinResourceRetriever.hpp>
+#include <dart/dart.hpp>
+#include <dart/utils/urdf/DartLoader.hpp>
+
+using dart::dynamics::SkeletonPtr;
```
jslee02: We should avoid `using` in headers because it can pollute namespaces by including this header. Let's remove and use explicitly `dart::dynamics::SkeletonPtr` in headers. You could use `using` in sources.

include/aikido/io/util.hpp
```diff
@@ -0,0 +1,22 @@
+#ifndef AIKIDO_IO_UTIL_HPP_
+#define AIKIDO_IO_UTIL_HPP_
+
+#include <aikido/io/CatkinResourceRetriever.hpp>
```
jslee02: Nit: Our convention is to:
(1) use `"..."` for including AIKIDO headers
(2) include headers in order (in header file): (a) system headers (i.e., STL), (2) 3rd party libraries, (3) AIKIDO headers.

So this should be:
```cpp
#include <dart/dart.hpp>
#include <dart/utils/urdf/DartLoader.hpp>
#include "aikido/io/CatkinResourceRetriever.hpp"
```

src/io/util.cpp
```diff
@@ -0,0 +1,24 @@
...
+    const Eigen::Isometry3d& transform)
+{
+  dart::utils::DartLoader urdfLoader;
+  const SkeletonPtr skeleton = urdfLoader.parseSkeleton(uri, resourceRetriever);
+
+  if (!skeleton)
+    throw std::runtime_error("unable to load '" + uri + "'");
+
+  dynamic_cast<dart::dynamics::FreeJoint*>(skeleton->getJoint(0))
+      ->setTransform(transform);
```
jslee02: Need nullity check after dynamic casting.

src/io/util.cpp
```diff
@@ -0,0 +1,24 @@
...
+    const std::string& uri,
+    const Eigen::Isometry3d& transform)
+{
+  dart::utils::DartLoader urdfLoader;
+  const SkeletonPtr skeleton = urdfLoader.parseSkeleton(uri, resourceRetriever);
+
+  if (!skeleton)
+    throw std::runtime_error("unable to load '" + uri + "'");
+
+  dynamic_cast<dart::dynamics::FreeJoint*>(skeleton->getJoint(0))
```
brianhou: I think we can also make this `skeleton->getRootJoint()` instead of `skeleton->getJoint(0)`. @jslee02 can you confirm?

jslee02: Correct!

Also, we should check whether `skeleton` is not empty.

include/aikido/io/util.hpp
Tonkel: I'll change it but could you tell me if there is a specific reason for the header include order? Does it have an advantage to do it this way?

include/aikido/io/util.hpp
```diff
@@ -0,0 +1,19 @@
...
+#define AIKIDO_IO_UTIL_HPP_
+
+#include <dart/dart.hpp>
+#include <dart/utils/urdf/DartLoader.hpp>
+#include "aikido/io/CatkinResourceRetriever.hpp"
+
+namespace aikido {
+namespace io {
+
+dart::dynamics::SkeletonPtr loadSkeletonFromURDF(
```
gilwoolee: Needs docstring.

src/io/util.cpp
```diff
@@ -0,0 +1,37 @@
+#include "aikido/io/util.hpp"
+
+namespace aikido {
+namespace io {
+
+dart::dynamics::SkeletonPtr loadSkeletonFromURDF(
```
gilwoolee: Nit: We typically put "//==================================" at the beginning of a new method. Please check other cpp files.

src/io/util.cpp
```diff
@@ -0,0 +1,37 @@
...
+    const dart::common::ResourceRetrieverPtr resourceRetriever,
+    const std::string& uri,
+    const Eigen::Isometry3d& transform = Eigen::Isometry3d::Identity())
```
gilwoolee: I would prefer making the default param `Eigen::Isometry3d::identity()` be specified at the header file to make it more explicit.

src/io/util.cpp
```diff
@@ -0,0 +1,37 @@
...
+{
+  dart::utils::DartLoader urdfLoader;
+  const dart::dynamics::SkeletonPtr skeleton
+      = urdfLoader.parseSkeleton(uri, resourceRetriever);
+
+  if (!skeleton)
+    throw std::runtime_error("unable to load '" + uri + "'");
```
gilwoolee: Nit: `unable`-> `Unable`

src/io/util.cpp
```diff
@@ -0,0 +1,37 @@
...
+    throw std::runtime_error("Skeleton is empty.");
+  }
+
+  auto freeJoint
+      = dynamic_cast<dart::dynamics::FreeJoint*>(skeleton->getRootJoint(0));
+
+  if (freeJoint == nullptr)
+  {
+    throw std::runtime_error(
+        "unable to cast Skeleton's root joint to FreeJoint.");
```
gilwoolee: Nit: `unable`->`Unable`

include/aikido/io/util.hpp
psigen: This header ordering generally follows the Google C++ style convention:
https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes

Issue #384, opened 2018-04-26
Sampling IKs close to the current configuration


aditya-vk:
Even Snap Planners can potentially give out very long paths if the IK sampled is very far in the joint space from the current configuration. The sampling strategy should account for the current configuration and sample IKs minimizing the joint movements.

@jslee02 This should be easy to do with `DART` right? We should tie the sampler to an objective function or is there a simpler solution?

brianhou:
Please make this issue title more descriptive. Ideally, we should be able to read the title and have some idea of what this issue is about.

aditya-vk:
The problem stems from how `planToTSR` currently is. The seed values for the IK solver need to be more intelligently picked and not arbitrarily. If arbitrary is fine, then it doesn't hurt to start with the current configuration as the initial guess.

aditya-vk:
Solution discussed offline to address this correctly:
1. Create a Sequential Sampleable: https://github.com/personalrobotics/aikido/pull/393
2. API to make IKSampleable to return a set of IK solutions
3. Rank the samples obtained.

planToTSR takes in a SequentialSampleable with [FiniteSampleable(initialConfiguration), Sampleable]

brianhou:
I don't think (2) requires an API change: we can just keep calling `IkSampleable::sample` when trying to plan to a TSR.

PrPy defined a few methods for ranking that we should implement. The one that we probably care about the most (at least for this current issue) is [`NominalConfiguration`](https://github.com/personalrobotics/prpy/blob/master/src/prpy/ik_ranking.py#L55).

I think we can also just allow `planToTSR` to take in the uniform `Sampleable` over the configuration space (what it's probably using now) and then augment that inside of `planToTSR` before we pass that to `IkSampleable` by creating the `SequentialSampleable` as you described.

aditya-vk:
This has been addressed by #393 and #423 .

PR #389, opened 2018-05-01
Allow for current configuration to be used as seed for solving IK

aditya-vk:
Allows the current configuration to be used as a seed (initial guess) by the DART IK Solver. This is especially useful when solving for trajectories. IF the current configuration fails to give a solution, we allow for random configuration to be used as initial guesses for the solver.

Looking for comments on the general solution/idea for doing this. A few things:
1. We could have a sampler that samples from a finite set and use them as initial seeds before trying random initial guesses. This can, however, be set at a higher level. For example in `planToTSR` we might want to pass in a sampler that samples from a region around the current configuration since we want smooth and short trajectories. [will be addressed in a different PR] 
2. We can get rid of random sampler as an input to `inverseKinematicsSampleable` completely and instead pass a vector of seeds which can be preferred before using random guesses [and pass these seeds to DART solver]. However, as discussed with @jslee02 this can be disadvantageous since a Sampleable is potentially more generic/extendable, and the user might want to sample more intelligently when required than DARTs general sampler. Same example as in point 1.

Need to add tests to test both boolean instances for whether to use the current configuration as a seed.

***

### DIFF
--- a/include/aikido/constraint/dart/InverseKinematicsSampleable.hpp
+++ b/include/aikido/constraint/dart/InverseKinematicsSampleable.hpp
@@ -31,12 +31,13 @@ public:
   /// \param _maxNumTrials Max number of trials for its sample generator
   ///        to retry sampling and finding an inverse kinematics solution.
   InverseKinematicsSampleable(
-      statespace::dart::MetaSkeletonStateSpacePtr _metaSkeletonStateSpace,
-      ::dart::dynamics::MetaSkeletonPtr _metaskeleton,
-      SampleablePtr _poseConstraint,
-      SampleablePtr _seedConstraint,
-      ::dart::dynamics::InverseKinematicsPtr _inverseKinematics,
-      int _maxNumTrials);
+      statespace::dart::MetaSkeletonStateSpacePtr metaSkeletonStateSpace,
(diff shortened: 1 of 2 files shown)


brianhou:
I think I would prefer the first solution that you proposed (introducing a new `Sampleable` class), rather than the one implemented in this PR.

That said, I don't believe that either of these solutions fixes the issue that you raised in https://github.com/personalrobotics/aikido/issues/384. To do that, I think we'd need to have a way to rank IK solutions, similar to what we had in prpy.

aditya-vk:
1. I prefer the first method as well.
2. This PR only corrects for the usage of the current configuration. We were previously sampling random configurations for the initial guess. I don't think this completely solves the issue in #384 either but it is an essential component. 
3. I would still have the tiny change in this PR even if there's a new sampleable class (~composite) which samples from both the finite sampleable provided by the user and from a random configuration sampler. I'll leave the new sampler to another PR, it shouldn't change how this class is written.
4. I think the ranking should be done outside this class [with the relevant criterion by the user, or in `planToTSR` by computing the distances between current position and IK]. This would need a way to return multiple IK solutions from the ikSampleable.

brianhou:
As we discussed in person, let's close this PR and implement a new `Sampleable` class.

PR #393, opened 2018-05-05
Create a Sequential Sampleable

aditya-vk:
This PR adds a Sequential Sampleable. It samples in the order given, moving to the next one upon exhaustion of the current sampleable. This is useful, for instance, in solving IK, in sampling the current configuration first before other random/informed sampling.

***


src/constraint/SequentialSampleable.cpp
```diff
@@ -0,0 +1,117 @@
...
+  SequentialSampleGenerator(
+      statespace::StateSpacePtr stateSpace,
+      std::vector<std::unique_ptr<SampleGenerator>> generators)
+    : mStateSpace(std::move(stateSpace))
+    , mGenerators(std::move(generators))
+    , mIndex(0)
+  {
+    // Do nothing
+    // TODO (avk): Should we again check equivalence of statespaces between
+    // that of each generator and mStateSpace? I don't think it's necessary.
```
jslee02: Let's use assertion in this case; we're pretty sure this case never happen but just ensure that in debug mode.

jslee02: It seems this class is vulnerable in the case of empty `generators`.

src/constraint/SequentialSampleable.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+    if (canSample())
+      return mGenerators[mIndex]->sample(state);
+
+    return false;
+  }
+
+  int getNumSamples() const override
+  {
+    return mGenerators[mIndex]->getNumSamples();
```
jslee02: I'm not sure if this logic is following the description of `getNumSamples()`:
> Gets an upper bound on the number of samples remaining or NO_LIMIT.

src/constraint/SequentialSampleable.cpp
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  bool canSample() const override
+  {
+    if (mIndex >= mGenerators.size())
+      return false;
+
+    return true;
```
jslee02: I think this function should return true when:

1. the number of remaining generators is not zero _and_
2. one of the remaining generators is able to sample.

It seems the current implementation is missing 2.

src/constraint/SequentialSampleable.cpp
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+private:
+  statespace::StateSpacePtr mStateSpace;
+  std::vector<std::unique_ptr<SampleGenerator>> mGenerators;
+  std::size_t mIndex;
```
jslee02: Please add docstrings.

src/constraint/SequentialSampleable.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+  statespace::StateSpacePtr getStateSpace() const override
+  {
+    return mStateSpace;
+  }
+
+  bool sample(statespace::StateSpace::State* state)
+  {
+    if (!mGenerators[mIndex]->canSample())
+      mIndex++;
```
jslee02: I think this function should keep iterating until it finds a sampleable generator instead of just check only one generator.

src/constraint/SequentialSampleable.cpp
```diff
@@ -0,0 +1,117 @@
+#include <aikido/constraint/SequentialSampleable.hpp>
+
+#include <dart/common/StlHelpers.hpp>
+
+namespace aikido {
+namespace constraint {
+
+// TODO (avk): Should I declare the class first and then define?
+// What is "good" C++ convention? Don't tell me it's a preference JS! -.-
```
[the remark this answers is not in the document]
aditya-vk: In that case, shall we add docstring during definitions, which we generally do not do right?

src/constraint/SequentialSampleable.cpp
aditya-vk: `getNumSamples()` returns 0 after checking for empty vector.

src/constraint/SequentialSampleable.cpp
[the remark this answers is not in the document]
jslee02: > shall we add docstring during definitions

I think so. Although they are not exposed to the users, it would be very helpful for us (i.e., developers).

include/aikido/constraint/SequentialSampleable.hpp
```diff
@@ -0,0 +1,39 @@
...
+/// the next one in the set upon exhaustion.
+class SequentialSampleable : public Sampleable
+{
+public:
+  /// Constructor.
+  /// \param[in] stateSpace StateSpace in which this constraint operates.
+  /// \param[in] sampleables Set of sampleables.
+  SequentialSampleable(
+      statespace::StateSpacePtr stateSpace,
+      std::vector<SampleablePtr>& sampleables);
```
jslee02: I meant passing by **const** reference rather than just reference. :smile:

include/aikido/constraint/SequentialSampleable.hpp
```diff
@@ -0,0 +1,39 @@
...
+  // TODO (avk): const-correctness after planner API is merged.
+  statespace::StateSpacePtr getStateSpace() const override;
+
+  // Documentation inherited.
+  std::unique_ptr<SampleGenerator> createSampleGenerator() const override;
+
+private:
+  /// StateSpace in which the constraint operates.
+  statespace::StateSpacePtr mStateSpace;
+  /// Set of sampleables.
```
jslee02: Nit: Add empty line between members.

src/constraint/SequentialSampleable.cpp
```diff
@@ -0,0 +1,132 @@
...
+namespace aikido {
+namespace constraint {
+
+//==============================================================================
+class SequentialSampleGenerator : public SampleGenerator
+{
+public:
+  SequentialSampleGenerator(
+      statespace::StateSpacePtr stateSpace,
+      std::vector<std::unique_ptr<SampleGenerator>>& generators)
```
jslee02: I think we should pass by value for `generators` (and move) because it's transferring the onwerships.

src/constraint/SequentialSampleable.cpp
```diff
@@ -0,0 +1,132 @@
...
+    : mStateSpace(std::move(stateSpace))
+    , mGenerators(generators)
+    , mIndex(0)
+  {
+    for (std::size_t i = 0; i < mGenerators.size(); ++i)
+      assert(mGenerators[i]->getStateSpace() == mStateSpace);
```
jslee02: Please note that `assert` is ignored in release mode. This code will not be built in release mode since the for-statement is empty.

We could either:
```cpp
#ifndef NDEBUG // debug mode
  for (const auto& generator : mGenerators)
    assert(generator->getStateSpace() == mStateSpace);
#endif
```
or
```cpp
  for (const auto& generator : mGenerators)
  {
    assert(generator->getStateSpace() == mStateSpace);
  }
```

I slightly prefer former version. :smile:

src/constraint/SequentialSampleable.cpp
jslee02: It sounds like `canSample()` works correctly only when it's used in `sample()`, which I don't think that's a good idea. Also, this doesn't agree with the [docstrion of `canSample()`](https://github.com/personalrobotics/aikido/blob/master/include/aikido/constraint/Sampleable.hpp#L60):
> /// Returns whether getNumSamples() > 0.

I propose changing this function to return `getNumSamples() > 0`; and changing `canSample()` not to call this function and to check the size of `mGenerators` directly in `canSample()`.

src/constraint/SequentialSampleable.cpp
```diff
@@ -0,0 +1,132 @@
...
+  bool sample(statespace::StateSpace::State* state)
+  {
+    while (canSample())
+    {
+      if (mGenerators[mIndex]->canSample())
+        return mGenerators[mIndex]->sample(state);
+
+      mIndex++;
+    }
+    return false;
```
jslee02: This logic (i.e., "try the current generator to sample and return the result") is possible to return false even when one of the remaining generators can sample, which I don't think that's the expected behavior of `sample`.

I propose changing this logic to something like:
```python
index2 = index
for generator in generators[index2:]:
  result = generator.sample(state);
  index += 1
  if result:
    return True
return False
```
so that it tries to sample until it can sample and returns false when all the remaining generators are not able to draw a sample.

src/constraint/SequentialSampleable.cpp
brianhou: Would it be possible for one of the remaining generators to not be sampleable? Then even though you have generators remaining (this implementation would return `true`), it still seems like you wouldn't be sample (logically, maybe it should return `false`). What if we just do something like

```
for (std::size_t i = mIndex; i < mGenerators.size(); ++i)
  if (mGenerators[i].canSample())
    return true;
return false;

File: src/constraint/SequentialSampleable.cpp [hunk 12]
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  // Documentation inherited
+  int getNumSamples() const override
+  {
+    if (mGenerators.empty())
```
brianhou: I think this might be unnecessary. If this happens, the for loop wouldn't execute and `numSamples` would be 0 anyway.

src/constraint/SequentialSampleable.cpp
jslee02: @brianhou 's version would be more efficient because returns `getNumSamples() > 0` can be more expensive as it computes the exact number of samples while @brianhou 's version can early terminate.

src/constraint/SequentialSampleable.cpp
aditya-vk: Ah. Got it. 
Your proposal will account for `sample() may transiently fail, i.e. return false, at any point before then`

src/constraint/SequentialSampleable.cpp
```diff
@@ -0,0 +1,137 @@
...
+{
+public:
+  SequentialSampleGenerator(
+      statespace::StateSpacePtr stateSpace,
+      std::vector<std::unique_ptr<SampleGenerator>> generators)
+    : mStateSpace(std::move(stateSpace))
+    , mGenerators(std::move(generators))
+    , mIndex(0)
+  {
+    #ifndef NDEBUG
```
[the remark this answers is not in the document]
jslee02: We haven't explicitly decided yet, but most people don't indent due to preprocessors. So my preference is like:
```cpp
  {
#ifndef NDEBUG // debug mode
    for (...)
      assert(...);
#endif
  }
```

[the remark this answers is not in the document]
aditya-vk: Looks like our current settings for formatting prefer the former.

src/constraint/SequentialSampleable.cpp
```diff
@@ -0,0 +1,137 @@
...
+  {
+    while (canSample())
+    {
+      if (mGenerators[mIndex]->canSample())
+      {
+        if (mGenerators[mIndex]->sample(state))
+          return true;
+      }
+      mIndex++;
+    }
```
jslee02: This logic is still inefficient because of calling `canSample()`; `canSample()` iterates over the generators remaining while `canSample()` is called repeatidly in this logic.

It seems that calling both of `canSample()` is not necessary. What would be the problem if we remove calling `canSample()` and just calling `sample(...)` iterating the generators?

aditya-vk: Yeah that was my concern as well. I wondered if changing `mIndex` within `canSample()` would make it any better but I guess that wouldn't be as clean or ideal.

jslee02: I don't think that's a desirable behavior.

Could you point me out what's your concern with the following proposal?:
```cpp
for (auto i = mIndex; i < mGenerators.size(); ++i)
{
  if (mGenerators[mIndex]->sample(state))
    return true;

  mIndex++;
}

return false;
```

src/constraint/SequentialSampleable.cpp
```diff
@@ -0,0 +1,137 @@
...
+  SequentialSampleGenerator(
+      statespace::StateSpacePtr stateSpace,
+      std::vector<std::unique_ptr<SampleGenerator>> generators)
+    : mStateSpace(std::move(stateSpace))
+    , mGenerators(std::move(generators))
+    , mIndex(0)
+  {
+#ifndef NDEBUG
+    for (const auto& generator : mGenerators)
+      assert(generator->getStateSpace() == mStateSpace);
```
jslee02: Nit: Could we add an assertion for checking nullity of the generators?

src/constraint/SequentialSampleable.cpp
aditya-vk: The current [updated] implementation has the same behavior as your proposal.

aditya-vk: Oh wait, actually I think it is beneficial to get rid of `canSample()`, for instance if a `SequentialSampler` is composed of more `SequentialSampler`s, the current implementation would be still inefficient.

src/constraint/SequentialSampleable.cpp
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  // Documentation inherited
+  statespace::StateSpacePtr getStateSpace() const override
+  {
+    return mStateSpace;
+  }
+
+  // Documentation inherited
+  bool sample(statespace::StateSpace::State* state)
```
jslee02: Nit: Missing override specifier.

src/constraint/SequentialSampleable.cpp
```diff
@@ -0,0 +1,135 @@
+#include "aikido/constraint/SequentialSampleable.hpp"
+
+#include <dart/common/StlHelpers.hpp>
```
jslee02: Nit: I guess you meant `dart/common/Memory.hpp` for [`dart::common::make_unique`].

include/aikido/constraint/SequentialSampleable.hpp
```diff
@@ -0,0 +1,40 @@
...
+  statespace::StateSpacePtr getStateSpace() const override;
+
+  // Documentation inherited.
+  std::unique_ptr<SampleGenerator> createSampleGenerator() const override;
+
+private:
+  /// StateSpace in which the constraint operates.
+  statespace::StateSpacePtr mStateSpace;
+
+  /// Set of sampleables.
```
brianhou: Nit: `set` -> `sequence`

include/aikido/constraint/SequentialSampleable.hpp
```diff
@@ -0,0 +1,40 @@
+#ifndef AIKIDO_CONSTRAINT_SEQUENTIALSAMPLEABLE_HPP_
+#define AIKIDO_CONSTRAINT_SEQUENTIALSAMPLEABLE_HPP_
+
+#include "aikido/constraint/Sampleable.hpp"
+
+namespace aikido {
+namespace constraint {
+
+/// Sampleable supporting a composition of multiple sampleables which
```
brianhou: I think this is confusing. Here's my stab at it:

`Sampleable` that wraps a sequence of `Sampleable`s. Exhausts the current `Sampleable` before sampling from the next `Sampleable` in the sequence.

src/constraint/SequentialSampleable.cpp
```diff
@@ -0,0 +1,145 @@
...
+      }
+    }
+
+#ifndef NDEBUG
+    for (const auto& generator : mGenerators)
+      assert(generator->getStateSpace() == mStateSpace);
+#endif
+  }
+
+  // Documentation inherited
```
brianhou: Nit: `//==============================================================================`, not `// Documentation inherited`.

src/constraint/SequentialSampleable.cpp
```diff
@@ -0,0 +1,145 @@
...
+  statespace::StateSpacePtr getStateSpace() const override
+  {
+    return mStateSpace;
+  }
+
+  // Documentation inherited
```
brianhou: Nit: `//==============================================================================`, not `// Documentation inherited`.

src/constraint/SequentialSampleable.cpp
```diff
@@ -0,0 +1,145 @@
...
+    {
+      if (mGenerators[mIndex]->sample(state))
+        return true;
+
+      mIndex++;
+    }
+    return false;
+  }
+
+  // Documentation inherited
```
brianhou: Nit: `//==============================================================================`, not `// Documentation inherited`.

src/constraint/SequentialSampleable.cpp
```diff
@@ -0,0 +1,145 @@
...
+    {
+      if (mGenerators[i]->getNumSamples() == NO_LIMIT)
+        return NO_LIMIT;
+
+      numSamples += mGenerators[i]->getNumSamples();
+    }
+    return numSamples;
+  }
+
+  // Documentation inherited
```
brianhou: Nit: `//==============================================================================`, not `// Documentation inherited`.

src/constraint/SequentialSampleable.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+    if (sampleable->getStateSpace() != mStateSpace)
+    {
+      std::stringstream msg;
+      msg << "Mismatch between statespace of Sampleable " << i
+          << " and given statespace.";
+      throw std::invalid_argument(msg.str());
+    }
+  }
+  // TODO (avk): Somewhere we need to give warning if initial samplers are
```
brianhou: This is a good idea. I think something like "the i'th sampleable is infinite, will ignore N subsequent sampleables" would be helpful (if N > 0).

src/constraint/SequentialSampleable.cpp
aditya-vk: @jslee02  Should it be an _assertion_? :thinking:

src/constraint/SequentialSampleable.cpp
aditya-vk: In that case, I prefer the current formatting where the declaration happens before definition.

src/constraint/SequentialSampleable.cpp
```diff
@@ -0,0 +1,174 @@
...
+  for (const auto& generator : mGenerators)
+    assert(generator->getStateSpace() == mStateSpace);
+#endif
+
+  for (std::size_t i = 0u; i < mGenerators.size() - 1; ++i)
+  {
+    if (mGenerators[i]->getNumSamples() == NO_LIMIT)
+      dtwarn << "Sampleable " << i << " is infinite. "
+             << "Remaining " << mGenerators.size() - i - 1
+             << " sampleable(s) will potentially be ignored.";
```
jslee02: This warning seems very informative! It would be slightly better:
* to `break` once print a warning
* not to warn if `i` is equal to `mGenerators.size() - 1`; otherwise, `mGenerators.size() - i - 1` would be the greatest number of `std::size_t`.

brianhou: Agree with the first point, but I think the second point should never happen since the for loop condition is `i < mGenerators.size() - 1`?

jslee02: Oh, I missed that. Please ignore the second point.

src/constraint/SequentialSampleable.cpp
```diff
@@ -0,0 +1,177 @@
...
+class SequentialSampleGenerator : public SampleGenerator
+{
+public:
+  /// Sample Generator for Sequential Sampleable
+  /// \param[in] stateSpace Statespace the associated constraint operates in
+  /// \param[in] generators Sequence of generators associated with corresponding
+  /// sequence of sampleables
+  SequentialSampleGenerator(
+      statespace::StateSpacePtr stateSpace,
+      std::vector<std::unique_ptr<SampleGenerator>> generators);
```
brianhou: Should this be passed by reference? (https://app.codacy.com/app/personalrobotics/aikido/pullRequest?prid=1615713)

src/constraint/SequentialSampleable.cpp
```diff
@@ -0,0 +1,177 @@
...
+  std::vector<std::unique_ptr<SampleGenerator>> mGenerators;
+
+  /// Index of the active sampleable/generator.
+  std::size_t mIndex;
+};
+
+//==============================================================================
+SequentialSampleGenerator::SequentialSampleGenerator(
+    statespace::StateSpacePtr stateSpace,
+    std::vector<std::unique_ptr<SampleGenerator>> generators)
```
brianhou: Should this be passed by reference? (https://app.codacy.com/app/personalrobotics/aikido/pullRequest?prid=1615713)

src/constraint/SequentialSampleable.cpp
jslee02: Passing by _const_ reference is not allowed because `std::unique_ptr` is not copiable (see [this](https://www.reddit.com/r/cpp_questions/comments/7yvu89/stdvector_of_unique_ptr/)).

The codacy reports regarding smart pointers are somewhat inaccurate. :disappointed:

PR #397, opened 2018-05-07
Setup Travis CI to use Docker

jslee02:
***

- [X] Set version target by selecting a milestone on the right side
- [x] Remove AppVeyor because it doesn't support parallel build for open projects
- [x] Choose which build configurations to enable


.ci/install_linux.sh
```diff
@@ -1,6 +1,8 @@
 #!/usr/bin/env bash
 
-set -e
+set -ex
+
+cd "${HOME}/workspace"
 
 if [ ${TRAVIS} ]; then
   AIKIDO_BUILD_DIR="${TRAVIS_BUILD_DIR}"
```
brianhou: Since we're only using Travis, I think we can delete this `if` statement and just use `${TRAVIS_BUILD_DIR}` in line 13.

.ci/env.list
```diff
@@ -0,0 +1,11 @@
+TRAVIS
```
brianhou: Can we move this file to `.ci/docker/env.list`?

Issue #398, opened 2018-05-07
Add `getBodyNodeOrThrow` to `aikido::robot::util`


gilwoolee:
`getBodyNodeOrThrow` seems to be useful enough to be defined in every robot classes, so let's have it in `robot::util`.

https://github.com/personalrobotics/libada/blob/master/src/Ada.cpp#L86

jslee02:
It would be more general if it takes `MetaSkeleton` instead of `Skeleton`. Also, it should either (1) take const reference or (2) take `const ConstMetaSkeleton*` and check the nullity.

Tonkel:
Pull Request #399

brianhou:
Resolved by #399.

PR #399, opened 2018-05-07
Add get body node or throw

Tonkel:
According to issue #398. I copied the method from Ada.cpp in libada into aikido::robot::util and tested it with the feeding demo.


include/aikido/robot/util.hpp
```diff
@@ -255,8 +255,8 @@ Eigen::Isometry3d getLookAtIsometry(
 /// param[in] skeleton MetaSkeleton that should contain the BodyNode
 /// param[in] bodyNodeName name of the BodyNode we are looking for
```
aditya-vk: Nit: `\param[in]`

jslee02:
Needs `make format`

PR #400, opened 2018-05-07
Cancel in-progress trajectories.

brianhou:
There was a comment that suggested that this isn't supported by rewd_controllers yet, but it seems like it [might be](https://github.com/personalrobotics/rewd_controllers/blob/master/src/JointTrajectoryControllerBase.cpp#L210)? We should test this out on the real hardware.

***


src/control/ros/RosTrajectoryExecutor.cpp
```diff
@@ -234,7 +234,11 @@ void RosTrajectoryExecutor::step(
 //==============================================================================
 void RosTrajectoryExecutor::abort()
 {
-  // TODO: cancel the actionlib goal (once there is support in ReWD controller)
+  std::lock_guard<std::mutex> lock(mMutex);
+  DART_UNUSED(lock); // Suppress unused variable warning.
+
+  if (mInProgress)
+    mGoalHandle.cancel();
```
jslee02: Would it be useful to print a warning something like "Attempting to abort trajectory execution but no trajectory in progress."?

brianhou: That sounds like a good idea. We should do this to the other `TrajectoryExecutor::abort` implementations as well.

src/control/ros/RosTrajectoryExecutor.cpp
```diff
@@ -234,7 +234,11 @@ void RosTrajectoryExecutor::step(
 //==============================================================================
 void RosTrajectoryExecutor::abort()
```
gilwoolee: I'd prefer "cancel" instead of "abort". It seems like "cancel" and "abort" are different terms in ros_control; "cancel" is where the client requested to cancel the trajectory, and "abort" is where the server had to stop it for some other reason (e.g. force threshold has been violated).

http://wiki.ros.org/actionlib/DetailedDescription

gilwoolee:
Shouldn't we add the method to `TrajectoryExecutor` and support the same functionality in `KinematicSimulationTrajectoryExecutor`?

brianhou:
I think `KinematicSimulationTrajectoryExecutor` already has this `abort` implemented (although as you said, it should probably be renamed to `cancel`).

My understanding is that we're not using `actionlib` in simulation, so this particular implementation is not helpful there.

gilwoolee:
Oops- sorry I wasn't aware that the base class / `KinematicSimulationTrajectoryExecutor` had this method. I think it's all good as long as we rename it then!

include/aikido/control/InstantaneousTrajectoryExecutor.hpp
```diff
@@ -43,7 +43,7 @@ class InstantaneousTrajectoryExecutor : public TrajectoryExecutor
       const std::chrono::system_clock::time_point& /*timepoint*/) override;
 
   // Do nothing.
```
jslee02: Nit: `/// Does nothing`? Note that three slashes.

include/aikido/control/TrajectoryExecutor.hpp
```diff
@@ -4,6 +4,7 @@
 #include <chrono>
 #include <future>
 #include <set>
+#include <dart/common/Console.hpp>
```
jslee02: Nit: This can be moved to the cpp file.

include/aikido/control/ros/RosTrajectoryExecutor.hpp
```diff
@@ -58,8 +58,8 @@ class RosTrajectoryExecutor : public aikido::control::TrajectoryExecutor
   /// Regularly checks for the completion of a sent trajectory.
   void step(const std::chrono::system_clock::time_point& timepoint) override;
 
-  // Do nothing.
-  void abort() override;
+  // \copydoc TrajectoryExecutor::cancel()
```
jslee02: Nit: `///`

src/control/ros/RosTrajectoryExecutor.cpp
```diff
@@ -232,9 +232,16 @@ void RosTrajectoryExecutor::step(
...
 {
-  // TODO: cancel the actionlib goal (once there is support in ReWD controller)
+  std::lock_guard<std::mutex> lock(mMutex);
+  DART_UNUSED(lock); // Suppress unused variable warning.
+
+  if (mInProgress)
+    mGoalHandle.cancel();
+  else
+    dtwarn << "[RosTrajectoryExecutor::cancel] Attempting to "
+           << "cancel trajectory, but no trajectory in progress.\n";
```
jslee02: Nit: Please wrap multiple lines with curly brackets.

PR #401, opened 2018-05-08
Minor changes to aikido::io::loadSkeletonFromURDF.

brianhou:
- Changed API to take in a DART `Uri` rather than string. (Strings can be automatically converted to `Uri`s.)
- Only try to convert to a `FreeJoint` if a transform has been provided. (We might want to load `Skeleton`s with fixed joints, but wouldn't set a transform in this case.)

***


jslee02:
> Only try to convert to a FreeJoint if a transform has been provided. (We might want to load Skeletons with fixed joints, but wouldn't set a transform in this case.)

How about we print a warning in that case? One might pass non-identity transform for fixed joint expecting transformed root body, which is not the case here.

Tonkel:
If someone passes a non-identity transform for a fixed joint, it throws an error already, doesn't it?

jslee02:
@Tonkel Ah, that's true.

PR #402, opened 2018-05-09
Cleanup build matrix

jslee02:
**Summary**
* Remove unnecessary environment variable (i.e., `CATKIN_CONFIG_OPTIONS`) from macOS builds
* Fix trying to deploy the docs on macOS
* Fix running `codecov` twice
* Introduce `BUILD_NAME` for each build to identify them more conveniently


.ci/script_linux.sh
```diff
@@ -9,10 +9,20 @@ set -e
 ./scripts/internal-run.sh catkin build --no-status --no-deps -p 1 -i --make-args check-format -- aikido
 
 # Manually build Aikido's tests; they are not built automatically because it is not a Catkin package.
-./scripts/internal-run.sh catkin build --no-status --no-deps -p 1 -i --cmake-args -DCMAKE_BUILD_TYPE=$BUILD_TYPE -DTREAT_WARNINGS_AS_ERRORS=ON -DCODECOV=$CODECOV --make-args tests -- aikido
+if [ $BUILD_NAME = TRUSTY_FULL_RELEASE ]; then
```
brianhou: Shouldn't this be `TRUSTY_FULL_DEBUG`? (That's when `CODECOV` used to be `ON`, and I think it might be necessary to run with debugging information?)

.travis.yml
```diff
@@ -19,24 +19,24 @@ matrix:
...
+      env: BUILD_NAME=TRUSTY_FULL_DEBUG BUILD_TYPE=Debug COMPILER=GCC CATKIN_CONFIG_OPTIONS="-DCMAKE_BUILD_TYPE=Debug -DTREAT_WARNINGS_AS_ERRORS=ON" DOWNLOAD_TAGFILES=OFF
     - os: linux
       compiler: gcc
-      env: BUILD_TYPE=Release CODECOV=OFF COMPILER=GCC CATKIN_CONFIG_OPTIONS="-DCMAKE_BUILD_TYPE=Release -DTREAT_WARNINGS_AS_ERRORS=ON" DOWNLOAD_TAGFILES=ON
+      env: BUILD_NAME=TRUSTY_FULL_RELEASE BUILD_TYPE=Release COMPILER=GCC CATKIN_CONFIG_OPTIONS="-DCMAKE_BUILD_TYPE=Release -DTREAT_WARNINGS_AS_ERRORS=ON" DOWNLOAD_TAGFILES=ON
     - os: osx
       osx_image: xcode9.3
       compiler: clang
-      env: BUILD_TYPE=Debug CODECOV=OFF COMPILER=CLANG CATKIN_CONFIG_OPTIONS="-DCMAKE_BUILD_TYPE=Debug -DTREAT_WARNINGS_AS_ERRORS=ON"
+      env: BUILD_NAME=XCODE93_DEBUG BUILD_TYPE=Debug COMPILER=CLANG -DTREAT_WARNINGS_AS_ERRORS=ON"
```
brianhou: There's a trailing `"` here from the definition of `CATKIN_CONFIG_OPTIONS` that used to be here. How should we set `TREAT_WARNINGS_AS_ERRORS` correctly?

.travis.yml
```diff
@@ -19,24 +19,24 @@ matrix:
   include:
     - os: linux
       compiler: gcc
-      env: REQUIRED_ONLY=--required-only BUILD_TYPE=Debug CODECOV=ON COMPILER=GCC CATKIN_CONFIG_OPTIONS="-DCMAKE_BUILD_TYPE=Debug -DTREAT_WARNINGS_AS_ERRORS=ON" DOWNLOAD_TAGFILES=OFF
+      env: BUILD_NAME=TRUSTY_DEBUG REQUIRED_ONLY=--required-only BUILD_TYPE=Debug COMPILER=GCC CATKIN_CONFIG_OPTIONS="-DCMAKE_BUILD_TYPE=Debug -DTREAT_WARNINGS_AS_ERRORS=ON" DOWNLOAD_TAGFILES=OFF
```
brianhou: Rather than setting `TREAT_WARNINGS_AS_ERRORS = ON` in all these matrix entries, can we just make that the default in `CMakeLists.txt`? I think we also don't need a `DOWNLOAD_TAGFILES` environment variable; we can just set that depending on the `BUILD_NAME` like you did above with `CODECOV`.

brianhou: Maybe we can just set `DOWNLOAD_TAGFILES = ON` in `.ci/build_docs.sh` (not as an environment variable), and then only call that script if `BUILD_NAME = TRUSTY_RELEASE`.

.travis.yml
jslee02: This is just an oversight. `TREAT_WARNINGS_AS_ERRORS` should be set in `script_macos.sh`.

.travis.yml
jslee02: > Rather than setting TREAT_WARNINGS_AS_ERRORS = ON in all these matrix entries, can we just make that the default in CMakeLists.txt?

Note that `TREAT_WARNINGS_AS_ERRORS = ON` is part of `CATKIN_CONFIG_OPTIONS` that is used in (`internal-build.sh`](https://github.com/personalrobotics/pr-cleanroom/blob/740e4de50e047f612fb83b763113b5719d71eda1/internal-build.sh#L27).

jslee02: > Maybe we can just set DOWNLOAD_TAGFILES = ON in .ci/build_docs.sh (not as an environment variable), and then only call that script if BUILD_NAME = TRUSTY_FULL_RELEASE.

Hm, how about executing `build_docs.sh` only for `TRUSTY_FULL_RELEASE` and always passing ON for `DOWNLOAD_TAGFILES` in `build_docs.sh`?

brianhou: Great, let's delete `DOWNLOAD_TAGFILES` everywhere in `.travis.yml` then.

PR #403, opened 2018-05-09
Build multiple versions of documentation

### LINKED ISSUE #359: Upload documentation for multiple versions
The current CI setup only supports building documentation from the master branch. Once we have different versions, we should support different versions of the documentation as well.

We're currently piggybacking off the release test build, but it might be helpful to create a separate entry in the build matrix just for building all the documentation versions.

brianhou:
Resolves #359.

***


.ci/docs_versions.txt
```diff
@@ -0,0 +1 @@
+master
```
jslee02: Could we add other versions (v0.2.0 and v0.1.0) for a test? We could remove later if it's not necessary.

brianhou: Yeah, I'll do that once I get this to build correctly for just one version :)

PR #408, opened 2018-05-10
fix geodesic for SO(2)

aditya-vk:
This PR restricts the SO(2) state representation to lie between `+pi` and `-pi`.
1. This ensures planning, for instance with `SnapPlanner`, between two angles gives the geodesic.
2. Is similar to other statespace representations [bijective with transformation matrix representation].

This will possibly require keeping a track of another parameter to know what phase of rotation the joint is in?

Looking for proposals and suggestions on handling this in the cleanest possible way [while keeping in mind the trajectories sent to the robot].

~(Did not modify tests yet.)~

***


src/planner/SnapPlanner.cpp
```diff
@@ -37,8 +37,12 @@ trajectory::InterpolatedPtr planSnap(
     }
   }
 
+  // Redefine the goal state
```
aditya-vk: This shouldn't be necessary.

mkoval:
I don't think this is a good solution to the stated problem.

> This ensures planning, for instance with SnapPlanner, between two angles gives the geodesic.

There is no mapping from SO(2) from R that has this property.

> Is similar to other statespace representations [bijective with transformation matrix representation].

I don't see any similarity here. The whole point of the Aikido `StateSpace` class is to make our algorithms independent of our choice of parameterization. We do this by operating on opaque `State` objects using Lie group operations, rather than operating on raw Eigen vectors.

---

Our current implementation of `GeodesicInterpolator` looks good to me. I don't see why it would behave incorrectly in SO(2). What is the behavior that you are seeing?

gilwoolee:
I don't agree with Aditya's current proposal, but I can discuss the problem we have with the current SO2 implementation. 

I also agree with @mkoval that there's nothing wrong with current `GeodesicInterpolator`; it's just the SO2 space.

Here's the problem with our current SO2 implementation. We keep `double mAngle`  as our internal representation, and for `inverse`, `compose` and `logMap` we simply take `-mAngle`, `mAngle1 + mAngle2` and `mAngle`. If we kept `Eigen::Rotation2d rot` as our representation, the operation would've been `rot^-1`, `rot1 * rot2`, `logMap(rot)`.

Suppose we do geodesic interpolation from SO2(2*pi) to SO2(0). This should be a no-travel. Consider the halfway point. The tangent vector we get for from->to direction in our current operation is `-2pi`, so the half-way point ends up being `pi`.

Here are two tests that fail with the current implementation:
https://github.com/personalrobotics/aikido/blob/bug/SO2/tests/statespace/test_SO2.cpp#L125
https://github.com/personalrobotics/aikido/blob/bug/SO2/tests/statespace/test_SO2.cpp#L150

I am ok with the first one (`getAngle` inequality), but the 2nd one is definitely an issue.

Maybe I'm misunderstanding @mkoval 's comment here:
> There is no mapping from SO(2) from R that has this property.

aditya-vk:
I do not think there is anything wrong with the `GeodesicInterpolator` either. As @gilwoolee mentioned, the problem's with the `SO(2)` statespace. The underlying representation is currently an unbounded scalar value representing the angle of rotation while we expect the angle to wrap-around. The issue is in determining the `tangentVector` which turns out to be `target - source` in the current implementation, without regard to the right direction.

There are two ways to solve this right. 
1. We could clamp the angle to always lie between `-pi` and `pi`, which will lead to loss of some information, but we are guaranteed that it'll be sound with the logic in the geodesic interpolator i.e the direction of interpolation will always be correct. To add to Gilwoo's example, consider the case when my source and target are `pi/6` and `11pi/6` (~ `-pi/6`). The current implementation would interpolate along `11pi/6 - pi/6` while it needs to interpolate the other way. [ This has been my running example :) ]. By clamping it we will have interpolated along `-pi/6 - (pi/6)`.
2. We could use a cleaner rotation matrix representation to do the same thing but I am yet to completely trust Eigen's `Rotation2D.angle()` for that. Eigen 3.3 introduced Rotation2D.getShortestAngle()` which made me wary of the former, hence the quick proposal with the clamp which achieves the same thing.
3. I do not have a deep understanding about the `exp/logMap`, at least not as much, but according to @jslee02 and @gilwoolee, the fix might need to be in `logMap`. I had a brief discussion with @jslee02 about it this evening and he had thoughts I could not wrap my head around at the moment. Maybe he can explain this part?

@gilwoolee thanks for the tests! I should have started with that :)

aditya-vk:
Thanks @jslee02 !
The implementation in this PR currently handles 1 and 4 which ensures the correct behavior. Let me change the function names for more clarity, and add some tests for us to be convinced some corner cases aren't missed. :)

aditya-vk:
Updated for further discussion. :+1:

src/statespace/SO2.cpp
```diff
@@ -73,53 +74,52 @@ std::size_t SO2::getStateSizeInBytes() const
...
-    StateSpace::State* _out) const
+    const StateSpace::State* state1,
+    const StateSpace::State* state2,
+    StateSpace::State* out) const
 {
-  // TODO: Disable this in release mode.
-  if (_state1 == _out || _state2 == _out)
-    throw std::invalid_argument("Output aliases input.");
+#ifndef NDEBUG // Debug mode
+  assert(state1 != out && state2 != out);
```
jslee02: I don't know why we wanted to disable this in release mode, but it seems we need this precondition check (using exception) in both modes according to the [docstring](https://github.com/personalrobotics/aikido/blob/master/include/aikido/statespace/StateSpace.hpp#L92-L93).

src/statespace/SO2.cpp
```diff
@@ -73,53 +74,52 @@ std::size_t SO2::getStateSizeInBytes() const
...
-void SO2::getInverse(
-    const StateSpace::State* _in, StateSpace::State* _out) const
+void SO2::getInverse(const StateSpace::State* in, StateSpace::State* out) const
 {
-  // TODO: Disable this in release mode.
-  if (_out == _in)
-    throw std::invalid_argument("Output aliases input.");
+#ifndef NDEBUG // Debug mode
+  assert(out != in);
+#endif
```
jslee02: Ditto with `SO2::compose()`.

src/statespace/SO2.cpp
```diff
@@ -134,44 +134,37 @@ void SO2::copyState(
...
-  if (_tangent.rows() != 1)
-  {
-    std::stringstream msg;
-    msg << "_tangent has incorrect size: expected 1"
-        << ", got " << _tangent.rows() << ".\n";
-    throw std::runtime_error(msg.str());
-  }
+#ifndef NDEBUG // Debug mode
+  assert(tangent.rows() == 1);
+#endif
```
jslee02: Hm, the original warning message seems useful. Let's keep it.

src/statespace/SO2.cpp
aditya-vk: Good catch :+1:

aditya-vk:
Sure. I addressed the minor comments. I'll add a few more tests for catching invalid arguments and notify by removing the WIP label. Thanks!

src/statespace/SO2.cpp
```diff
@@ -3,37 +3,38 @@
...
 
 //==============================================================================
-void SO2::State::setAngle(double _angle)
+void SO2::State::fromAngle(double angle)
 {
-  mAngle = _angle;
+  double boundedAngle = std::fmod(angle, 2.0 * M_PI);
+  if (boundedAngle > M_PI)
+    boundedAngle -= 2.0 * M_PI;
+  if (boundedAngle < -M_PI)
```
brianhou: It looks like this permits `boundedAngle` to be `-M_PI`, while the documentation says that it is bounded in `(-pi, pi]`.

src/statespace/SO2.cpp
```diff
@@ -73,53 +74,50 @@ std::size_t SO2::getStateSizeInBytes() const
...
 //==============================================================================
-void SO2::getInverse(
-    const StateSpace::State* _in, StateSpace::State* _out) const
+void SO2::getInverse(const StateSpace::State* in, StateSpace::State* out) const
 {
-  // TODO: Disable this in release mode.
-  if (_out == _in)
-    throw std::invalid_argument("Output aliases input.");
+  if (out == in)
+    throw std::invalid_argument("Output aliases input");
```
brianhou: Undo deleted period?

include/aikido/statespace/SO2.hpp
```diff
@@ -71,78 +34,109 @@ class SO2 : virtual public StateSpace
...
+  double toAngle(const State* state) const;
 
-  /// Sets state to a rotation angle.
+  /// Sets state from a rotation angle.
   ///
-  /// \param _state input state
-  /// \param _angle rotation angle
-  void setAngle(State* _state, double _angle) const;
+  /// \param[out] state State corresponding to angle
+  /// \param[in] angle Rotation angle.
```
gilwoolee: Can we specify the range for `angle` here? (If it allows (-inf, inf), mention that)

include/aikido/statespace/SO2.hpp
```diff
@@ -71,78 +34,109 @@ class SO2 : virtual public StateSpace
...
+  explicit State(double angle = 0.0);
+
+  ~State() = default;
+
+  /// Returns state as a rotation angle in (-pi, pi].
+  double toAngle() const;
+
+  /// Sets state from a rotation angle.
+  ///
+  /// \param[in] angle Rotation angle
```
gilwoolee: Same as above, specify the range for `angle`.

src/distance/SO2Angular.cpp
```diff
@@ -25,9 +25,9 @@ double SO2Angular::distance(
     const aikido::statespace::StateSpace::State* _state2) const
 {
   // Difference between angles
-  double diff = mStateSpace->getAngle(
+  double diff = mStateSpace->toAngle(
```
gilwoolee: Nit: Are we correcting the variable names in this PR or not? (`_state`)

tests/statespace/test_SO2.cpp
```diff
@@ -8,38 +8,59 @@ using Eigen::Rotation2Dd;
 
 static constexpr double TOLERANCE{1e-6};
 
+TEST(SO2, BoundedAngle)
+{
+  SO2::State s1(3 * M_PI_2);
```
gilwoolee: I would add tests to check the exact bounds `-pi` and `pi` as well.

src/distance/SO2Angular.cpp
aditya-vk: I was doing it only in the `SO2.cpp/hpp` since I was mostly modifying them for this PR. I think I'll leave that for another PR and just have SO2 stuff cleaned up in code style.

jslee02:
@aditya-vk Please don't forget to update the changelog for completeness.

PR #423, opened 2018-05-15
Ranking IK Solutions

aditya-vk:
This PR introduces ranker for IK solutions.

Looking for discussion on:
1. Should we make it more general to also be able to rank TSRs with the same (templated) classes? 
I prefer having them separate to make it simple and free of potential errors.
  - Decided to keep them separate
2. Where do we have the files? It seemed `constraint/dart/` might be a good place since it's still a constraint over the IKs which by itself is a constraint. Can also be under `planner`.
  - Decided to have them under `distance`
3. Naming convention for the classes.
4. Efficiency and implementation details.

~WIP: Tests. I started a test for FIFO (which is more or less a random IK ranker like in `prpy`).~
***


include/aikido/constraint/dart/IKRankingStrategy.hpp
```diff
@@ -0,0 +1,69 @@
...
+  IKRankingStrategy(
+      statespace::dart::ConstMetaSkeletonStateSpacePtr metaSkeletonStateSpace,
+      ::dart::dynamics::ConstMetaSkeletonPtr metaSkeleton,
+      std::size_t numIKSolutions);
+
+  /// Destructor
+  virtual ~IKRankingStrategy() = default;
+
+  /// Returns the statespace.
+  statespace::ConstStateSpacePtr getStateSpace() const;
```
brianhou: Is this necessary? What do we think we would need the state space for?

include/aikido/constraint/dart/IKRankingStrategy.hpp
```diff
@@ -0,0 +1,69 @@
...
+  // TODO(avk): What happens when the number of IK solutions
+  // is less than the number specified in constructor?
+
+  /// Returns the vector of ranked IK solutions.
+  std::vector<std::pair<statespace::StateSpace::State*, double>>&
+  getRankedIKSolutions();
+
+  /// Add IK solution to Ranker
+  /// Evaluates, creates a pair and stores it in the vector.
+  void addIKSolution(statespace::StateSpace::State* solution);
```
brianhou: I think it might be easier to require that we're given a vector of states, rather than adding one state at a time.

include/aikido/constraint/dart/IKRankingStrategy.hpp
```diff
@@ -0,0 +1,69 @@
...
+  // 2. https://stackoverflow.com/questions/18479295/
+  // 2. member-initializer-does-not-name-a-non-static-data-member-or-base-class
+  /// Constructor
+  ///
+  /// \param[in] metaSkeletonStateSpace Statespace of the skeleton.
+  /// \param[in] metaskeleton Metaskeleton of the robot.
+  /// \param[in] numIKSolution Number of IK solutions to rank.
+  IKRankingStrategy(
+      statespace::dart::ConstMetaSkeletonStateSpacePtr metaSkeletonStateSpace,
+      ::dart::dynamics::ConstMetaSkeletonPtr metaSkeleton,
```
brianhou: What would the `MetaSkeleton` be used for? If it's just to extract the current configuration, maybe this should be `currentState` instead?

include/aikido/constraint/dart/IKRankingStrategy.hpp
brianhou: Oh, is this for the `JointLimitAvoidanceStrategy`?

brianhou:
> Should we make it more general to also be able to rank TSRs with the same (templated) classes? I prefer having them separate to make it simple and free of potential errors.

Separate sounds fine for me. If we implement both and find that they're more similar than we thought, maybe we can merge them then.

> Where do we have the files? It seemed `constraint/dart/` might be a good place since it's still a constraint over the IKs which by itself is a constraint. Can also be under `planner`.

I don't really consider this ranking a constraint. Maybe `aikido::distance`?

include/aikido/constraint/dart/IKRankingStrategy.hpp
aditya-vk: Moreover, since there is no notion of removing IKs from the container, what you suggest is much neater than having this function, true! Good point!

include/aikido/constraint/dart/IKRankingStrategy.hpp
aditya-vk: That was the idea, yeah.

include/aikido/distance/RandomConfigurationRanker.hpp
```diff
@@ -0,0 +1,30 @@
+#ifndef AIKIDO_DISTANCE_RANDOMCONFIGURATIONRANKER_HPP_
+#define AIKIDO_DISTANCE_RANDOMCONFIGURATIONRANKER_HPP_
+
+#include "aikido/distance/ConfigurationRanker.hpp"
+
+namespace aikido {
+namespace distance {
+
+class RandomConfigurationRanker : public ConfigurationRanker
```
aditya-vk: Is this class required, since the configurations are anyway generated in a random order (except maybe when sequential sampleable with start state is used).

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@
   * Fixed bug in StepSequence iterator: [#320](https://github.com/personalrobotics/aikido/pull/320)
   * Cleaned up doxygen errors: [#357](https://github.com/personalrobotics/aikido/pull/357)
 
+* Distance
+
+	* Added methods to rank configurations based on specified metric: [#423](https://github.com/personalrobotics/aikido/pull/423)
```
brianhou: Nit: indent with spaces, not tabs.

include/aikido/distance/ConfigurationRanker.hpp
```diff
@@ -0,0 +1,52 @@
...
+public:
+  /// Constructor
+  ///
+  /// \param[in] metaSkeletonStateSpace Statespace of the skeleton.
+  /// \param[in] metaskeleton Metaskeleton of the robot.
+  /// \param[in] ikSolutions List of IK solutions to be ranked.
+  ConfigurationRanker(
+      statespace::dart::ConstMetaSkeletonStateSpacePtr metaSkeletonStateSpace,
+      ::dart::dynamics::ConstMetaSkeletonPtr metaSkeleton,
+      const std::vector<statespace::StateSpace::State*> ikSolutions);
```
brianhou: I don't think we need to take in the `ikSolutions` here. Maybe we'd want to reuse the ranker with a different set of `ikSolutions`?

include/aikido/distance/ConfigurationRanker.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+  /// Destructor
+  virtual ~ConfigurationRanker() = default;
+
+  /// Returns the statespace.
+  statespace::ConstStateSpacePtr getStateSpace() const;
+
+  /// Returns the vector of ranked IK solutions.
+  std::vector<std::pair<statespace::StateSpace::State*, double>>&
```
brianhou: I think this method might be more useful if it's something like `std::vector<State*>& rank(std::vector<State*>& ikSolutions)`. Perhaps it can take in the vector of states and sort them in place? I don't think it's necessary for an external user to see the resulting scores.

include/aikido/distance/ConfigurationRanker.hpp
```diff
@@ -0,0 +1,52 @@
...
+      statespace::StateSpace::State* solution) const = 0;
+
+  /// Statespace of the skeleton.
+  statespace::dart::ConstMetaSkeletonStateSpacePtr mMetaSkeletonStateSpace;
+
+  /// Metaskeleton of the robot.
+  ::dart::dynamics::ConstMetaSkeletonPtr mMetaSkeleton;
+
+  /// Vector to hold IK solutions and corresponding score.
+  std::vector<std::pair<statespace::StateSpace::State*, double>> mIKSolutions;
```
brianhou: I don't think this is necessary.

include/aikido/distance/RandomConfigurationRanker.hpp
brianhou: I don't think this is necessary, unless we decide that we want to have a "default" ranker (and we don't want that defualt to be something like `NominalConfigurationRanker`).

src/distance/ConfigurationRanker.cpp
```diff
@@ -0,0 +1,64 @@
...
+    throw std::invalid_argument("MetaSkeletonStateSpace is nullptr.");
+
+  if (!mMetaSkeleton)
+    throw std::invalid_argument("MetaSkeleton is nullptr.");
+
+  if (ikSolutions.empty())
+    throw std::invalid_argument("Vector of IK Solutions is empty.");
+
+  mIKSolutions.resize(ikSolutions.size());
+  mIKSolutions.shrink_to_fit();
```
brianhou: I think not having `mIKSolutions` allows us to skip all this initialization code.

src/distance/JointAvoidanceConfigurationRanker.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+  auto lowerLimitState = mMetaSkeletonStateSpace->createState();
+  mMetaSkeletonStateSpace->convertPositionsToState(
+      lowerLimitPosition, lowerLimitState);
+
+  auto upperLimitState = mMetaSkeletonStateSpace->createState();
+  mMetaSkeletonStateSpace->convertPositionsToState(
+      lowerLimitPosition, upperLimitState);
+
+  auto dmetric = createDistanceMetricFor(
```
brianhou: Should this distance metric be a member variable in `ConfigurationRanker` that we can reuse?

src/distance/NominalConfigurationRanker.cpp
```diff
@@ -0,0 +1,36 @@
...
+  : ConfigurationRanker(metaSkeletonStateSpace, metaSkeleton, ikSolutions)
+{
+  // Do nothing
+}
+
+//==============================================================================
+double NominalConfigurationRanker::evaluateIKSolution(
+    statespace::StateSpace::State* solution) const
+{
+  auto currentState = mMetaSkeletonStateSpace->createState();
```
brianhou: Should this just be `mMetaSkeletonStateSpace->getScopedStateFromMetaSkeleton(mMetaSkeleton.get())`?

include/aikido/distance/ConfigurationRanker.hpp
```diff
@@ -0,0 +1,54 @@
...
+      statespace::dart::ConstMetaSkeletonStateSpacePtr metaSkeletonStateSpace,
+      ::dart::dynamics::ConstMetaSkeletonPtr metaSkeleton);
+
+  /// Destructor
+  virtual ~ConfigurationRanker() = default;
+
+  /// Returns the statespace.
+  statespace::ConstStateSpacePtr getStateSpace() const;
+
+  /// Returns the vector of ranked configurations.
```
brianhou: Nit: update this docstring now that we're modifying the vector rather than returning it?

include/aikido/distance/JointAvoidanceConfigurationRanker.hpp
```diff
@@ -0,0 +1,32 @@
...
+  JointAvoidanceConfigurationRanker(
+      statespace::dart::ConstMetaSkeletonStateSpacePtr metaSkeletonStateSpace,
+      ::dart::dynamics::ConstMetaSkeletonPtr metaSkeleton);
+
+protected:
+  /// Returns score as distance from position limits.
+  double evaluateConfiguration(
+      statespace::StateSpace::State* solution) const override;
+
+  std::vector<std::size_t> mUnboundedLowerLimitsIndices;
```
brianhou: Nit: document these member variables?

include/aikido/distance/NominalConfigurationRanker.hpp
```diff
@@ -0,0 +1,34 @@
...
+      statespace::dart::ConstMetaSkeletonStateSpacePtr metaSkeletonStateSpace,
+      ::dart::dynamics::ConstMetaSkeletonPtr metaSkeleton,
+      const statespace::StateSpace::State* nominalConfiguration);
+
+protected:
+  /// Returns score as distance from the Nominal Configuration.
+  double evaluateConfiguration(
+      statespace::StateSpace::State* solution) const override;
+
+  const statespace::StateSpace::State* mNominalConfiguration;
```
brianhou: Nit: document this.

src/distance/ConfigurationRanker.cpp
```diff
@@ -0,0 +1,71 @@
...
+{
+  std::vector<std::pair<statespace::CartesianProduct::State*, double>>
+      scoredConfigurations(configurations.size());
+  for (std::size_t i = 0; i < configurations.size(); ++i)
+  {
+    scoredConfigurations[i].first = configurations[i];
+    scoredConfigurations[i].second = evaluateConfiguration(configurations[i]);
+  }
+
+  struct sortingFunction
```
brianhou: Can we move this out of this method into an anonymous namespace? Also, maybe a more descriptive name would be helpful.

src/distance/ConfigurationRanker.cpp
```diff
@@ -0,0 +1,71 @@
...
+      scoredConfigurations.begin(),
+      scoredConfigurations.end(),
+      sortingFunction());
+
+  configurations.clear();
+  std::transform(
+      scoredConfigurations.begin(),
+      scoredConfigurations.end(),
+      std::back_inserter(configurations),
+      [](const std::pair<statespace::CartesianProduct::State*, int>& item) {
```
brianhou: Is `int` correct? Seems like it should be `double`.

src/distance/JointAvoidanceConfigurationRanker.cpp
```diff
@@ -0,0 +1,56 @@
...
+  auto lowerLimits = mMetaSkeleton->getPositionLowerLimits();
+  auto upperLimits = mMetaSkeleton->getPositionUpperLimits();
+
+  for (auto index : mUnboundedLowerLimitsIndices)
+    lowerLimits[index] = solutionPosition[index];
+
+  for (auto index : mUnboundedUpperLimitsIndices)
+    upperLimits[index] = solutionPosition[index];
+
+  auto lowerLimitsState
```
brianhou: It seems that we're going to be creating the `lowerLimitsState` and `upperLimitsState` each time we call `evaluateConfiguration`. Can the initialization of those states go in the constructor instead? Then we can call `convertPositionsToState` with the updated position vector.

include/aikido/distance/ConfigurationRanker.hpp
```diff
@@ -0,0 +1,54 @@
...
+  void rankConfigurations(
+      std::vector<statespace::CartesianProduct::State*>& configurations);
+
+protected:
+  /// Returns the score of the configuration
```
brianhou: I think we should document that a lower "score" means that it'll be ranked first? Score then seems like a confusing term for it, since we generally want high scores -- maybe "cost"?

include/aikido/statespace/dart/MetaSkeletonStateSpace.hpp
```diff
@@ -171,6 +171,12 @@ class MetaSkeletonStateSpace : public CartesianProduct
   void convertStateToPositions(
       const State* _state, Eigen::VectorXd& _positions) const;
 
+  /// Returns a \c State in this state space that corresponds to
+  /// specified DART \c MetaSkeleton positions.
+  ///
+  /// \param[in] positions DART \c MetaSkeleton positions
+  ScopedState getStateFromPositions(Eigen::VectorXd& positions) const;
```
brianhou: I'm not sure whether this method is useful.

tests/distance/test_JointAvoidanceConfigurationRanker.cpp
```diff
@@ -0,0 +1,127 @@
...
+  states.emplace_back(seedStateOne);
+  states.emplace_back(seedStateTwo);
+  states.emplace_back(seedStateThree);
+
+  JointAvoidanceConfigurationRanker ranker(mStateSpace, mManipulator);
+  ranker.rankConfigurations(states);
+
+  Eigen::VectorXd rankedStateOne(2), rankedStateTwo(2), rankedStateThree(2);
+  mStateSpace->convertStateToPositions(
+      mStateSpace->cloneState(states[0]), rankedStateOne);
```
brianhou: Why are these `cloneState`s necessary?

tests/distance/test_JointAvoidanceConfigurationRanker.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+  JointAvoidanceConfigurationRanker ranker(mStateSpace, mManipulator);
+  DART_UNUSED(ranker);
+}
+
+TEST_F(JointAvoidanceConfigurationRankerTest, OrderTest)
+{
+  Eigen::VectorXd jointPosition(2);
+
+  jointPosition << 0.3, 0.1;
```
brianhou: There's a lot of duplicated code here, can we try to put this into a loop? We're creating the `Eigen::Vector2d` objects below anyway, so maybe we can use them up here.

tests/distance/test_JointAvoidanceConfigurationRanker.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+  Eigen::VectorXd rankedStateOne(2), rankedStateTwo(2), rankedStateThree(2);
+  mStateSpace->convertStateToPositions(
+      mStateSpace->cloneState(states[0]), rankedStateOne);
+  mStateSpace->convertStateToPositions(
+      mStateSpace->cloneState(states[1]), rankedStateTwo);
+  mStateSpace->convertStateToPositions(
+      mStateSpace->cloneState(states[2]), rankedStateThree);
+
+  EXPECT_TRUE(rankedStateOne.isApprox(Eigen::Vector2d(0.3, 0.1)));
```
brianhou: Why wouldn't these be identical? Also, I think we should use `EXPECT_EIGEN_EQUAL` here.

include/aikido/statespace/dart/MetaSkeletonStateSpace.hpp
aditya-vk: I see the following code block repeating a lot:
```
auto state = space->createState();
space->convertPositionsToState(position, state);
```
Figured I'd make a helper function.

aditya-vk:
Cleaned up! Should be good to go (I think). Will follow up with updates to planToTSR methods once this gets merged :)

The tests are failing for reasons that don't have to do with the code itself.

include/aikido/distance/ConfigurationRanker.hpp
```diff
@@ -0,0 +1,56 @@
+#ifndef AIKIDO_DISTANCE_CONFIGURATIONRANKER_HPP_
+#define AIKIDO_DISTANCE_CONFIGURATIONRANKER_HPP_
+
+#include "aikido/distance/CartesianProductWeighted.hpp"
```
brianhou: Nit: is this include necessary?

include/aikido/distance/ConfigurationRanker.hpp
```diff
@@ -0,0 +1,56 @@
...
+
+  /// Destructor
+  virtual ~ConfigurationRanker() = default;
+
+  /// Returns the statespace.
+  statespace::ConstStateSpacePtr getStateSpace() const;
+
+  /// Returns a vector of configurations ranked in increasing order of costs.
+  /// \param[in] configurations Vector of configurations to rank.
+  std::vector<statespace::CartesianProduct::State*> rankConfigurations(
```
brianhou: Let's change this from `CartesianProduct::State` to `MetaSkeletonStateSpace::State`, etc.

include/aikido/distance/ConfigurationRanker.hpp
```diff
@@ -0,0 +1,56 @@
+#ifndef AIKIDO_DISTANCE_CONFIGURATIONRANKER_HPP_
+#define AIKIDO_DISTANCE_CONFIGURATIONRANKER_HPP_
+
+#include "aikido/distance/CartesianProductWeighted.hpp"
+#include "aikido/distance/DistanceMetric.hpp"
+#include "aikido/distance/defaults.hpp"
+#include "aikido/statespace/CartesianProduct.hpp"
```
brianhou: I think this is also unnecessary.

include/aikido/distance/ConfigurationRanker.hpp
```diff
@@ -0,0 +1,56 @@
...
+  /// \param[in] metaskeleton Metaskeleton of the robot.
+  ConfigurationRanker(
+      statespace::dart::ConstMetaSkeletonStateSpacePtr metaSkeletonStateSpace,
+      ::dart::dynamics::ConstMetaSkeletonPtr metaSkeleton);
+
+  /// Destructor
+  virtual ~ConfigurationRanker() = default;
+
+  /// Returns the statespace.
+  statespace::ConstStateSpacePtr getStateSpace() const;
```
brianhou: This method doesn't seem to be used. I'd prefer to remove it if we don't see a use for it, but I think this should return a `ConstMetaSkeletonStateSpacePtr` if we decide to keep it.

include/aikido/distance/NominalConfigurationRanker.hpp
```diff
@@ -0,0 +1,35 @@
...
+namespace distance {
+
+class NominalConfigurationRanker : public ConfigurationRanker
+{
+public:
+  /// Constructor
+  ///
+  /// \param[in] metaSkeletonStateSpace Statespace of the skeleton.
+  /// \param[in] metaskeleton Metaskeleton of the robot.
+  /// \param[in] nominalConfiguration Nominal Configuration. The current
```
brianhou: Nit: `Nominal Configuration` -> `Nominal configuration`.

Can we also provide a `nullptr` default?

include/aikido/statespace/dart/MetaSkeletonStateSpace.hpp
brianhou: In that case, let's add it as a separate/future PR instead so this one is just about the IK rankers.

src/distance/ConfigurationRanker.cpp
```diff
@@ -0,0 +1,81 @@
...
+  , mMetaSkeleton(std::move(metaSkeleton))
+{
+  if (!mMetaSkeletonStateSpace)
+    throw std::invalid_argument("MetaSkeletonStateSpace is nullptr.");
+
+  if (!mMetaSkeleton)
+    throw std::invalid_argument("MetaSkeleton is nullptr.");
+
+  mDistanceMetric = createDistanceMetricFor(
+      std::dynamic_pointer_cast<statespace::CartesianProduct>(
```
brianhou: Are both of these casts necessary? Does `createDistanceMetricFor` not work for `const statespace::CartesianProduct`?

src/distance/ConfigurationRanker.cpp
```diff
@@ -0,0 +1,81 @@
...
+}
+
+//==============================================================================
+statespace::ConstStateSpacePtr ConfigurationRanker::getStateSpace() const
+{
+  return mMetaSkeletonStateSpace;
+}
+
+//==============================================================================
+std::vector<statespace::CartesianProduct::State*>
```
brianhou: It seems that we're returning raw pointers to the `CartesianProduct::State`s, rather than `CartesianProduct::ScopedState`s. What would happen if the input configurations are destructed before the ranking results are accessed?

src/distance/JointAvoidanceConfigurationRanker.cpp
```diff
@@ -0,0 +1,58 @@
...
+namespace distance {
+
+using statespace::dart::ConstMetaSkeletonStateSpacePtr;
+using ::dart::dynamics::ConstMetaSkeletonPtr;
+
+//==============================================================================
+JointAvoidanceConfigurationRanker::JointAvoidanceConfigurationRanker(
+    ConstMetaSkeletonStateSpacePtr metaSkeletonStateSpace,
+    ConstMetaSkeletonPtr metaSkeleton)
+  : ConfigurationRanker(metaSkeletonStateSpace, metaSkeleton)
```
brianhou: Should this `std::move` the `MetaSkeletonStateSpace` and `MetaSkeleton`?

brianhou:
This looks pretty good overall! Thanks for the changes.

@jslee02 would you mind looking over the `State` stuff to make sure we're using them correctly? Thanks!

include/aikido/distance/ConfigurationRanker.hpp
aditya-vk: I'll remove it for now.

src/distance/ConfigurationRanker.cpp
aditya-vk: The const-cast should not be necessary.

aditya-vk:
Addressed the comments except for the `State`-handling. I'll wait for @jslee02 before modifications. I assumed that the way we'd be using the ranker we wouldn't be destroying the input states but that's a possibility in any case that needs taken care of.

brianhou:
I think it's reasonable to do something like `return ranker.rankConfigurations(states);`, which I think would lose the `states`.

jslee02:
I'm not sure if I can review this PR soon (hopefully late this week?). Please go ahead if this is blocked by me. :sweat_smile:

tests/distance/test_JointAvoidanceConfigurationRanker.cpp
```diff
@@ -0,0 +1,115 @@
...
+using aikido::statespace::dart::MetaSkeletonStateSpacePtr;
+using dart::dynamics::Skeleton;
+using dart::dynamics::SkeletonPtr;
+using dart::dynamics::BodyNode;
+using dart::dynamics::BodyNodePtr;
+using dart::dynamics::RevoluteJoint;
+
+static constexpr double EPS = 1e-6;
+
+static BodyNode::Properties create_BodyNodeProperties(const std::string& _name)
```
jslee02: Nit: No more leading underscore in parameter name!

src/distance/NominalConfigurationRanker.cpp
```diff
@@ -0,0 +1,33 @@
...
+{
+  if (!mNominalConfiguration)
+    mNominalConfiguration
+        = mMetaSkeletonStateSpace->getScopedStateFromMetaSkeleton(
+            mMetaSkeleton.get());
+}
+
+//==============================================================================
+double NominalConfigurationRanker::evaluateConfiguration(
+    statespace::dart::MetaSkeletonStateSpace::State* solution) const
```
jslee02: Why `solution` is not const?

src/distance/NominalConfigurationRanker.cpp
```diff
@@ -0,0 +1,33 @@
+#include "aikido/distance/NominalConfigurationRanker.hpp"
+
+namespace aikido {
+namespace distance {
+
+using statespace::dart::ConstMetaSkeletonStateSpacePtr;
+using statespace::CartesianProduct;
```
jslee02: Nit: This type alias isn't used

src/distance/JointAvoidanceConfigurationRanker.cpp
```diff
@@ -0,0 +1,59 @@
...
+      mUnboundedUpperLimitsIndices.emplace_back(i);
+  }
+
+  mLowerLimitsState = mMetaSkeletonStateSpace->createState();
+  mUpperLimitsState = mMetaSkeletonStateSpace->createState();
+}
+
+//==============================================================================
+double JointAvoidanceConfigurationRanker::evaluateConfiguration(
+    statespace::dart::MetaSkeletonStateSpace::State* solution) const
```
jslee02: Maybe const?

src/distance/JointAvoidanceConfigurationRanker.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+//==============================================================================
+JointAvoidanceConfigurationRanker::JointAvoidanceConfigurationRanker(
+    ConstMetaSkeletonStateSpacePtr metaSkeletonStateSpace,
+    ConstMetaSkeletonPtr metaSkeleton)
+  : ConfigurationRanker(
+        std::move(metaSkeletonStateSpace), std::move(metaSkeleton))
+{
+  auto lowerLimits = mMetaSkeleton->getPositionLowerLimits();
+  auto upperLimits = mMetaSkeleton->getPositionUpperLimits();
```
jslee02: Nit: `const auto&` since `getPosition(Lower/Upper)Limits()` returns const reference.

src/distance/JointAvoidanceConfigurationRanker.cpp
```diff
@@ -0,0 +1,59 @@
...
+  {
+    if (lowerLimits[i] == -dart::math::constantsd::inf())
+      mUnboundedLowerLimitsIndices.emplace_back(i);
+
+    if (upperLimits[i] == dart::math::constantsd::inf())
+      mUnboundedUpperLimitsIndices.emplace_back(i);
+  }
+
+  mLowerLimitsState = mMetaSkeletonStateSpace->createState();
+  mUpperLimitsState = mMetaSkeletonStateSpace->createState();
```
jslee02: `mUpperLimitsState` should be scoped state. Otherwise it will be destructed out of this function, and `mUpperLimitsState` will become dangling pointer.

include/aikido/distance/ConfigurationRanker.hpp
```diff
@@ -0,0 +1,54 @@
...
+  /// \param[in] metaSkeletonStateSpace Statespace of the skeleton.
+  /// \param[in] metaskeleton Metaskeleton of the robot.
+  ConfigurationRanker(
+      statespace::dart::ConstMetaSkeletonStateSpacePtr metaSkeletonStateSpace,
+      ::dart::dynamics::ConstMetaSkeletonPtr metaSkeleton);
+
+  /// Destructor
+  virtual ~ConfigurationRanker() = default;
+
+  /// Ranks/Sorts the vector of configurations in increasing order of costs.
```
brianhou: Nit: I think `sorts` is redundant here?

include/aikido/distance/ConfigurationRanker.hpp
```diff
@@ -0,0 +1,54 @@
...
+  /// \param[in] configurations Vector of configurations to rank.
```
brianhou: Nit: I believe the Doxygen convention for this is `\param[in,out]`.

include/aikido/distance/JointAvoidanceConfigurationRanker.hpp
```diff
@@ -0,0 +1,46 @@
...
+      const override;
+
+  /// Vector of indices corresponding to unbounded lower position limits.
+  std::vector<std::size_t> mUnboundedLowerLimitsIndices;
+
+  /// Vector of indices corresponding to unbounded upper position limits.
+  std::vector<std::size_t> mUnboundedUpperLimitsIndices;
+
+  /// (Modified) State corresponding to the lower position limits.
+  /// The positions at the indices in mUnboundedLowerLimitsIndices are modified
```
brianhou: I think we can change the descriptions for both lower/upper to clarify that the states are modified by `evaluateConfiguration` to match the input state before computing the distance from the lower limits.

include/aikido/distance/ConfigurationRanker.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+#include "aikido/distance/DistanceMetric.hpp"
+#include "aikido/distance/defaults.hpp"
+#include "aikido/statespace/dart/MetaSkeletonStateSpace.hpp"
+
+#include <dart/dynamics/dynamics.hpp>
+
+namespace aikido {
+namespace distance {
+
```
brianhou: Nit: document `ConfigurationRanker`.

include/aikido/distance/JointAvoidanceConfigurationRanker.hpp
```diff
@@ -0,0 +1,46 @@
+#ifndef AIKIDO_DISTANCE_JOINTAVOIDANCECONFIGURATIONRANKER_HPP_
+#define AIKIDO_DISTANCE_JOINTAVOIDANCECONFIGURATIONRANKER_HPP_
+
+#include "aikido/distance/ConfigurationRanker.hpp"
+
+namespace aikido {
+namespace distance {
+
```
brianhou: Nit: document `JointAvoidanceConfigurationRanker`.

I think we should mention here that only joints with limits contribute to the distance to the lower/upper limits.

include/aikido/distance/NominalConfigurationRanker.hpp
```diff
@@ -0,0 +1,37 @@
+#ifndef AIKIDO_DISTANCE_NOMINALCONFIGURATIONRANKER_HPP_
+#define AIKIDO_DISTANCE_NOMINALCONFIGURATIONRANKER_HPP_
+
+#include "aikido/distance/ConfigurationRanker.hpp"
+
+namespace aikido {
+namespace distance {
+
```
brianhou: Nit: document `NominalConfigurationRanker`.

include/aikido/distance/NominalConfigurationRanker.hpp
```diff
@@ -0,0 +1,37 @@
...
+class NominalConfigurationRanker : public ConfigurationRanker
+{
+public:
+  /// Constructor
+  ///
+  /// \param[in] metaSkeletonStateSpace Statespace of the skeleton.
+  /// \param[in] metaskeleton Metaskeleton of the robot.
+  /// \param[in] nominalConfiguration Nominal configuration. The current
+  /// configuration is considered if set to nullptr.
```
brianhou: Nit: clarify that it's the current configuration of the `metaSkeleton`.

include/aikido/distance/NominalConfigurationRanker.hpp
```diff
@@ -0,0 +1,37 @@
...
+namespace aikido {
+namespace distance {
+
+class NominalConfigurationRanker : public ConfigurationRanker
+{
+public:
+  /// Constructor
+  ///
+  /// \param[in] metaSkeletonStateSpace Statespace of the skeleton.
+  /// \param[in] metaskeleton Metaskeleton of the robot.
```
brianhou: Nit: `metaskeleton` -> `metaSkeleton`.

include/aikido/distance/JointAvoidanceConfigurationRanker.hpp
```diff
@@ -0,0 +1,46 @@
...
+class JointAvoidanceConfigurationRanker : public ConfigurationRanker
+{
+public:
+  /// Constructor
+  ///
+  /// \param[in] metaSkeletonStateSpace Statespace of the skeleton.
+  /// \param[in] metaskeleton Metaskeleton of the robot.
```
brianhou: Nit: `metaskeleton` -> `metaSkeleton`.

include/aikido/distance/ConfigurationRanker.hpp
```diff
@@ -0,0 +1,54 @@
...
+class ConfigurationRanker
+{
+public:
+  /// Constructor
+  ///
+  /// \param[in] metaSkeletonStateSpace Statespace of the skeleton.
+  /// \param[in] metaskeleton Metaskeleton of the robot.
```
brianhou: Nit: `metaskeleton` -> `metaSkeleton`.

src/distance/JointAvoidanceConfigurationRanker.cpp
```diff
@@ -0,0 +1,58 @@
...
+    ConstMetaSkeletonPtr metaSkeleton)
+  : ConfigurationRanker(
+        std::move(metaSkeletonStateSpace), std::move(metaSkeleton))
+  , mLowerLimitsState(mMetaSkeletonStateSpace->createState())
+  , mUpperLimitsState(mMetaSkeletonStateSpace->createState())
+{
+  auto lowerLimits = mMetaSkeleton->getPositionLowerLimits();
+  auto upperLimits = mMetaSkeleton->getPositionUpperLimits();
+
+  for (std::size_t i = 0; i < mMetaSkeletonStateSpace->getDimension(); ++i)
```
brianhou: Quick question: is there ever a case where there are lower limits but not upper limits (or vice versa)? If not, then rather than checking against `inf`s, maybe we can use `DegreeOfFreedom::hasPositionLimit`. I think it'd look something like this:
```cpp
auto properties = mMetaSkeletonStateSpace->getProperties();
for (std::size_t i = 0; i < properties.getNumDofs(); ++i)
{
  if (mMetaSkeleton->getDof(i)->hasPositionLimit())
    mUnboundedIndices.emplace_back(i);
}
```

jslee02:
Hm, for some reason, [`test_NominalConfigurationRanker` fails on macOS](https://travis-ci.org/personalrobotics/aikido/jobs/404681752#L6011).

PR #425, opened 2018-05-16
[WIP] Planning adapters

brianhou:
This is a very rough sketch to adapt planners that solve one type of problem to solve other types. The main class is `PlannerAdapter<DelegatePlanner>`.

To use a `ConfigurationToConfigurationPlanner` to solve a `ConfigurationToTSR` problem, we would need to implement `PlannerAdapter<ConfigurationToConfigurationPlanner>::plan(ConfigurationToTSR problem)`.

***


brianhou:
That sounds good to me, but I think it'd also be useful to have a `CompositePlannerAdapter` that allows us to automatically upgrade a `SingleProblemPlanner` to solve all the `Problem`s that we define adapters for. My thought was that when we create a `MetaPlanner` for a robot like HERB, we'd want to make it a sequence of `CompositePlannerAdapter`s rather than manually calling all the adapters in HERB's constructor.

Here's an extension to your proposal:
- Concrete planners (e.g. `SnapConfigurationToConfigurationPlanner`) remain as they are.
- `PlannerAdapter`s are implemented as you suggest to solve only one new `Problem`.
- We introduce a `CompositePlannerAdapter<DelegatePlanner>` that's similar to this current implementation, combining the `PlannerAdapter<ConfigurationToConfigurationPlanner, *>`s that we've defined and delegating to the appropriate [adapted] planner.
- We define `SnapPlanner`, which is a `CompositePlannerAdapter<ConfigurationToConfigurationPlanner>` and can now solve `ConfigurationToConfiguration`, `ConfigurationToConfigurations`, `ConfigurationToTSR`, etc.
- HERB's `SequenceMetaPlanner` contains `SnapPlanner`.

brianhou:
Closing this PR in favor of #437.

PR #426, opened 2018-05-17
Refactor VFP with new Planner API

sniyaz:
This PR has changes that use the new Planner API to refactor VFP.

Right now I've only refactored `planToEndEffectorOffset` by creating the new `VectorFieldConfigurationToEndEffectorOffsetPlanner` class. While these changes are looked over, I'll refactor `planToEndEffectorPose` as well.

**Summary of changes**

1) Create new `ConfigurationToEndEffectorOffsetPlanner` class for the `ConfigurationToEndEffectorOffset`  Problem.

2) Create `VectorFieldConfigurationToEndEffectorOffsetPlanner` class that extends the above and calls the old `planToEndEffectorOffset` method in `VectorFieldPlanner`.

**TODOs**

1) We need to figure out to handle the `startState` provided by `ConfigurationToEndEffectorOffset`. The old VFP code gets the current configuration from the metaSkeleton, so right now I just ignore this instance variable.

2) The old VFP code took `minDistance` and `maxDistance` for `planToEndEffectorOffset`. I provide these by taking a tolerance around the `signedDistance` provided by `ConfigurationToEndEffectorOffset`, but we should discuss whether this is appropriate (I also need to handle the negative case).

***

### DIFF
--- a/include/aikido/planner/ConfigurationToEndEffectorOffsetPlanner.hpp
+++ b/include/aikido/planner/ConfigurationToEndEffectorOffsetPlanner.hpp
@@ -1,0 +1,42 @@
+#ifndef AIKIDO_PLANNER_CONFIGURATIONTOENDEFFECTOROFFSETPLANNER_HPP_
+#define AIKIDO_PLANNER_CONFIGURATIONTOENDEFFECTOROFFSETPLANNER_HPP_
+
+#include "aikido/planner/ConfigurationToEndEffectorOffset.hpp"
+#include "aikido/planner/SingleProblemPlanner.hpp"
+#include "aikido/trajectory/Trajectory.hpp"
+
+namespace aikido {
+namespace planner {
+
+/// Base planner class for ConfigurationToEndEffectorOffset planning problem.
+class ConfigurationToEndEffectorOffsetPlanner
+    : public SingleProblemPlanner<ConfigurationToEndEffectorOffsetPlanner,
+                                  ConfigurationToEndEffectorOffset>
+{
+public:
+  // Expose the implementation of Planner::plan(const Problem&, Result*) in
+  // SingleProblemPlanner. Note that plan() of the base class takes Problem
+  // while the virtual function defined in this class takes SolverbleProblem,
+  // which is simply ConfigurationToEndEffectorOffset.
+  using SingleProblemPlanner::plan;
+
+  /// Constructor
+  ///
+  /// \param[in] stateSpace State space that this planner associated with.
+  explicit ConfigurationToEndEffectorOffsetPlanner(
+      statespace::ConstStateSpacePtr stateSpace);
+
+  /// Solves \c problem returning the result to \c result.
+  ///
+  /// \param[in] problem Planning problem to be solved by the planner.
+  /// \param[out] result Result of planning procedure.
+  virtual trajectory::TrajectoryPtr plan(
+      const SolvableProblem& problem, Result* result = nullptr)
+      = 0;
+  // Note: SolvableProblem is defined in SingleProblemPlanner.
+};
+
+} // namespace planner
+} // namespace aikido
+
+#endif // AIKIDO_PLANNER_CONFIGURATIONTOENDEFFECTOROFFSETPLANNER_HPP_
--- a/include/aikido/planner/ConfigurationToEndEffectorOffsetPlanner.hpp
+++ b/include/aikido/planner/ConfigurationToEndEffectorOffsetPlanner.hpp
@@ -39,6 +39,11 @@ public:
   /// Return ConstMetaSkeletonStateSpacePtr by performing a static cast on
   /// mStateSpace.
   statespace::dart::ConstMetaSkeletonStateSpacePtr getMetaSkeletonStateSpace();
+
+protected:
+  /// Stores stateSpace pointer as aConstMetaSkeletonStateSpacePtr. Prevents
+  /// use of an expensive dynamic cast.
+  statespace::dart::ConstMetaSkeletonStateSpacePtr mMetaSkeletonStateSpace;
 };
 
 } // namespace planner
--- a/include/aikido/planner/ConfigurationToEndEffectorOffsetPlanner.hpp
+++ b/include/aikido/planner/ConfigurationToEndEffectorOffsetPlanner.hpp
@@ -3,6 +3,7 @@
 
 #include "aikido/planner/ConfigurationToEndEffectorOffset.hpp"
 #include "aikido/planner/SingleProblemPlanner.hpp"
+#include "aikido/statespace/dart/MetaSkeletonStateSpace.hpp"
 #include "aikido/trajectory/Trajectory.hpp"
 
 namespace aikido {
@@ -24,7 +25,7 @@ public:
   ///
   /// \param[in] stateSpace State space that this planner associated with.
   explicit ConfigurationToEndEffectorOffsetPlanner(
-      statespace::ConstStateSpacePtr stateSpace);
+      statespace::dart::ConstMetaSkeletonStateSpacePtr stateSpace);
 
   /// Solves \c problem returning the result to \c result.
   ///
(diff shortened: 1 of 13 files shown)


sniyaz:
This is blocking on https://github.com/personalrobotics/aikido/pull/429 right now. Once that gets merged, I'll rebase the actual Planner API changes from this PR on top of that one, and we should be good! 😄

src/planner/vectorfield/VectorFieldConfigurationToEndEffectorOffsetPlanner.cpp
```diff
@@ -0,0 +1,68 @@
...
+  // TODO (sniyaz): Check equality between state space of this planner and given
+  // problem.
+
+  using aikido::planner::vectorfield::planToEndEffectorOffset;
+  using aikido::statespace::dart::MetaSkeletonStateSpace;
+
+  auto metaskeletonStateSpace
+      = std::dynamic_pointer_cast<const MetaSkeletonStateSpace>(mStateSpace);
+  if (!metaskeletonStateSpace)
+    throw std::invalid_argument("mStateSpace is not MetaSkeletonStateSpace!");
```
dqyi11: It is better to check this in the constructor

include/aikido/planner/vectorfield/VectorFieldConfigurationToEndEffectorOffsetPlanner.hpp
```diff
@@ -0,0 +1,94 @@
+#ifndef AIKIDO_PLANNER_VECTORFIELD_VECTORFIELDCONFIGURATIONTOENDEFFECTOROFFSETPLANNER_HPP_
+#define AIKIDO_PLANNER_VECTORFIELD_VECTORFIELDCONFIGURATIONTOENDEFFECTOROFFSETPLANNER_HPP_
+
+#include "aikido/planner/ConfigurationToEndEffectorOffset.hpp"
+#include "aikido/planner/ConfigurationToEndEffectorOffsetPlanner.hpp"
+#include "aikido/planner/vectorfield/VectorFieldPlanner.hpp"
```
jslee02: Nit: It's not necessary to include this header here. Please move to the source.

include/aikido/planner/vectorfield/VectorFieldConfigurationToEndEffectorOffsetPlanner.hpp
```diff
@@ -0,0 +1,94 @@
...
+  /// The planner returns success if the resulting trajectory satisfies
+  /// constraint at some resolution and failure (returning \c nullptr)
+  /// otherwise. The reason for the failure is stored in the \c result output
+  /// parameter.
+  ///
+  /// \param[in] problem Planning problem.
+  /// \param[out] result Information about success or failure.
+  /// \return Trajectory or \c nullptr if planning failed.
+  /// \throw If \c problem is not ConfigurationToEndEffectorOffset.
+  /// \throw If \c result is not ConfigurationToEndEffectorOffset::Result.
```
dqyi11: It seems that these two validations are not yet added in the implementation.

include/aikido/planner/ConfigurationToEndEffectorOffsetPlanner.hpp
```diff
@@ -0,0 +1,42 @@
...
+  // SingleProblemPlanner. Note that plan() of the base class takes Problem
+  // while the virtual function defined in this class takes SolverbleProblem,
+  // which is simply ConfigurationToEndEffectorOffset.
+  using SingleProblemPlanner::plan;
+
+  /// Constructor
+  ///
+  /// \param[in] stateSpace State space that this planner associated with.
+  explicit ConfigurationToEndEffectorOffsetPlanner(
+      statespace::ConstStateSpacePtr stateSpace);
```
jslee02: This planner only works for `MetaSkeletonStateSpace`. Please change this class to:
* take `ConstMetaSkeletonStateSpace` in the constructor
* have a getter `getMetaSkeletonStateSpace()` that returns the `ConstStatespacePtr` by static-casting it to `ConstMetaSkeletonStateSpacePtr`.
to prevent passing incorrect state space in compile time.

include/aikido/planner/vectorfield/VectorFieldConfigurationToEndEffectorOffsetPlanner.hpp
```diff
@@ -0,0 +1,94 @@
...
+  /// \param[in] angularTolerance How a planned trajectory is allowed to deviate
+  /// from a given direction.
+  /// \param[in] initialStepSize Initial step size.
+  /// \param[in] jointLimitTolerance If less then this distance to joint
+  /// limit, velocity is bounded in that direction to 0.
+  /// \param[in] constraintCheckResolution Resolution used in constraint
+  /// checking.
+  /// \param[in] timelimit timeout in seconds.
+  VectorFieldConfigurationToEndEffectorOffsetPlanner(
+      statespace::ConstStateSpacePtr stateSpace,
```
jslee02: Please change to take `ConstMetaSkeletonStateSpacePtr` for the same reason above.

include/aikido/planner/vectorfield/VectorFieldConfigurationToEndEffectorOffsetPlanner.hpp
```diff
@@ -0,0 +1,94 @@
...
+      dart::dynamics::MetaSkeletonPtr metaskeleton,
+      double distanceTolerance,
+      double positionTolerance,
+      double angularTolerance,
+      double initialStepSize,
+      double jointLimitTolerance,
+      double constraintCheckResolution,
+      std::chrono::duration<double> timelimit);
```
jslee02: Parameters that are not planner specific parameters should move to the problem class. @dqyi11 Could you verify all the parameters are planner specific?

src/planner/vectorfield/VectorFieldConfigurationToEndEffectorOffsetPlanner.cpp
sniyaz: For sure! My thinking is that this would require two casts: one in the constructor to check, and one when it's actually used. Are we OK with this?

sniyaz: NVM, this is moot with the discussion below since they type is enforced by the constructor argument now.

sniyaz:
I think the start state should be the last thing to figure out now. 

@jslee02 I had to use a `dynamic_pointer_cast` to go from `ConstStatespacePtr` ---> `ConstMetaSkeletonStateSpace`. Apparently you can't use `static_cast` to downcast when virtual inheritance is involved? There's **definitely** a chance I'm wrong on this, so LMK.

include/aikido/planner/vectorfield/VectorFieldConfigurationToEndEffectorOffsetPlanner.hpp
dqyi11: I agree with the current scope of parameters defined in the problem class, which is consistent  with the design in `prpy`.  https://github.com/personalrobotics/prpy/blob/master/tests/planning/methods/PlanToEndEffectorOffset.py
All these parameters are planner-specific. Though some of them are used in both `vector field planner` and `CRRT planner`, they can be set very differently (for example `constraintCheckResolution`). `distanceTolerance`, `positionTolerance` and  `angularTolerance` are also not strictly aligned in planning end-effector offset between using vector field planner and CRRT planner.

src/planner/vectorfield/VectorFieldConfigurationToEndEffectorOffsetPlanner.cpp
```diff
@@ -0,0 +1,68 @@
...
+  // Just call the core VFP function.
+  // TODO: How should start state be handled?
+  return planToEndEffectorOffset(
+      metaskeletonStateSpace,
+      mMetaskeleton,
+      problem.getEndEffectorBodyNode(),
+      problem.getConstraint(),
+      problem.getDirection(),
+      // TODO: Need to handle case when negative.
+      problem.getDistance() - mDistanceTolerance,
```
dqyi11: It is better to check whether that `problem.getDistance() - mDistanceTolerance` and `problem.getDistance() + mDistanceTolerance` have the same sign...

sniyaz:
I've handled the `startState` issue by setting the `metaSkeleton` to the `startState` given in the `ConfigurationToEndEffectorOffset` problem class. I've also made the `stateSpace` and `startState` arguments to that class constructor more restrictive, since `ConfigurationToEndEffectorOffset` problems can only be used with `MetaSkeletonStateSpace`.

PR #432, opened 2018-05-22
Move MetaSkeleton-dependent problems to aikido::planner::dart.

brianhou:
Some problem types are associated with a `MetaSkeleton` (`ConfigurationToTSR`, `ConfigurationToEndEffector[Offset|Pose]`). However, the `MetaSkeleton` that we use to solve those problems can't be stored in the problem itself because we currently take `Problem`s as `const`.

@jslee02 and I decided that it would make sense for `Planner`s that solve these problems to take in a non-const `MetaSkeleton` in the constructor. This PR introduces the `aikido::planner::dart` namespace for these problem types.

In a future PR, we should add corresponding `SingleProblemPlanner`s for these problems. It would also be useful to have an `aikido::planner::dart::ConfigurationToConfigurationPlanner` that takes in a non-const `MetaSkeleton` in its constructor so that we can adapt them more easily.

***


src/planner/vectorfield/MoveEndEffectorOffsetVectorField.cpp
```diff
@@ -50,7 +50,7 @@ MoveEndEffectorOffsetVectorField::MoveEndEffectorOffsetVectorField(
 bool MoveEndEffectorOffsetVectorField::evaluateCartesianVelocity(
     const Eigen::Isometry3d& pose, Eigen::Vector6d& cartesianVelocity) const
 {
-  using dart::math::logMap;
+  using ::dart::math::logMap;
```
sniyaz: Not really a review thing but more a question from somebody who doesn't know C++ as well- what is the purpose of adding `::` here?

sniyaz:
This personally looks good to me 👍

src/planner/vectorfield/MoveEndEffectorOffsetVectorField.cpp
jslee02: The leading `::` is used to represent the following namespace is a top-level namespace. We use this when there is the same name namespace in the "active" namespace. In our case, we use it to distinguish `::dart` (i.e., the DART's top-level namespace) and `::aikido::statespace::dart`/`::aikido::planner::dart`.

src/planner/dart/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
+set(sources
```
jslee02: I believe we only make a new sub-component when it requires optional 3rd-party dependencies. For example, other dart subdirectories of `statespace` and `constraint` are not components but included its parent components.

In this sense, I think `aikido_planner_ompl` and `aikido_planner_vectorfield` also shouldn't be sub-components.

PR #437, opened 2018-05-24
Planning adapters

brianhou:
Re-implementation of #425.

Main changes:
- This PR adds a `MetaSkeleton` argument to the constructor of all "DART planners" in `planner::dart`. All DART planners now also define `getMetaSkeletonStateSpace` and `getMetaSkeleton` to make adapters possible. These methods are currently re-implemented in each planner, but they could (should?) be moved into a `dart::SingleProblemPlanner` class that all DART planners inherit from.
- This PR introduces the `dart::DartPlannerAdapter` class for converting planners that don't contain a `MetaSkeleton` (i.e. `SnapConfigurationToConfigurationPlanner`) into their corresponding DART planner (in this case, `dart::ConfigurationToConfigurationPlanner`). This correspondence isn't enforced anywhere, but maybe some fancy template code could do so?
- The `dart::PlannerAdapter` class is essentially the same single-problem adapter from #425, but only operates on DART planners that contain a `MetaSkeleton`. We can't adapt a planner that doesn't contain a `MetaSkeleton` into one that should.
- Preliminary implementation of the adapter from `ConfigurationToConfiguration` to `ConfigurationToTSR`. It does a few `const_cast`s, which we should look hard at to make sure that they're absolutely necessary.

The pipeline to use a `SnapConfigurationToConfigurationPlanner` as a `dart::ConfigurationToTSRPlanner` now takes an extra step:
```
auto snapDelegate = std::make_shared<SnapConfigurationToConfigurationPlanner>(metaSkeletonStateSpace, interpolator);
auto delegate = std::make_shared<DartPlannerAdapter<SnapConfigurationToConfigurationPlanner, dart::ConfigurationToConfigurationPlanner>>(snapDelegate, metaSkeleton);
auto planner = std::make_shared<ConfigurationToConfiguration_to_ConfigurationToTSR>(delegate);
```
The new intermediate step is necessary to adapt the non-DART `SnapConfigurationToConfigurationPlanner` into a DART `ConfigurationToConfigurationPlanner`.

Issues raised during this PR:
- Should all `Planner`s have an `RNG`? This is needed to create a bounds sampler for the `IKSampleable`, so it might be convenient to include this with all `Planner`s in case an adapter needs it.

***


src/planner/dart/ConfigurationToConfiguration_to_ConfigurationToTSR.cpp
```diff
@@ -0,0 +1,111 @@
...
+  auto delegateProblem = ConfigurationToConfiguration(
+      mMetaSkeletonStateSpace,
+      problem.getStartState(),
+      goalState,
+      problem.getConstraint());
+
+  while (generator->canSample())
+  {
+    // Sample from TSR
+    {
```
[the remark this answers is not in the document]
brianhou: Adding the braces creates a temporary scope, so all of the local variables in that scope will be destructed when it ends. A [`std::lock_guard`](https://en.cppreference.com/w/cpp/thread/lock_guard) releases the mutex when it's destructed.

jslee02:
> All DART planners now also define getMetaSkeletonStateSpace and getMetaSkeleton to make adapters possible. These methods are currently re-implemented in each planner, but they could (should?) be moved into a dart::SingleProblemPlanner class that all DART planners inherit from.

:+1: 

> This correspondence isn't enforced anywhere, but maybe some fancy template code could do so?

This sounds good to me. We could save this for furture though. In that case, it'd be good to create an issue to track this.

src/planner/dart/ConfigurationToConfiguration_to_ConfigurationToTSR.cpp
```diff
@@ -0,0 +1,111 @@
...
+    const ConfigurationToTSR& problem, Planner::Result* result)
+{
+  // TODO: Check equality between state space of this planner and given problem.
+
+  // TODO: DART may be updated to check for single skeleton
+  if (mMetaSkeleton->getNumDofs() == 0)
+    throw std::invalid_argument("MetaSkeleton has 0 degrees of freedom.");
+
+  auto skeleton = mMetaSkeleton->getDof(0)->getSkeleton();
+  for (size_t i = 1; i < mMetaSkeleton->getNumDofs(); ++i)
```
jslee02: Nit: `std::size_t`

src/planner/dart/ConfigurationToConfiguration_to_ConfigurationToTSR.cpp
```diff
@@ -0,0 +1,111 @@
...
+    if (mMetaSkeleton->getDof(i)->getSkeleton() != skeleton)
+      throw std::invalid_argument("MetaSkeleton has more than 1 skeleton.");
+  }
+
+  // Create an IK solver with MetaSkeleton DOFs
+  ::dart::dynamics::BodyNodePtr endEffectorBodyNode;
+  for (std::size_t i = 0; i < mMetaSkeleton->getNumBodyNodes(); ++i)
+  {
+    if (mMetaSkeleton->getBodyNode(i)->getName()
+        == problem.getEndEffectorBodyNode()->getName())
```
jslee02: I think the equality check should be done over the pointer instead of their names. A `MetaSkeleton` could have multiple body nodes of the same name.

src/planner/dart/ConfigurationToConfiguration_to_ConfigurationToTSR.cpp
```diff
@@ -0,0 +1,111 @@
...
+        "End-effector BodyNode not found in Planner's MetaSkeleton.");
+
+  auto ik = InverseKinematics::create(endEffectorBodyNode);
+  ik->setDofs(mMetaSkeleton->getDofs());
+
+  // Convert TSR constraint into IK constraint
+  InverseKinematicsSampleable ikSampleable(
+      mMetaSkeletonStateSpace,
+      mMetaSkeleton,
+      std::const_pointer_cast<TSR>(problem.getGoalTSR()),
```
jslee02: Nit: Could you make a note here saying this const-casting should be removed once `InverseKinematicsSampleable` is changed to take const constraints?

include/aikido/planner/dart/detail/DartPlannerAdapter-impl.hpp
```diff
@@ -0,0 +1,35 @@
+#include "aikido/planner/dart/DartPlannerAdapter.hpp"
+
+using aikido::statespace::dart::MetaSkeletonStateSpace;
```
jslee02: Technically, `-impl.hpp` is also a header file so using `using namespace ~` can pollute the namespaces. Please remove this.

include/aikido/planner/dart/DartPlannerAdapter.hpp
```diff
@@ -0,0 +1,51 @@
...
+
+#include "aikido/planner/dart/ConfigurationToConfigurationPlanner.hpp"
+#include "aikido/planner/dart/ConfigurationToTSRPlanner.hpp"
+#include "aikido/planner/dart/PlannerAdapter.hpp"
+
+namespace aikido {
+namespace planner {
+namespace dart {
+
+/// Adapts a non-DART DelegatePlanner to solve the single problem that
```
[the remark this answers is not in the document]
jslee02: I have also the similar impression, but couldn't come up with a better name. :sweat:

src/planner/dart/ConfigurationToConfiguration_to_ConfigurationToTSR.cpp
brianhou: My thought was that `BodyNode` that the `ConfigurationToTSR` problem was constructed with could be from a different `Skeleton` than the one that the delegate planner's `MetaSkeleton` refers to. This might happen if we try multiple planners in parallel and therefore have several `Skeleton` copies that are being used. In this case, I think the `BodyNode` pointers would not be the same, but the names would be.

include/aikido/planner/dart/DartPlannerAdapter.hpp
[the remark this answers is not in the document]
brianhou: `Planner_to_DartPlanner`?

src/planner/dart/ConfigurationToConfiguration_to_ConfigurationToTSR.cpp
jslee02: That makes sense.

`MetaSkeleton` has a [function](https://github.com/dartsim/dart/blob/master/dart/dynamics/MetaSkeleton.hpp#L126) that returns all the `BodyNode`s that has the queried name. How about we use that function and prints a warning if it's more than one?

src/planner/dart/ConfigurationToConfiguration_to_ConfigurationToTSR.cpp
```diff
@@ -0,0 +1,108 @@
...
+  for (std::size_t i = 1; i < mMetaSkeleton->getNumDofs(); ++i)
+  {
+    if (mMetaSkeleton->getDof(i)->getSkeleton() != skeleton)
+      throw std::invalid_argument("MetaSkeleton has more than 1 skeleton.");
+  }
+
+  // Create an IK solver with MetaSkeleton DOFs
+  auto matchingNodes = mMetaSkeleton->getBodyNodes(
+      problem.getEndEffectorBodyNode()->getName());
+  if (matchingNodes.size() == 0)
```
jslee02: Nit: Checking emptiness is generally faster or equal to comparing the size. So let's change this to `if (!matchingNodes.empty())`.

PR #424, opened 2018-05-16
Change planning problem classes to hold ScopedState instead of State*

jslee02:
Otherwise, states in a `Problem` can become dangling pointers. This PR changes the planning problems to clone from given state space and state and store the clones, which are `ScopedState`s.

Depends on #419 and #422.

***

**Before creating a pull request**

- [x] Document new methods and classes
- [x] Format code with `make format`

**Before merging a pull request**

- [x] Set version target by selecting a milestone on the right side
- [ ] Summarize this change in `CHANGELOG.md`
- [x] Add unit test(s) for this change

PR #442, opened 2018-06-06
Fix issues discovered during demo.

sniyaz:
This PR adresses issues related to the new Planner. These issues were discovered while trying to use the new code with the soda demo.

**Summary of Changes**

1) Change `Problem` derivative classes relative to the demo to use `ScopedState` internally to prevent dangling pointers.

2) Add a `ConfigurationToConfigurationPlanner` in the DART planner space that takes a MetaSkeleton.

3) Add `ConfigurationToConfiguration_to_ConfigurationToConfiguration` adapter that turns the standard `ConfigurationToConfiguration` planner into the DART one.

4) Remove start states from DART `Problem` classes. This is because the start state should be taken form the MetaSkeleton, and it's why 2 and 3 are necessary.

5) Make `direction` passed to `ConfigurationToEndEffectorOffset` constructor a `boost::optional`. If it's empty, `getDirection` returns the current direction of the end effector body node (needed to make planning straight with VFP possible in the demo).

6) Add `getTrajectoryPostProcessor` method to `ConcreteRobot` class. We need this in libherb.

7) `ConfigurationToConfiguration_to_ConfigurationToTSR` has been fixed by taking an RNG in the constructor and using this when creating the sampler. The plan() method now checks whether the problem's bodynode is in the skeleton the MetaSkeleton corresponds to (in cases where the hand is not in the MetaSkeleton itself, such as with HERB).

***


brianhou:
I think the issues with 4 and 5 arise because we're trying to plan with MAGI's `SequenceAction`. Since we construct these planning problems before solving any of them, the `MetaSkeleton` will still be at the initial configuration for _all_ of these planning problems (crucially, not the final configuration of the solution to the previous problem). The `MetaSkeleton` is only in the right place when planning begins, not when the planning problem is constructed.

I believe `getDirection` is called at the beginning of planning, so it does the correct thing for 5. So I think adding additional constructors that take `MetaSkeleton`s might work, as long as `getStartState` is changed accordingly to either return the start state that was passed in the constructor _or_ get the current state of the `MetaSkeleton`.

sniyaz:
Could you all take a look when you have a chance (resurrecting this post-demo, since ideally it would get merged before something else comes along that could break it).

I've changed some things from the original PR based on other problems that came up from the demo.

1) Made all DART planning problems use multiple constructors. One of these takes a start state explicitly, while the other takes a MetaSkeleton and gets the start state from its current configuration when `getStartState` is called.

2) A similar thing is done with `ConfigurationToEndEffectorOffset`. There are constructors that take a direction, and constructors that don't. If a constructor is used that doesn't take a direction, `getDirection` returns the current direction of the end effector BodyNode.

These changes were made to address planning problems that consist of a sequence of subproblems (i.e. when using MAGI). Here, the start state of a problem later in the sequence is not known a priori, such as when doing `ConfigurationToTSR` and then `ConfigurationToConfiguration`. As such, it can't be set when the planning problem is constructed.

3) A version of the `ConfigurationToConfiguration` problem was made in the DART planning namespace with the feature from item 1 above (this is used with MAGI in the demo). And adapter is also included to convert a non-DART `ConfigurationToConfiguration` planner into the DART version.

I still need to make the tests build, but want to make sure these changes are OK first.

src/planner/dart/ConfigurationToEndEffectorOffset.cpp
```diff
@@ -48,13 +106,30 @@ ConfigurationToEndEffectorOffset::getEndEffectorBodyNode() const
...
 }
 
 //==============================================================================
-const Eigen::Vector3d& ConfigurationToEndEffectorOffset::getDirection() const
+const Eigen::Vector3d ConfigurationToEndEffectorOffset::getDirection() const
 {
-  return mDirection;
+  if (!mDirection)
+  {
+    return mEndEffectorBodyNode->getWorldTransform()
```
brianhou: Seems like this is copied from `ConcreteManipulator::getEndEffectorDirection`, maybe we can try to reuse that instead?

sniyaz:
When I last spoke to @jslee02 about https://github.com/personalrobotics/aikido/pull/424, it seemed that PR was being abandoned and he wanted to redo it. I've started some of that here- I'd be happy to do more once this PR gets merged.

include/aikido/planner/dart/ConfigurationToConfiguration_to_ConfigurationToTSR.hpp
```diff
@@ -10,16 +10,22 @@ namespace planner {
...
                             ConfigurationToTSRPlanner>
 {
 public:
   ConfigurationToConfiguration_to_ConfigurationToTSR(
-      std::shared_ptr<ConfigurationToConfigurationPlanner> planner,
-      ::dart::dynamics::MetaSkeletonPtr metaSkeleton);
+      std::shared_ptr<aikido::planner::ConfigurationToConfigurationPlanner>
+          planner,
+      ::dart::dynamics::MetaSkeletonPtr metaSkeleton,
+      std::unique_ptr<common::RNG> rng);
```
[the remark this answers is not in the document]
sniyaz: I agree that putting something in the base `ConfigurationToConfigurationPlanner` class (or even `SingleProblemPlanner`) would make the constructor neater. There are two issues I wanted to discuss though:

1) When the need for an RNG pops up, it's a pretty important change from the original planner (i.e. introduction of sampling like we have here). For something like that, I think being able to see the RNG used (from the view of the programmer) on construction is beneficial.

2)  How often do we think an adapter will (potentially) need to add an RNG to the delegate planner? I'm worried that we're adding stuff to a class for something that might be a specific use case.

@jslee02 what do you think?

src/planner/dart/ConfigurationToEndEffectorOffset.cpp
sniyaz: After thinking about it, just moving this method into robot/util seems like the best way around this.

sniyaz:
I've implelemented @brianhou 's suggestion about including an `mRng` member in the base `Planner` class. This includes the `getRng` getter, which will return `nullptr` if no RNG is present (different planners can take an RNG in their constructor and eventually pass them to the base `Planner` constructor. I've changed `ConfigurationToConfiguration` to take an RNG in its constructor so that derivative classes that wrap OMPL planners can have their own RNG. `ConfigurationToConfiguration_to_ConfigurationToTSR` has been changed to get an RNG from the delegate planner, and create one if it is null.

The downside is that we can't control directly which RNG is used, but after looking at the change I'm also convinced this is the way to do it since it _does_ make things neater.

I also moved `getEndEffectorDirection` into a new `planner/dart/util` file, so it can be used in both `ConfigurationToEndEffectorOffset` and `ConcreteManipulator::planEndEffectorStraight`, without the need to pass a `ConcreteManipulator`. (It couldn't be put in `robot/util` since this creates a cyclic dependency between `planner` and `robot`). 

Let me know your thoughts!

src/planner/Planner.cpp
```diff
@@ -18,6 +19,17 @@ statespace::ConstStateSpacePtr Planner::getStateSpace() const
...
 
+//==============================================================================
+std::unique_ptr<common::RNG> Planner::getRng() const
+{
+  if (!mRng)
+    return nullptr;
+
+  // NOTE: We usually use unique pointers to pass RNGs, so cloning is needed
+  // here.
+  return std::move(cloneRNGFrom(*mRng)[0]);
```
jslee02: This function is somewhat misleading because it returns a clone instead of returning the original RNG in this planner. I think we should either (1) return a raw pointer or (2) change the function name to something like `cloneRng()`. I'm inclined to (1) because the user could clone from a raw pointer of RNG if necessary.

If we change so, this function might need to be non-const. Otherwise, for the sake of const-correctness, it should return `const common::RNG*`, but it would prevent us from generating random values since `RNG::operator()` is not a const function.

Nit: For the future reference, `move` in `return std::move(...)` is superfluous (please see this [post](https://stackoverflow.com/a/14856553/3122234))

src/planner/dart/ConfigurationToConfiguration_to_ConfigurationToTSR.cpp
```diff
@@ -5,12 +5,19 @@
...
 using aikido::constraint::dart::TSR;
 using aikido::constraint::dart::createSampleableBounds;
 using aikido::statespace::dart::MetaSkeletonStateSaver;
 using aikido::statespace::dart::MetaSkeletonStateSpace;
+using ::dart::dynamics::BodyNode;
 using ::dart::dynamics::InverseKinematics;
+using ::dart::dynamics::ConstBodyNodePtr;
+using ::dart::dynamics::BodyNodePtr;
+
+using aikido::common::cloneRNGFrom;
```
jslee02: Nit: Let's not to make this line orphan (please move this to the above `using aikido::...` group)  :grin:

include/aikido/planner/dart/ConfigurationToEndEffectorOffset.hpp
```diff
@@ -16,24 +17,82 @@ namespace dart {
 class ConfigurationToEndEffectorOffset : public Problem
 {
 public:
-  /// Constructor.
+  /// Constructor. Note that this constructor takes the start state from the
+  /// current state of the passed MetaSkeleton, and sets the direction on
+  /// construction.
   ///
   /// \param[in] stateSpace State space.
+  /// param[in] metaSkeleton MetaSkeleton that getStartState will return the
```
jslee02: Nit: \param[in]

sniyaz:
We are- right now we're just using `boost::none` as the stored direction (private variable) when a constructor not passing a direction is used.

sniyaz:
@jslee02 Right now `boost::optional` is just used internally to store `mDirection` when a constructor is used that doesn't pass a direction- but all of that is private, and should never be exposed to the user. Is that implementation alright with you?

src/planner/Planner.cpp
```diff
@@ -6,10 +6,15 @@ namespace aikido {
...
+Planner::Planner(statespace::ConstStateSpacePtr stateSpace, common::RNG* rng)
   : mStateSpace(std::move(stateSpace))
 {
-  // Do nothing
+  if (!rng)
+    mRng = std::move(
+        std::unique_ptr<common::RNG>(
+            new common::RNGWrapper<std::mt19937>(std::random_device{}())));
+  else
+    mRng = std::move(cloneRNGFrom(*rng)[0]);
```
jslee02: Nit: For readability, please use curly brackets for multiple lines of control statements like:
```cpp
 if (!rng)
 {
   mRng = std::move(
       std::unique_ptr<common::RNG>(
           new common::RNGWrapper<std::mt19937>(std::random_device{}())));
 }
 else
 {
   mRng = std::move(cloneRNGFrom(*rng)[0]);
 }
```

src/planner/Planner.cpp
```diff
@@ -6,10 +6,15 @@ namespace aikido {
...
 //==============================================================================
-Planner::Planner(statespace::ConstStateSpacePtr stateSpace)
+Planner::Planner(statespace::ConstStateSpacePtr stateSpace, common::RNG* rng)
   : mStateSpace(std::move(stateSpace))
 {
-  // Do nothing
+  if (!rng)
+    mRng = std::move(
+        std::unique_ptr<common::RNG>(
+            new common::RNGWrapper<std::mt19937>(std::random_device{}())));
```
jslee02: Nit: Maybe this can be simplified as:
```cpp
mRng.reset(new common::RNGWrapper<std::mt19937>(std::random_device{}());
```

src/planner/Planner.cpp
```diff
@@ -18,6 +23,12 @@ statespace::ConstStateSpacePtr Planner::getStateSpace() const
   return mStateSpace;
 }
 
+//==============================================================================
+common::RNG* Planner::getRng() const
```
jslee02: This violates const-correctness. It should be `common::RNG* getRng()`. I think const version (i.e., `const common::RNG* getRng() const`) wouldn't make sense because RNG needs to be non-const as the random generation operator isn't const.

src/planner/dart/ConfigurationToConfiguration_to_ConfigurationToConfiguration.cpp
```diff
@@ -0,0 +1,44 @@
...
+using aikido::statespace::dart::MetaSkeletonStateSpace;
+
+namespace aikido {
+namespace planner {
+namespace dart {
+
+//==============================================================================
+ConfigurationToConfiguration_to_ConfigurationToConfiguration::
+    ConfigurationToConfiguration_to_ConfigurationToConfiguration(
+        std::shared_ptr<aikido::planner::ConfigurationToConfigurationPlanner>
```
jslee02: Nit: `aikido::` isn't necessary.

src/planner/dart/ConfigurationToConfiguration_to_ConfigurationToConfiguration.cpp
```diff
@@ -0,0 +1,44 @@
...
+            planner,
+        ::dart::dynamics::MetaSkeletonPtr metaSkeleton)
+  : PlannerAdapter<aikido::planner::ConfigurationToConfigurationPlanner,
+                   aikido::planner::dart::ConfigurationToConfigurationPlanner>(
```
jslee02: Nit: `aikido::` isn't necessary.

src/planner/dart/ConfigurationToConfiguration_to_ConfigurationToConfiguration.cpp
```diff
@@ -0,0 +1,44 @@
...
+namespace planner {
+namespace dart {
+
+//==============================================================================
+ConfigurationToConfiguration_to_ConfigurationToConfiguration::
+    ConfigurationToConfiguration_to_ConfigurationToConfiguration(
+        std::shared_ptr<aikido::planner::ConfigurationToConfigurationPlanner>
+            planner,
+        ::dart::dynamics::MetaSkeletonPtr metaSkeleton)
+  : PlannerAdapter<aikido::planner::ConfigurationToConfigurationPlanner,
```
jslee02: Nit: `aikido::` isn't necessary.

src/planner/dart/ConfigurationToConfiguration_to_ConfigurationToConfiguration.cpp
```diff
@@ -0,0 +1,44 @@
...
+                   aikido::planner::dart::ConfigurationToConfigurationPlanner>(
+        std::move(planner), std::move(metaSkeleton))
+{
+  // Do nothing
+}
+
+//==============================================================================
+trajectory::TrajectoryPtr
+ConfigurationToConfiguration_to_ConfigurationToConfiguration::plan(
+    const aikido::planner::dart::ConfigurationToConfiguration& problem,
```
jslee02: Nit: `aikido::` isn't necessary.

src/planner/dart/ConfigurationToConfiguration_to_ConfigurationToTSR.cpp
```diff
@@ -19,9 +25,10 @@ namespace dart {
 //==============================================================================
 ConfigurationToConfiguration_to_ConfigurationToTSR::
     ConfigurationToConfiguration_to_ConfigurationToTSR(
-        std::shared_ptr<ConfigurationToConfigurationPlanner> planner,
+        std::shared_ptr<aikido::planner::ConfigurationToConfigurationPlanner>
```
jslee02: Nit: `aikido::` isn't necessary.

src/planner/dart/ConfigurationToConfiguration_to_ConfigurationToTSR.cpp
```diff
@@ -19,9 +25,10 @@ namespace dart {
...
+            planner,
         ::dart::dynamics::MetaSkeletonPtr metaSkeleton)
-  : PlannerAdapter<ConfigurationToConfigurationPlanner,
+  : PlannerAdapter<aikido::planner::ConfigurationToConfigurationPlanner,
```
jslee02: Nit: `aikido::` isn't necessary.

src/planner/dart/ConfigurationToConfiguration_to_ConfigurationToTSR.cpp
```diff
@@ -57,30 +64,27 @@ ConfigurationToConfiguration_to_ConfigurationToTSR::plan(
   auto ik = InverseKinematics::create(endEffectorBodyNode);
   ik->setDofs(mMetaSkeleton->getDofs());
 
-  // Convert TSR constraint into IK constraint
+  // Get the start state form the MetaSkeleton, since this is a DART planner.
+  auto startState = mMetaSkeletonStateSpace->createState();
+  mMetaSkeletonStateSpace->getState(mMetaSkeleton.get(), startState);
+
+  auto rng = std::move(cloneRNGFrom(*mDelegate->getRng())[0]);
```
jslee02: I could be wrong, but it seems `std::move()` isn't necessary.

sniyaz: It appears this is needed, since this is a unique pointer and it can't be assigned, only moved.

sniyaz:
I fixed the nits, but I'm trying to fix a build issue with Clan/OSX on Travis.

brianhou:
Hmm, it looks like a potentially weird dependency issue with dart/nlopt/libccd? [Here's](https://travis-ci.org/personalrobotics/aikido/jobs/398432627#L5258) the error message (which is repeated several times):

> CMake Error in src/common/CMakeLists.txt:
>   Imported target "dart-optimizer-nlopt" includes non-existent path
>     "/usr/local/Cellar/libccd/2.0_2/include"
>   in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:
>   * The path was deleted, renamed, or moved to another location.
>   * An install or uninstall procedure did not complete successfully.
>   * The installation package was faulty and references files it does not
>   provide.

jslee02:
I believe the macOS build failure is due to that the bottles of `dartsim` haven't updated after `libccd` updated. This should be fixed once https://github.com/Homebrew/homebrew-core/pull/29620 is merged.

include/aikido/planner/Planner.hpp
```diff
@@ -44,6 +49,9 @@ class Planner
 protected:
   /// State space associated with this planner.
   statespace::ConstStateSpacePtr mStateSpace;
+
+  /// RNG the planner uses.
+  std::unique_ptr<common::RNG> mRng;
```
brianhou: If we're taking in a raw `RNG` pointer, do we need to store this as a `unique_ptr`?

src/planner/dart/ConfigurationToConfiguration.cpp
```diff
@@ -0,0 +1,75 @@
...
+//==============================================================================
+ConfigurationToConfiguration::ConfigurationToConfiguration(
+    statespace::dart::ConstMetaSkeletonStateSpacePtr stateSpace,
+    ::dart::dynamics::ConstMetaSkeletonPtr metaSkeleton,
+    const statespace::dart::MetaSkeletonStateSpace::State* goalState,
+    constraint::ConstTestablePtr constraint)
+  : Problem(stateSpace, std::move(constraint))
+  , mMetaSkeletonStateSpace(stateSpace)
+  , mMetaSkeleton(std::move(metaSkeleton))
+  , mStartState(std::move(mMetaSkeletonStateSpace->createState()))
```
brianhou: @jslee02 Are these explicit `std::move`s for `mStartState` and `mGoalState` necessary? Based on my limited understanding of rvalues, those expressions might already be rvalues and therefore `std::move` is unnecessary?

src/planner/dart/ConfigurationToConfiguration.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+//==============================================================================
+const statespace::dart::MetaSkeletonStateSpace::State*
+ConfigurationToConfiguration::getStartState() const
+{
+  // Take start state from MetaSkeleton if passed. Store in the ScopedState
+  // instance variable to avoid dangling pointers.
+  if (mMetaSkeleton)
+  {
+    auto startState = mMetaSkeletonStateSpace->createState();
```
brianhou: Could this just be
```cpp
if (mMetaSkeleton)
  mMetaSkeletonStateSpace->getState(mMetaSkeleton.get(), mStartState);
```

It seems that `mStartState` is already a `MetaSkeletonStateSpace::ScopedState`, so I'm not sure why `mStateSpace` needs to be used here.

include/aikido/planner/dart/ConfigurationToConfiguration_to_ConfigurationToConfiguration.hpp
```diff
@@ -0,0 +1,33 @@
...
+#include "aikido/planner/ConfigurationToConfigurationPlanner.hpp"
+#include "aikido/planner/dart/ConfigurationToConfigurationPlanner.hpp"
+#include "aikido/planner/dart/PlannerAdapter.hpp"
+
+namespace aikido {
+namespace planner {
+namespace dart {
+
+class ConfigurationToConfiguration_to_ConfigurationToConfiguration
+    : public PlannerAdapter<aikido::planner::
```
brianhou: Nit: I think `aikido::` is unnecessary here?

src/planner/dart/ConfigurationToConfiguration_to_ConfigurationToTSR.cpp
```diff
@@ -57,30 +63,27 @@ ConfigurationToConfiguration_to_ConfigurationToTSR::plan(
   auto ik = InverseKinematics::create(endEffectorBodyNode);
   ik->setDofs(mMetaSkeleton->getDofs());
 
-  // Convert TSR constraint into IK constraint
+  // Get the start state form the MetaSkeleton, since this is a DART planner.
```
brianhou: Typo: `form` -> `from`

include/aikido/planner/dart/ConfigurationToEndEffectorOffset.hpp
```diff
@@ -16,27 +17,85 @@ namespace dart {
 class ConfigurationToEndEffectorOffset : public Problem
```
brianhou: We should augment this documentation here with the types of constructors that are supported.

src/planner/dart/ConfigurationToEndEffectorOffset.cpp
```diff
@@ -48,13 +107,27 @@ ConfigurationToEndEffectorOffset::getEndEffectorBodyNode() const
 const statespace::dart::MetaSkeletonStateSpace::State*
 ConfigurationToEndEffectorOffset::getStartState() const
 {
+  // Take start state from MetaSkeleton if passed. Store in the ScopedState
+  // instance variable to avoid dangling pointers.
+  if (mMetaSkeleton)
+  {
+    auto startState = mMetaSkeletonStateSpace->createState();
```
brianhou: Same as above?

include/aikido/robot/ConcreteRobot.hpp
```diff
@@ -102,6 +102,20 @@ class ConcreteRobot : public Robot
       const dart::dynamics::MetaSkeletonPtr& metaSkeleton,
       const constraint::dart::CollisionFreePtr& collisionFree) const override;
 
+  // Get a smoothing post postprocessor that respects velocity and acceleration
+  // limits, as well as the passed constraint.
+  /// \param[in] metaSkeleton The MetaSkeleton whose limits must be respected.
+  std::shared_ptr<aikido::planner::TrajectoryPostProcessor>
+  getTrajectoryPostProcessor(
```
brianhou: I'm fine with this for now, but in light of the discussion from #443 it seems that maybe it doesn't make sense to have this here? I think we should probably set up the postprocessor in the concrete robot constructor, similar to how we set up the metaplanner.

include/aikido/planner/dart/ConfigurationToConfiguration.hpp
```diff
@@ -0,0 +1,75 @@
...
+  /// \param[in] stateSpace State space.
+  /// param[in] metaSkeleton MetaSkeleton that getStartState will return the
+  /// current state of when called.
+  /// \param[in] goalState Goal state to plan to.
+  /// \param[in] constraint Trajectory-wide constraint that must be satisfied.
+  ConfigurationToConfiguration(
+      statespace::dart::ConstMetaSkeletonStateSpacePtr stateSpace,
+      ::dart::dynamics::ConstMetaSkeletonPtr metaSkeleton,
+      const statespace::dart::MetaSkeletonStateSpace::State* goalState,
+      constraint::ConstTestablePtr constraint);
```
gilwoolee: Can we allow `nullptr` for constraint and make it default to `Satisfied` if `nullptr` is passed?

include/aikido/planner/dart/ConfigurationToConfiguration_to_ConfigurationToConfiguration.hpp
```diff
@@ -0,0 +1,33 @@
...
+                                ConfigurationToConfigurationPlanner,
+                            aikido::planner::dart::
+                                ConfigurationToConfigurationPlanner>
+{
+public:
+  ConfigurationToConfiguration_to_ConfigurationToConfiguration(
```
gilwoolee: Please add docstring to this. :)

include/aikido/planner/dart/ConfigurationToConfiguration_to_ConfigurationToTSR.hpp
```diff
@@ -10,12 +10,14 @@ namespace planner {
 namespace dart {
 
 class ConfigurationToConfiguration_to_ConfigurationToTSR
-    : public PlannerAdapter<ConfigurationToConfigurationPlanner,
+    : public PlannerAdapter<aikido::planner::
```
gilwoolee: Nit: same as @brianhou 's comment above, `aikido::planner` wouldn't be necessary here.

include/aikido/planner/dart/ConfigurationToTSR.hpp
```diff
@@ -15,21 +15,43 @@ namespace dart {
 class ConfigurationToTSR : public Problem
 {
 public:
-  /// Constructor.
+  /// Constructor. Note that this constructor takes the start state from the
+  /// current state of the passed MetaSkeleton.
   ///
   /// \param[in] stateSpace State space.
+  /// param[in] metaSkeleton MetaSkeleton that getStartState will return the
```
gilwoolee: Nit: typo. `\param[in]`

src/planner/dart/ConfigurationToEndEffectorOffset.cpp
```diff
@@ -9,21 +10,79 @@ namespace dart {
...
+  , mMetaSkeleton(std::move(metaSkeleton))
+  , mStartState(std::move(mMetaSkeletonStateSpace->createState()))
   , mEndEffectorBodyNode(std::move(endEffectorBodyNode))
-  , mStartState(startState)
-  , mDirection(direction.normalized())
+  , mDirection(direction)
   , mDistance(signedDistance)
 {
-  if (direction.squaredNorm() == 0)
+  if (mDirection.get().squaredNorm() == 0)
```
gilwoolee: Shouldn't this be `<= 1e-3` or something that's not exactly zero?

src/planner/dart/ConfigurationToEndEffectorOffset.cpp
```diff
@@ -9,21 +10,79 @@ namespace dart {
...
+    constraint::ConstTestablePtr constraint)
+  : Problem(stateSpace, std::move(constraint))
+  , mMetaSkeletonStateSpace(stateSpace)
+  , mMetaSkeleton(nullptr)
+  , mStartState(std::move(stateSpace->cloneState(startState)))
+  , mEndEffectorBodyNode(std::move(endEffectorBodyNode))
+  , mDirection(direction)
+  , mDistance(signedDistance)
+{
+  if (mDirection.get().squaredNorm() == 0)
```
gilwoolee: Same question as above.

src/planner/dart/ConfigurationToEndEffectorOffset.cpp
brianhou: I think this is only here because normalizing a zero vector throws an error. I don't think it's wrong to pass in a small-magnitude vector and use that resulting direction, so I think comparing exactly to zero is fine. However, I think it might be better to use `.isZero()` rather than `.squaredNorm() == 0`.

src/planner/dart/ConfigurationToConfiguration.cpp
sniyaz: Yeah, I agree with this. @jslee02 can you confirm?

include/aikido/planner/dart/ConfigurationToConfiguration.hpp
sniyaz: It looks like the default `Problem` class handles this:

https://github.com/personalrobotics/aikido/blob/fe09cbb5d2f57a3366eb1cc4a8ec932f1436be0c/src/planner/Problem.cpp#L17

jslee02: @sniyaz That's true. I think all we need to do here is just to pass `nullptr` as the default value to `constraint` so that the default `Problem` do the job.

include/aikido/planner/dart/ConfigurationToEndEffectorOffset.hpp
```diff
@@ -12,31 +13,94 @@ namespace planner {
...
       const Eigen::Vector3d& direction,
       double signedDistance,
       constraint::ConstTestablePtr constraint);
 
+  /// Constructor. Note that this constructor takes the start state from the
+  /// current state of the passed MetaSkeleton, and takes the direction from the
+  /// end-effector BodyNode's current direction.
+  ///
+  /// \param[in] stateSpace State space.
+  /// param[in] metaSkeleton MetaSkeleton that getStartState will return the
```
jslee02: Nit: \param[in]

include/aikido/planner/dart/ConfigurationToEndEffectorOffset.hpp
```diff
@@ -47,26 +111,34 @@ class ConfigurationToEndEffectorOffset : public Problem
   /// while maintaining the current orientation.
   ::dart::dynamics::ConstBodyNodePtr getEndEffectorBodyNode() const;
 
-  /// Returns the start state.
+  /// Return the start state to plan from, either set on construction or
```
jslee02: Nit: Returns

include/aikido/planner/dart/ConfigurationToEndEffectorOffset.hpp
```diff
@@ -12,31 +13,94 @@ namespace planner {
...
+  /// offest while maintaining the current orientation. Current direction is
+  /// returned when getDirection is called.
+  /// \param[in] signedDistance Signed distance to move, in meters.
+  /// \param[in] constraint Trajectory-wide constraint that must be satisfied.
+  ConfigurationToEndEffectorOffset(
+      statespace::dart::ConstMetaSkeletonStateSpacePtr stateSpace,
+      const statespace::dart::MetaSkeletonStateSpace::State* startState,
+      ::dart::dynamics::ConstBodyNodePtr endEffectorBodyNode,
+      double signedDistance,
+      constraint::ConstTestablePtr constraint);
```
jslee02: Nit: Let's pass `nullptr` to `constraint` as the default value for the same reason.

include/aikido/planner/dart/ConfigurationToEndEffectorOffset.hpp
```diff
@@ -47,26 +111,34 @@ class ConfigurationToEndEffectorOffset : public Problem
...
   const statespace::dart::MetaSkeletonStateSpace::State* getStartState() const;
 
-  /// Returns the direction of motion specified in the world frame.
-  ///
-  /// \note The direction is allowed to be zero vector.
-  const Eigen::Vector3d& getDirection() const;
+  /// Returns the direction of motion specified in the world frame. Note that
+  /// if no direction was passed, the current direction of the end effector body
+  /// node is returned.
+  const Eigen::Vector3d getDirection() const;
```
jslee02: Nit: No point to return const value.

include/aikido/planner/dart/ConfigurationToEndEffectorPose.hpp
```diff
@@ -14,19 +14,38 @@ namespace dart {
 class ConfigurationToEndEffectorPose : public Problem
 {
 public:
-  /// Constructor.
+  /// Constructor. Note that this constructor takes the start state from the
+  /// current state of the passed MetaSkeleton.
   ///
   /// \param[in] stateSpace State space.
+  /// param[in] metaSkeleton MetaSkeleton that getStartState will return the
```
jslee02: Nit: \param[in]

include/aikido/planner/dart/ConfigurationToEndEffectorPose.hpp
```diff
@@ -14,19 +14,38 @@ namespace dart {
...
-  /// \param[in] startState Start state.
   /// \param[in] goalPose Goal pose.
   /// \throw If \c stateSpace is not compatible with \c constraint's state
   /// space.
   ConfigurationToEndEffectorPose(
       statespace::dart::ConstMetaSkeletonStateSpacePtr stateSpace,
+      ::dart::dynamics::ConstMetaSkeletonPtr metaSkeleton,
       ::dart::dynamics::ConstBodyNodePtr endEffectorBodyNode,
+      const Eigen::Isometry3d& goalPose,
+      constraint::ConstTestablePtr constraint);
```
jslee02: Nit: Let's pass `nullptr` to constraint as the default value for the same reason.

include/aikido/planner/dart/ConfigurationToEndEffectorPose.hpp
```diff
@@ -14,19 +14,38 @@ namespace dart {
...
+  /// pose.
+  /// \param[in] goalPose Goal pose.
+  /// \throw If \c stateSpace is not compatible with \c constraint's state
+  /// space.
+  ConfigurationToEndEffectorPose(
+      statespace::dart::ConstMetaSkeletonStateSpacePtr stateSpace,
       const statespace::dart::MetaSkeletonStateSpace::State* startState,
+      ::dart::dynamics::ConstBodyNodePtr endEffectorBodyNode,
       const Eigen::Isometry3d& goalPose,
       constraint::ConstTestablePtr constraint);
```
jslee02: Nit: Let's pass `nullptr` to constraint as the default value for the same reason.

include/aikido/planner/dart/ConfigurationToTSR.hpp
```diff
@@ -15,21 +15,43 @@ namespace dart {
...
   /// \param[in] constraint Trajectory-wide constraint that must be satisfied.
   /// \throw If \c stateSpace is not compatible with \c constraint's state
   /// space.
   ConfigurationToTSR(
       statespace::dart::ConstMetaSkeletonStateSpacePtr stateSpace,
+      ::dart::dynamics::ConstMetaSkeletonPtr metaSkeleton,
       ::dart::dynamics::ConstBodyNodePtr endEffectorBodyNode,
+      std::size_t maxSamples,
+      constraint::dart::ConstTSRPtr goalTSR,
+      constraint::ConstTestablePtr constraint);
```
jslee02: Nit: Let's pass `nullptr` to constraint as the default value for the same reason.

include/aikido/planner/dart/ConfigurationToTSR.hpp
```diff
@@ -15,21 +15,43 @@ namespace dart {
...
+  /// \param[in] constraint Trajectory-wide constraint that must be satisfied.
+  /// \throw If \c stateSpace is not compatible with \c constraint's state
+  /// space.
+  ConfigurationToTSR(
+      statespace::dart::ConstMetaSkeletonStateSpacePtr stateSpace,
       const statespace::dart::MetaSkeletonStateSpace::State* startState,
+      ::dart::dynamics::ConstBodyNodePtr endEffectorBodyNode,
       std::size_t maxSamples,
       constraint::dart::ConstTSRPtr goalTSR,
       constraint::ConstTestablePtr constraint);
```
jslee02: Nit: Let's pass `nullptr` to constraint as the default value for the same reason.

include/aikido/planner/dart/ConfigurationToTSR.hpp
```diff
@@ -46,22 +68,30 @@ class ConfigurationToTSR : public Problem
   /// Returns the maximum number of TSR samples to plan to.
   std::size_t getMaxSamples() const;
 
-  /// Returns the start state.
+  /// Return the start state to plan from, either set on construction or
```
jslee02: Nit: Returns

brianhou:
It looks like the only failures are due to the pending Homebrew PR, so I'm going to merge this. Thanks for all the changes!

Issue #433, opened 2018-05-22
Trajopt implementation


jslee02:
#### Procedure
1. Implement most basic functionalities for trajectory optimization that uses nonlinear optimization solver like nlopt, ipopt.
  1. define the cost of a trajectory CHOMP-like as the integral of distances from obstacles, including self-collisions, and whatever else you want.
  1. Define constraints on joint positions [if you need to be dynamic include vel and acc]
  1. Send this to an optimizer.
1. Add functionality of swept-volume collision checking
1. Add SQP optimization

#### Todos
* Available optimizers through DART are `nlopt` and `ipopt`, which are for nonlinear programming. It would be nice to search quadratic programming (QP) specific solvers.

#### References
* [trajopt](http://rll.berkeley.edu/trajopt/doc/sphinx_build/html/) ([github](https://github.com/joschu/trajopt))
* https://en.wikipedia.org/wiki/Quadratic_programming
* https://doc.cgal.org/latest/QP_solver/index.html
* http://www.numerical.rl.ac.uk/people/nimg/qp/qp.html
* http://plato.asu.edu/sub/nlores.html#QP-problem

PR #443, opened 2018-06-11
Adding another kinodynamic timer

dqyi11:
This PR introduces another kinodynamic timer by using a time-optimal trajectory generation algorithm, which is developed by Tobias Kunz. 
Code: [https://github.com/tobiaskunz/trajectories](https://github.com/tobiaskunz/trajectories)
Publication: T. Kunz, M. Stilman. **Time-Optimal Trajectory Generation for Path Following with Bounded Acceleration and Velocity.** Robotics: Science and Systems (RSS), 2012. [http://www.roboticsproceedings.org/rss08/p27.html](http://www.roboticsproceedings.org/rss08/p27.html)

This type of timer accepts both `Interpolated` and `Spline` paths. It firstly converts a non-differentiable linear-piecewise trajectory to a differentiable one by adding circular blends. It then applies a time-optimal exact path-following algorithm that maximizes the path velocities at every point along the path. 

The preprocessing of adding circular blends might slightly change the shape of a path. However, unlike how the existing parabolic timer does, the timed trajectory does not make a complete stop (zero velocities) at each waypoint.

QUESTIONs: 
- Shall we create a `postprocessing` folder for all the postprocessing methods? 
- How are we going to enable `Robot` class call different timing/smoothing methods?


brianhou:
I'll try to take a more detailed look over this PR in the next couple days, but I have one quick question: how does this differ from Mintos? My understanding was that we had an implementation in `features/minTOS` that was working on an early version of AIKIDO, so it seemed like it wouldn't have been too difficult to update.

> Shall we create a `postprocessing` folder for all the postprocessing methods?

I think we can continue creating a namespace/directory underneath `aikido::planner` for each type of planner or postprocessor. If we decide we don't like it, we can always change it later.

> How are we going to enable `Robot` class call different timing/smoothing methods?

I don't have a clear understanding of when we should use one or the other. Is it always better to use this over the parabolic smoother, which forces us to stop at waypoints? Or are there certain planning problems where we believe that the parabolic smoother would be better? If we should always use one, then I think we can continue doing what we're doing with constructing a single postprocessor in the robot constructor. Otherwise, I guess we can construct all the postprocessors in the constructor (similar to how we construct all the planners in the robot constructor) and then delegate accordingly.

dqyi11:
Thanks! @brianhou 
I think that `minTOS` is going to be used as an alternative smoother that considers constraint manifolds and kinematic limitations. I am seeing them as different postprocessing methods by the changes in path geometry. `ParabolicTimer` stops at each waypoint so no geometric change is introduced. This timing method will add circular blend at each waypoint so the path geometry is adjusted to be differentiable. The change of path geometry can be kept under a tolerance threshold by the parameter `maxDeviation`. The `ParabolicSmoother` and upcoming `minTOS` will "optimize" the path geometry according to some criteria.  

I suggest to keep different postprocessing methods. When we want a robot shall strictly follow a planned path (e.g. welding?), ParabolicTiming can be used. When we want a robot to keep the path shape with a tolerance (e.g. forward pushing?), this timing can be used. When we want a robot to generate an efficient and smooth motion, smoothers can be used.

So my original suggestion was whether we can have an `aikido::planner::postprocessing` namespace for these different postprocessing methods. The current `aikido::planner::parabolic` can be moved under that one.

CHANGELOG.md
```diff
@@ -43,6 +43,7 @@
   * Added flags to WorldStateSaver to specify what to save: [#339](https://github.com/personalrobotics/aikido/pull/339)
   * Changed interface for TrajectoryPostProcessor: [#341](https://github.com/personalrobotics/aikido/pull/341)
   * Planning calls with InverseKinematicsSampleable constraints explicitly set MetaSkeleton to solve IK with: [#379](https://github.com/personalrobotics/aikido/pull/379)
+  * Add a kinodynamic timer that generates a time-optimal smooth trajectory without completely stopping at each waypoint: [#443}(https://github.com/personalrobotics/aikido/pull/443)
```
brianhou: Typo: `[#433]`

src/planner/kinodynamic/KinodynamicTimer.cpp
```diff
@@ -0,0 +1,273 @@
+#include "aikido/planner/kinodynamic/KinodynamicTimer.hpp"
+#include <dart/dart.hpp>
+#include <aikido/common/Spline.hpp>
+#include <aikido/common/StepSequence.hpp>
+
+#include "Path.h"
```
brianhou: @jslee02 Are relative paths to the external files okay here or should this be something else?

src/planner/kinodynamic/KinodynamicTimer.cpp
```diff
@@ -0,0 +1,273 @@
...
+    const Eigen::VectorXd& maxVelocities,
+    const Eigen::VectorXd& maxAccelerations,
+    aikido::statespace::ConstStateSpacePtr stateSpace,
+    double maxDeviation,
+    double timeStep,
+    double startTime)
+{
+  Trajectory trajectory(
+      Path(waypoints, maxDeviation), maxVelocities, maxAccelerations, timeStep);
+  if (trajectory.isValid())
```
brianhou: Let's flip this to something like

```cpp
if (!trajectory.isValid())
  return nullptr;
...
```

include/aikido/planner/kinodynamic/KinodynamicTimer.hpp
```diff
@@ -0,0 +1,139 @@
...
+
+  /// Performs parabolic retiming on an input *spline* trajectory.
+  /// \copydoc TrajectoryPostProcessor::postprocess
+  std::unique_ptr<aikido::trajectory::Spline> postprocess(
+      const aikido::trajectory::Spline& inputTraj,
+      const aikido::common::RNG& rng,
+      const aikido::constraint::TestablePtr& constraint = nullptr) override;
+
+  /// Returns the velocity limits of the dimensions
+  const Eigen::VectorXd& getVelocityLimits() const;
```
brianhou: Are all these getters and setters useful? I'm not sure if they're necessary.

src/planner/kinodynamic/KinodynamicTimer.cpp
```diff
@@ -0,0 +1,273 @@
...
+#include "Trajectory.h"
+
+namespace aikido {
+namespace planner {
+namespace kinodynamic {
+
+//==============================================================================
+std::unique_ptr<aikido::trajectory::Spline>
+createSplineFromWaypointsAndConstraints(
+    const std::list<Eigen::VectorXd>& waypoints,
```
brianhou: Why a `list` over a `vector`?

src/planner/kinodynamic/KinodynamicTimer.cpp
```diff
@@ -0,0 +1,273 @@
...
+    auto outputTrajectory
+        = make_unique<aikido::trajectory::Spline>(stateSpace, startTime);
+
+    // create a sequence of time steps from start time to end time by time step
+    aikido::common::StepSequence sequence(
+        timeStep, true, true, startTime, endTime);
+
+    const Eigen::VectorXd zeroPosition = Eigen::VectorXd::Zero(dimension);
+    auto currState = stateSpace->createState();
+    for (std::size_t i = 1; i < sequence.getLength(); i++)
```
brianhou: I guess this is a personal preference, but can we start `i = 0` and go to `sequence.getLength() - 1`?

src/planner/kinodynamic/KinodynamicTimer.cpp
```diff
@@ -0,0 +1,273 @@
...
+#include "Path.h"
+#include "Trajectory.h"
+
+namespace aikido {
+namespace planner {
+namespace kinodynamic {
+
+//==============================================================================
+std::unique_ptr<aikido::trajectory::Spline>
+createSplineFromWaypointsAndConstraints(
```
brianhou: My understanding of this function is that we're
- Calling this library to generate a smooth and retimed trajectory from a sequence of waypoints (from our planned trajectory)
- Sampling points at each timestep from their trajectory, then fitting a spline to this sequence of points

I haven't read the paper so I don't know what their trajectory representation is, but it's not clear why discarding that representation and refitting a spline is necessary. I'm also concerned that it doesn't seem that we're checking that our planning constraints remain satisfied after postprocessing.

src/planner/kinodynamic/KinodynamicTimer.cpp
```diff
@@ -0,0 +1,273 @@
...
+
+    if (maxAcceleration[i] <= 0.)
+      throw std::invalid_argument("Acceleration limits must be positive.");
+    if (!std::isfinite(maxAcceleration[i]))
+      throw std::invalid_argument("Acceleration limits must be finite.");
+  }
+
+  double startTime = inputTrajectory.getStartTime();
+  // create waypoints from Interpolated path
+  std::list<Eigen::VectorXd> waypoints;
```
brianhou: We should reserve the length of this vector in advance, since we know how long it is.

src/planner/kinodynamic/KinodynamicTimer.cpp
```diff
@@ -0,0 +1,273 @@
...
+      throw std::invalid_argument("Velocity limits must be finite.");
+
+    if (maxAcceleration[i] <= 0.)
+      throw std::invalid_argument("Acceleration limits must be positive.");
+    if (!std::isfinite(maxAcceleration[i]))
+      throw std::invalid_argument("Acceleration limits must be finite.");
+  }
+
+  double startTime = inputTrajectory.getStartTime();
+  // create waypoints from Interpolated path
```
brianhou: I think it would be nice to create `Eigen::VectorXd getWaypoints(trajectory::Interpolated traj)` (and for splines), then call the appropriate function from here.

src/planner/kinodynamic/KinodynamicTimer.cpp
```diff
@@ -0,0 +1,273 @@
...
+      mAccelerationLimits,
+      mMaxDeviation,
+      mTimeStep);
+}
+
+//==============================================================================
+std::unique_ptr<aikido::trajectory::Spline> KinodynamicTimer::postprocess(
+    const aikido::trajectory::Spline& inputTraj,
+    const aikido::common::RNG& /*rng*/,
+    const aikido::constraint::TestablePtr& /*constraint*/)
```
brianhou: I think we need to test that this constraint isn't being violated.

src/planner/kinodynamic/KinodynamicTimer.cpp
```diff
@@ -0,0 +1,273 @@
...
+  , mTimeStep(timeStep)
+{
+  // Do nothing
+}
+
+//==============================================================================
+std::unique_ptr<aikido::trajectory::Spline> KinodynamicTimer::postprocess(
+    const aikido::trajectory::Interpolated& inputTraj,
+    const aikido::common::RNG& /*rng*/,
+    const aikido::constraint::TestablePtr& /*constraint*/)
```
brianhou: I think we need to test that this constraint isn't being violated.

tests/planner/kinodynamic/test_KinodynamicTimer.cpp
```diff
@@ -0,0 +1,392 @@
...
+  inputTrajectory.addWaypoint(1., state);
+
+  state.setValue(Vector2d(2., 3.));
+  inputTrajectory.addWaypoint(3., state);
+
+  auto timedTrajectory = computeKinodynamicTiming(
+      inputTrajectory, Vector2d::Constant(2.), Vector2d::Constant(1.));
+
+  timedTrajectory->evaluate(1., state);
+  EXPECT_TRUE(Vector2d(1.0, 2.0).isApprox(state.getValue()));
```
brianhou: I think there are some Eigen things in https://github.com/personalrobotics/aikido/blob/master/tests/eigen_tests.hpp that we could use instead? The tests would be more descriptive when they fail.

tests/planner/kinodynamic/test_KinodynamicTimer.cpp
```diff
@@ -0,0 +1,392 @@
...
+  EXPECT_TRUE(Vector2d(1.0, 1.0).isApprox(tangentVector));
+
+  timedTrajectory->evaluateDerivative(2.0, 1, tangentVector);
+  EXPECT_TRUE(Vector2d(1.0, 1.0).isApprox(tangentVector));
+
+  timedTrajectory->evaluateDerivative(2.5, 1, tangentVector);
+  EXPECT_TRUE(Vector2d(0.5, 0.5).isApprox(tangentVector));
+
+  timedTrajectory->evaluateDerivative(3.0, 1, tangentVector);
+  // TODO: isApprox does not work when comparing against zero.
```
brianhou: I think the same `eigen_tests.hpp` might also help here.

brianhou:
@dqyi11 It seems that #446 depends on this PR being merged first. Would you mind addressing the comments when you have the chance? Thanks!

dqyi11:
@brianhou Yes. I have started working on that. Will get that done soon.

include/aikido/planner/kinodynamic/KinodynamicTimer.hpp
dqyi11: I think the getters are useful. For example, when we want to validate whether a timed trajectory violates the limits but only have the access to the timer handle. I prefer to keep the setters as well. It provides a way of adjusting timer parameters without reconstructing a new timer object.

src/planner/kinodynamic/KinodynamicTimer.cpp
dqyi11: Done! Two new functions are added.

src/planner/kinodynamic/KinodynamicTimer.cpp
dqyi11: I would also like to know the standard. This one is following the style in `HauserParabolicTimer`.

src/planner/kinodynamic/KinodynamicTimer.cpp
dqyi11: Because the external codes take in a `list` of `Eigen::VectorXd`. If we use `vector`, we will either have to do an extra conversion or hack the external code.

src/planner/kinodynamic/KinodynamicTimer.cpp
dqyi11: I think this is relevant to another comment. I did not use `std::vector` to avoid extra conversion.

src/planner/kinodynamic/KinodynamicTimer.cpp
dqyi11: This is a great catch! "Theoretically", it is a timing method that won't change the geometry of a path. So the collision violation is not taken care of (so as for the parabolic timer), by assuming that it is done in the planning. However, it does introduce slight geometry changes in a preprocessing step, which uses circular blends to make a path differentiable. How much the change is controlled by a parameter `maxDeviation`. It is possible that the timing will fail because an unideal path geometry is provided. 

If a user choose a reasonable value for `maxDeviation`, it is very unlike that the planning constraints will be violated.

src/planner/kinodynamic/KinodynamicTimer.cpp
jslee02: It seems this is the current convention unless we change the `CMakeLists.txt` files in the external directories because the include directories of the targets are exactly the directory where the files are located (e.g., `<aikido_dir>/src/external/hauserparabolicsmoother/`).

One potential problem with this approach is that the file name can be ambiguous when there are two files with the same name in the different external libraries. I have a proposal to resolve this problem:

```
src/
  external/
    lib1/
      CMakeLists.txt  # include ./include
      include/
        lib1/
      src/
    lib2/
      CMakeLists.txt  # include ./include
      include/
        lib2/
      src/
```
then we can include the files without conflicts over external libraries as
```cpp
#include "lib1/same_name.h"
#include "lib2/same_name.h"
```

With this proposal, the path to the external headers in the external sources should be changed as the same way above. There is a way to avoid this modification, but I personally prefer this because the leading project name in the header inclusion would also allow the externals to avoid potential name conflicts when they also have dependency.

TL;DR: This is the convention. I have a proposal, which shouldn't block this PR anyway.

src/planner/kinodynamic/KinodynamicTimer.cpp
aditya-vk: Given that the max deviation can be upto half the distance of line segments between consecutive points, we might want to check for collisions even during this procedure. I am not sure if the assumption that planning constraints will not be violated is valid while `0 < maxDeviations < 0.5 x (edge Lengths)`.

include/aikido/planner/kinodynamic/KinodynamicTimer.hpp
```diff
@@ -0,0 +1,139 @@
...
+/// bounds. The output is a parabolic spline, encoded in cubic polynomials.
+/// It firstly preprocesses a non-differentiable path to a differentiable one
+/// by adding circular blends; and then \b exactly follows the preprocessed
+/// path.
+///
+/// The output trajectory consists of a sequence of trapezoidal velocity
+/// profiles that implement bang-bang control. This function curently only
+/// supports \c RealVector, \c SO2, and compound state spaces of those types.
+/// Additionally, this function requires that \c inputTrajectory to be
+/// interpolated using a \c GeodesicInterpolator.
```
aditya-vk: Why does the input trajectory need to have been interpolated using `GeodesicInterpolator`? Why the explicit mention? Just curious. Any sequence of waypoints should be fine right?

include/aikido/planner/kinodynamic/KinodynamicTimer.hpp
```diff
@@ -0,0 +1,139 @@
...
+///
+/// \param[in] inputTrajectory input piecewise Geodesic trajectory
+/// \param[in] maxVelocity maximum velocity for each dimension
+/// \param[in] maxAcceleration maximum acceleration for each dimension
+/// \param[in] maxDeviation maximum deviation from a waypoint in doing circular
+/// blending around the waypoint
+/// \param[in] timeStep time step in following the path
+/// \return time optimal trajectory that satisfies acceleration constraints
```
aditya-vk: Nit: mention velocity constraints too? :thinking:

src/planner/kinodynamic/KinodynamicTimer.cpp
dqyi11: @aditya-vk Hmm, I agree. I suggest that we can have a post validation, which is a constraint validation for a spline trajectory. There is already a ticket about this #212 . I can work on that one. This will be useful to a few other functions.

include/aikido/planner/kinodynamic/KinodynamicTimer.hpp
dqyi11: I think this is necessary. The interpolation implies the shape of a path given a sequence of waypoints. If you choose a different interpolator, the shape of a path will be different (length might also be changed). The timing algorithm works on the assumption of the geodesic interpolation.

dqyi11:
I would like to have suggestions about 
(1) the current namespace `aikido::planner::kinodynamic` (or where the source codes shall be in); and
(2) the current name of the external source folder name `optimal_time_path_following` in `aikido/src/external`.

brianhou:
(1) I think `kinodynamic` might not be a good namespace, since we might want to have actual kinodynamic planners. What about `optimal_retimer`?
  
I agree that eventually it makes sense to move all the postprocessing methods into their own namespace, although we'll also have to decide whether that would be `aikido::postprocessing` or `aikido::planner::postprocessing` or something else.

(2) How about `kunz_optimal_retimer`?

dqyi11:
@brianhou I have updated the namespace and folder name as suggested.
I would suggest `aikido::planner::postprocessing`. My reason is that `Planning`, `Control` and `Perception` are currently three main modules in `aikido`. `Postprocessing` seems part of `Planning`. I think `prpy` uses the same structure.

src/planner/optimalretimer/KinodynamicTimer.cpp
```diff
@@ -0,0 +1,292 @@
...
+#include "Path.h"
+#include "Trajectory.h"
+
+namespace aikido {
+namespace planner {
+namespace optimalretimer {
+
+//==============================================================================
+std::unique_ptr<aikido::trajectory::Spline>
+createSplineFromWaypointsAndConstraints(
```
brianhou: In order to match the parabolic smoother implementation more closely, I think it would be helpful to replace this function with three functions in the `detail` namespace:
- `convertToKunzTrajectory(Interpolated traj, VectorXd maxVelocity, VectorXd maxAcceleration, double maxDeviation, double timestep)`
- `convertToKunzTrajectory(Spline traj, VectorXd maxVelocity, VectorXd maxAcceleration, double maxDeviation, double timestep)`
- `convertToSpline(Trajectory traj, double startTime, StateSpace stateSpace)`

Then, `computeOptimalTiming` should basically look like `computeParabolicTiming`.

brianhou:
Also, @jslee02 if you could take a look over this PR to check if there are any nits that we should fix before merging, that would be great! (I know you're super busy, so just whenever you get a chance 😄)

brianhou:
After discussing with @aditya-vk, I've renamed `OptimalRetimer` to `KunzRetimer`. This is just a cosmetic change, so I believe the tests should still pass.

However, one of the tests (`KunzRetimerTests.timingArbitraryMultipleWaypoints`) is quite slow, initially taking 8 seconds on my laptop with the original parameter settings. I brought it down to 3 seconds by tweaking the `maxDeviation` parameter, but I'd still like it to be below the 1 second mark.

@dqyi11 do those waypoint values come from somewhere? If we're just choosing arbitrary points, maybe we can choose arbitrary ones that we can retime more quickly 😅

dqyi11:
@brianhou Thanks! Yes. They are just arbitrarily chosen ones. I will improve the test speed before merging it tomorrow morning.

PR #453, opened 2018-08-06
Add B-spline trajectory

jslee02:
(This is a work-in-progress pull request to get early feedback.)

This PR adds `trajectory::BSpline`, implementation of B-spline trajectory, and necessary classes in `common` namespace.

Unlike `trajectory::Spline`, `trajectory::BSpline` guarantees class `C^k` continuity (`k - 1` times differentiable) over the whole trajectory where `k` is the order of B-spline.

`Eigen::Spline` (renamed to `aikido::common::BSpline`) and `Eigen::SplineTraits` are copied into AIKIDO code base with some modifications. The motivation of the adjustment is that Eigen's B-spline module doesn't allow changing control points after construction while we need this feature for trajectory optimization, which we will introduce soon.

Unit tests will be added before merging this PR.

***

### DIFF
--- a/include/aikido/common/SplineFwd.hpp
+++ b/include/aikido/common/SplineFwd.hpp
@@ -1,0 +1,172 @@
+// This file is part of Eigen, a lightweight C++ template library
+// for linear algebra.
+//
+// Copyright (C) 20010-2011 Hauke Heibel <hauke.heibel@gmail.com>
+//
+// This Source Code Form is subject to the terms of the Mozilla
+// Public License v. 2.0. If a copy of the MPL was not distributed
+// with this file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+#ifndef AIKIDO_COMMON_SPLINEFWD_HPP_
+#define AIKIDO_COMMON_SPLINEFWD_HPP_
+
+#include <Eigen/Core>
+
+namespace aikido {
+namespace common {
+
+template <typename Scalar, int Dim, int Degree = Eigen::Dynamic>
+class BSpline;
+
+template <typename SplineType, int DerivativeOrder = Eigen::Dynamic>
+struct SplineTraits
+{
+};
+
+/**
+ * \ingroup Splines_Module
+ * \brief Compile-time attributes of the Spline class for Dynamic degree.
+ **/
+template <typename _Scalar, int _Dim, int _Degree>
+struct SplineTraits<BSpline<_Scalar, _Dim, _Degree>, Eigen::Dynamic>
+{
+  typedef _Scalar Scalar; /*!< The spline curve's scalar type. */
+  enum
+  {
+    Dimension = _Dim /*!< The spline curve's dimension. */
+  };
+  enum
+  {
+    Degree = _Degree /*!< The spline curve's degree. */
+  };
+
+  enum
+  {
+    OrderAtCompileTime
+    = _Degree == Eigen::Dynamic
+          ? Eigen::Dynamic
+          : _Degree + 1 /*!< The spline curve's order at compile-time. */
+  };
+  enum
+  {
+    NumOfDerivativesAtCompileTime = OrderAtCompileTime /*!< The number of
+                                                          derivatives defined
+                                                          for the current
+                                                          spline. */
+  };
+
+  enum
+  {
+    DerivativeMemoryLayout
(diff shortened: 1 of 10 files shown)


include/aikido/common/SplineFwd.hpp
```diff
@@ -0,0 +1,172 @@
+// This file is part of Eigen, a lightweight C++ template library
+// for linear algebra.
+//
+// Copyright (C) 20010-2011 Hauke Heibel <hauke.heibel@gmail.com>
```
dqyi11: Are we going to update the head description?

jslee02: I think we should keep this license as we copied this code from Eigen, and will add our license on top of this license for the part we modify/add.

src/trajectory/BSpline.cpp
```diff
@@ -0,0 +1,499 @@
...
+
+  KnotVectorType knots(numKnots);
+  knots.head(static_cast<ControlPointVectorType::Index>(degree))
+      .setConstant(startTime);
+  knots.tail(static_cast<ControlPointVectorType::Index>(degree))
+      .setConstant(endTime);
+
+  const common::StepSequence internalKnots(
+      startTime, endTime, numControlPoints + 1u - degree, true);
+  ControlPointVectorType::Index index = 1;
```
dqyi11: why `index`==0 is not updated?

src/trajectory/BSpline.cpp
```diff
@@ -0,0 +1,499 @@
...
+
+  mStateSpace->expMap(values, state);
+}
+
+//==============================================================================
+void BSpline::evaluateDerivative(
+    double /*t*/, int derivative, Eigen::VectorXd& /*tangentVector*/) const
+{
+  if (derivative < 1)
+    throw std::logic_error("Derivative must be positive.");
```
dqyi11: shall we just use `std::argument_error`?

include/aikido/common/SplineFwd.hpp
```diff
@@ -0,0 +1,172 @@
...
+                       _Dim,
+                       Eigen::Dynamic,
+                       DerivativeMemoryLayout,
+                       _Dim,
+                       NumOfDerivativesAtCompileTime>
+      DerivativeType;
+};
+
+/** \brief 2D float B-spline with dynamic degree. */
+typedef BSpline<float, 2> BSpline2f;
```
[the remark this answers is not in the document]
jslee02: I'm unsure whether we want to fully change this code or keep the changes as minimum as possible. Let's leave this code as it is and then revisit.

include/aikido/trajectory/BSpline.hpp
```diff
@@ -0,0 +1,253 @@
...
+#include "aikido/statespace/StateSpace.hpp"
+#include "aikido/trajectory/Trajectory.hpp"
+
+namespace aikido {
+namespace trajectory {
+
+AIKIDO_DECLARE_POINTERS(BSpline)
+
+/// B-spline trajectory define in a \c StateSpace.
+class BSpline : public Trajectory
```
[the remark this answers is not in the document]
jslee02: I'm doubtful if it's worth. Despite the similarity of their names, there isn't much in common.

src/trajectory/BSpline.cpp
```diff
@@ -0,0 +1,499 @@
...
+std::unique_ptr<Trajectory> BSpline::clone() const
+{
+  return dart::common::make_unique<BSpline>(*this);
+}
+
+//==============================================================================
+std::size_t BSpline::getDegree() const
+{
+  if (mStateSpace->getDimension() == 0u)
+    return 0; // TODO: Better idea?
```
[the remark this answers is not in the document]
jslee02: It can happen when the dimension of state space is zero (e.g., `statespace::dart::WeldJoint`).

src/trajectory/BSpline.cpp
jslee02: This line should be fixed to `index = degree`. We use multiple knots of the number of degree at the start and end knots so that the start and end of the trajectory are placed at the first and last control points, respectively. Let me document this to be clear.

jslee02:
@dqyi11 I believe the comments are addressed. Could you do a second round review?

PR #457, opened 2018-08-10
Bugfix in spline evaluation

aditya-vk:
The spline is currently fit by assuming the start of the segment to be the origin of the state space. For instance, see [this](https://github.com/personalrobotics/aikido/blob/de4ea8b4b5e9c7d74a55a1f30764068cfea7d996/src/planner/parabolic/ParabolicUtil.cpp#L165). When saving the segment of the spline, we also store the start state of that segment which can later be used to compute any point on the spline with respect to the origin.

During [evaluation](https://github.com/personalrobotics/aikido/blob/de4ea8b4b5e9c7d74a55a1f30764068cfea7d996/src/trajectory/Spline.cpp#L114) of the spline, we query the relative state, compute the `expMap` and compose it with the start state of the segment. This has issues with statespaces that have both angular and translation components.

Consider `a`, `b` to be the tangent vectors at `A` and `B` respectively.

The relativeVector `b - a` which is given by `evaluatePolynomial()` is not `logMap(inverse(A)*B)`. So `compose(A, expMap(relativeVector))` is not what we want. We either want:

1. `expMap(relativeVector + a)` which will give us `B`.
2. We enforce relativeVector to be `inverse(A)*B` and therefore make `compose(A, expMap(relativeVector))` technically correct.

(1) can be achieved by the changes made in the `evaluate` function in `Spline.cpp`. This considers the zero vector to be the origin always. See this [block](https://github.com/personalrobotics/aikido/blob/f800e81ffdbffce9af9e6466b355f9dcd2d47205/src/trajectory/Spline.cpp#L132) on this PR.

(2) can be achieved by fitting the spline considering the constraints differently. See this [line](https://github.com/personalrobotics/aikido/blob/f800e81ffdbffce9af9e6466b355f9dcd2d47205/src/planner/parabolic/ParabolicUtil.cpp#L190) on this PR. 

I discussed the two with Gilwoo. @gilwoolee, please correct me if I am wrong:
The frame of reference is different in the two cases. In the former it is always the origin while in the second case it is the start state of the segment. Therefore if the start state of the segment has a rotational component, this affects the translation components of the states on the spline. For example if I was solving a spline between `a(0.5, 0.5, 0)` to `b(1.0, 1.0, 0)`, the first term being angle, when we solve a spline between `(0, 0, 0)` and  `inverse(A)*B`, the translation in `y` is also affected (since `A` is the frame of reference), while if we consider a spline between `(0, 0, 0)` and `(0.5, 0.5, 0)`, the translation in `y` is not affected.

In the current PR I made changes to implement (1) and left (2) commented out for now.

@sanjibac: This is following the behaviour we saw on the racecar example. Your insight would be useful :D
***

### DIFF
--- a/src/planner/parabolic/ParabolicUtil.cpp
+++ b/src/planner/parabolic/ParabolicUtil.cpp
@@ -120,6 +120,27 @@ bool checkStateSpace(const statespace::StateSpace* _stateSpace)
   }
 }
 
+Eigen::VectorXd getTangentVector(
+    statespace::ConstStateSpacePtr stateSpace,
+    const Eigen::VectorXd from,
+    const Eigen::VectorXd to)
+{
+  auto state = stateSpace->createState();
+
+  auto fromInverse = stateSpace->createState();
+  stateSpace->expMap(from, state);
+  stateSpace->getInverse(state, fromInverse);
+
+  auto toMinusFrom = stateSpace->createState();
+  stateSpace->expMap(to, state);
+  stateSpace->compose(fromInverse, state, toMinusFrom);
+
+  Eigen::VectorXd tangentVector;
+  stateSpace->logMap(toMinusFrom, tangentVector);
+
+  return tangentVector;
+}
+
 std::unique_ptr<aikido::trajectory::Spline> convertToSpline(
     const ParabolicRamp::DynamicPath& _inputPath,
     double _startTime,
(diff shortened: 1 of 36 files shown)


src/planner/parabolic/ParabolicUtil.cpp
```diff
@@ -120,6 +120,27 @@ bool checkStateSpace(const statespace::StateSpace* _stateSpace)
   }
 }
 
+Eigen::VectorXd getTangentVector(
```
brianhou: It seems that there is some overlap between this function and `GeodesicInterpolator::getTangentVector`. Maybe we should put both functions somewhere in `aikido::statespace`?

src/planner/parabolic/ParabolicUtil.cpp
```diff
@@ -120,6 +120,27 @@ bool checkStateSpace(const statespace::StateSpace* _stateSpace)
...
+    statespace::ConstStateSpacePtr stateSpace,
+    const Eigen::VectorXd from,
```
brianhou: Nit: should these be `const Eigen::VectorXd&`?

brianhou:
Let's get this PR up-to-date and merged!

brianhou:
Bump

aditya-vk:
Closing as this has diverged quite a bit from master. If the spline evaluation is still buggy, could consider this PR as reference.

Issue #365, opened 2018-04-05
Segment faults on 32bit machine due to Alignment issues from Eigen


dqyi11:
A few unit tests failed on 32bit machine due to segment faults. 
>The following tests FAILED:
>         14 - test_FrameDifferentiable (OTHER_FAULT)
>         21 - test_RnBoxConstraint (OTHER_FAULT)
>         22 - test_SE2BoxConstraint (OTHER_FAULT)
>         23 - test_RnConstantSampler (OTHER_FAULT)
>         25 - test_FiniteSampleable (OTHER_FAULT)
>         26 - test_CyclicSampleable (OTHER_FAULT)
>         33 - test_CartesianProductProjectable (OTHER_FAULT)
>         43 - test_BarrettHandKinematicSimulationPositionCommandExecutor (OTHER_FAULT)
>         52 - test_ParabolicTimer (OTHER_FAULT)
>         64 - test_VectorFieldPlanner (OTHER_FAULT)
>Errors while running CTest

They are caused by alignment issues from `Eigen`.
https://eigen.tuxfamily.org/dox/group__DenseMatrixManipulation__Alignement.html

dqyi11:
Resolved in PR #368

brianhou:
If I remember correctly, there are still alignment issues but @dqyi11 believes that they're in DART rather than AIKIDO at this point. If that's the case, maybe we should try to create a minimal test case that only uses DART to help with debugging.

I think we also concluded that since it's only (apparently) a problem on 32-bit machines, we shouldn't invest too much time into it. But now that @dqyi11 is really good at sniffing out those issues, it might be worth it for the two of you to set up a short debugging session to see if you can fix it quickly.

jslee02:
@brianhou Thanks for the recalling the story. In that case, I would like to leave this issue open until we actually figure out whether if this is AIKIDO's issue or not.

PR #459, opened 2018-08-17
Fix segment fault on 32-bit machine

dqyi11:
This PR fixes the last issue left in #365 
I have tested it on 32-bit machine. 

The previous code uses `createState()` of a state space to a `State` raw point, which causes problem in ownership and thus memory leak. 
The timilimit in the unit test is increased for successul rate.

***

- [ ] Document new methods and classes (N/A)

- [ ] Add unit test(s) for this change (N/A)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,7 @@
   * Changed interface for TrajectoryPostProcessor: [#341](https://github.com/personalrobotics/aikido/pull/341)
   * Planning calls with InverseKinematicsSampleable constraints explicitly set MetaSkeleton to solve IK with: [#379](https://github.com/personalrobotics/aikido/pull/379)
   * Added a kinodynamic timer that generates a time-optimal smooth trajectory without completely stopping at each waypoint: [#443](https://github.com/personalrobotics/aikido/pull/443)
+  * Fix segment fault on 32 bit machine in vector-field planner: [#459](https://github.com/personalrobotics/aikido/pull/459)
 
 * Robot
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,7 +50,7 @@
   * Changed interface for TrajectoryPostProcessor: [#341](https://github.com/personalrobotics/aikido/pull/341)
   * Planning calls with InverseKinematicsSampleable constraints explicitly set MetaSkeleton to solve IK with: [#379](https://github.com/personalrobotics/aikido/pull/379)
   * Added a kinodynamic timer that generates a time-optimal smooth trajectory without completely stopping at each waypoint: [#443](https://github.com/personalrobotics/aikido/pull/443)
-  * Fixed segment fault on 32 bit machine in vector-field planner: [#459](https://github.com/personalrobotics/aikido/pull/459)
+  * Fixed segmentation fault on 32-bit machines in vector-field planner: [#459](https://github.com/personalrobotics/aikido/pull/459)
 
 * Robot
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,7 +49,7 @@
   * Changed interface for TrajectoryPostProcessor: [#341](https://github.com/personalrobotics/aikido/pull/341)
   * Planning calls with InverseKinematicsSampleable constraints explicitly set MetaSkeleton to solve IK with: [#379](https://github.com/personalrobotics/aikido/pull/379)
   * Added a kinodynamic timer that generates a time-optimal smooth trajectory without completely stopping at each waypoint: [#443](https://github.com/personalrobotics/aikido/pull/443)
-  * Fix segment fault on 32 bit machine in vector-field planner: [#459](https://github.com/personalrobotics/aikido/pull/459)
+  * Fixed segment fault on 32 bit machine in vector-field planner: [#459](https://github.com/personalrobotics/aikido/pull/459)
 
 * Robot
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -49,6 +49,7 @@
   * Changed interface for TrajectoryPostProcessor: [#341](https://github.com/personalrobotics/aikido/pull/341)
   * Planning calls with InverseKinematicsSampleable constraints explicitly set MetaSkeleton to solve IK with: [#379](https://github.com/personalrobotics/aikido/pull/379)
   * Added a kinodynamic timer that generates a time-optimal smooth trajectory without completely stopping at each waypoint: [#443](https://github.com/personalrobotics/aikido/pull/443)
+  * Fix segment fault on 32 bit machine in vector-field planner: [#459](https://github.com/personalrobotics/aikido/pull/459)
```
jslee02: Nit: Fixed

src/planner/vectorfield/detail/VectorFieldIntegrator.cpp
```diff
@@ -54,13 +54,20 @@ VectorFieldIntegrator::VectorFieldIntegrator(
...
-  mLastEvaluationTime = 0.0;
-  mDimension = mVectorField->getStateSpace()->getDimension();
-  mState = mVectorField->getStateSpace()->createState();
+  // Do Nothing
+}
+
+//==============================================================================
+VectorFieldIntegrator::~VectorFieldIntegrator()
+{
+  mVectorField->getStateSpace()->freeState(mState);
```
jslee02: I suggest using `ScopedState` instead of we take the responsibility of the memory allocation/deallocation.

dqyi11:
I have tested this branch on a 32bit machine. All the previous tests are passed.
There are two new tests failed, which are `49 - test_JointAvoidanceConfigurationRanker (OTHER_FAULT)` and `50 - test_NominalConfigurationRanker (Failed)`.

The error received in running `test_JointAvoidanceConfigurationRanker` is:
> Running main() from gtest_main.cc
[==========] Running 2 tests from 1 test case.
[----------] Global test environment set-up.
[----------] 2 tests from JointAvoidanceConfigurationRankerTest
[ RUN      ] JointAvoidanceConfigurationRankerTest.Constructor
[       OK ] JointAvoidanceConfigurationRankerTest.Constructor (1 ms)
[ RUN      ] JointAvoidanceConfigurationRankerTest.OrderTest
test_JointAvoidanceConfigurationRanker: /usr/include/eigen3/Eigen/src/Core/DenseStorage.h:78: Eigen::internal::plain_array<T, Size, MatrixOrArrayOptions, 16>::plain_array() [with T = double; int Size = 2; int MatrixOrArrayOptions = 0]: Assertion `(reinterpret_cast<size_t>(eigen_unaligned_array_assert_workaround_gcc47(array)) & 0xf) == 0 && "this assertion is explained here: " "http://eigen.tuxfamily.org/dox-devel/group__TopicUnalignedArrayAssert.html" " **** READ THIS WEB PAGE !!! ****"' failed.
Aborted (core dumped)

The error received in running `test_NominalConfigurationRanker` is:
>Running main() from gtest_main.cc
[==========] Running 2 tests from 1 test case.
[----------] Global test environment set-up.
[----------] 2 tests from NominalConfigurationRankerTest
[ RUN      ] NominalConfigurationRankerTest.Constructor
[       OK ] NominalConfigurationRankerTest.Constructor (1 ms)
[ RUN      ] NominalConfigurationRankerTest.OrderTest
/home/daqing/catkin_ws/src/aikido/tests/distance/test_NominalConfigurationRanker.cpp:109: Failure
Value of: aikido::tests::CompareEigenMatrices(rankedState, jointPositions[i], EPS)
  Actual: false (Matrices differ in row 0, column 0: 0.29999999999999999 !=: 0.10000000000000001.)
Expected: true
/home/daqing/catkin_ws/src/aikido/tests/distance/test_NominalConfigurationRanker.cpp:109: Failure
Value of: aikido::tests::CompareEigenMatrices(rankedState, jointPositions[i], EPS)
  Actual: false (Matrices differ in row 0, column 0: 0.10000000000000001 !=: 0.20000000000000001.)
Expected: true
/home/daqing/catkin_ws/src/aikido/tests/distance/test_NominalConfigurationRanker.cpp:109: Failure
Value of: aikido::tests::CompareEigenMatrices(rankedState, jointPositions[i], EPS)
  Actual: false (Matrices differ in row 0, column 0: 0.20000000000000001 !=: 0.29999999999999999.)
Expected: true
[  FAILED  ] NominalConfigurationRankerTest.OrderTest (1 ms)
[----------] 2 tests from NominalConfigurationRankerTest (2 ms total)
[----------] Global test environment tear-down
[==========] 2 tests from 1 test case ran. (2 ms total)
[  PASSED  ] 1 test.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] NominalConfigurationRankerTest.OrderTest

@aditya-vk do you have any idea?

jslee02:
@dqyi11 Thanks for the fix! However, I'm questionable if we really want to support 32 bit platforms. Do we have any use case? If we decide to do so, I think we should have a CI setup so that we can continuously have surveillance on the failings. Otherwise, let's not bother 32-bit issues. @brianhou @gilwoolee @aditya-vk thoughts? :)

brianhou:
I don't particularly care about 32-bit platforms, but it seems like it might reveal otherwise hidden bugs (e.g. this mixup between `State`s vs `ScopedState`s). However, I also feel like our builds already take an incredibly long time. Maybe this is something we can run only on `master`?

dqyi11:
I think the failures in 32-bit machine are caused by multiple reasons. Most of them are related with memory operation. Fixing them could also improve the robustness of aikido on 64-bit machine. 

Maybe we can close this one. Because all the problems we found previously are addressed.
We will open new issues with explained reasons/details if we encounter in the future.

PR #462, opened 2018-10-10
Trajectory utility functions

dqyi11:
This PR adds a few utility functions for trajectory classes, which include:
(1) trajectory conversion;
(2) trajectory concatenation;
(3) extract partial trajectory; and
(4) find the closest state in a trajectory given a state.
Those functions are migrated from ADA development.
A few unit tests are added for AIKIDO.

The `convertToSpline` is moved here from `aikido::planner::parabolic`. 

Please suggest which milestone it shall be for? I will update the `CHANGELOD.md` accordingly.
***

### DIFF
--- a/include/aikido/trajectory/Spline.hpp
+++ b/include/aikido/trajectory/Spline.hpp
@@ -132,6 +132,21 @@ public:
       int _derivative,
       Eigen::VectorXd& _tangentVector) const;
 
+  double getSegmentDuration(std::size_t _index) const
+  {
+    return mSegments[_index].mDuration;
+  }
+
+  const Eigen::MatrixXd& getSegmentCoefficients(std::size_t _index) const
+  {
+    return mSegments[_index].mCoefficients;
+  }
+
+  aikido::statespace::StateSpace::State* getSegmentState(std::size_t _index) const
+  {
+    return mSegments[_index].mStartState;
+  }
+
 private:
   struct PolynomialSegment
   {
--- a/include/aikido/trajectory/Spline.hpp
+++ b/include/aikido/trajectory/Spline.hpp
@@ -27,8 +27,8 @@ class Spline : public Trajectory
 public:
   /// Constructs an empty trajectory.
   ///
-  /// \param _stateSpace state space this trajectory is defined in
-  /// \param _startTime start time of the trajectory
+  /// \param _stateSpace State space this trajectory is defined in
+  /// \param _startTime Start time of the trajectory
   Spline(statespace::ConstStateSpacePtr _stateSpace, double _startTime = 0.);
 
   virtual ~Spline();
@@ -54,9 +54,9 @@ public:
   /// higher-order continuity. It is the responsibility of the user to pass in
   /// continuous spline coefficients if continuity is desired.
   ///
-  /// \param _coefficients polynomial coefficients
-  /// \param _duration duration of this segment, must be positive
-  /// \param _startState start state of the segment
+  /// \param _coefficients Polynomial coefficients
+  /// \param _duration Duration of this segment, must be positive
+  /// \param _startState Start state of the segment
   void addSegment(
       const Eigen::MatrixXd& _coefficients,
       double _duration,
@@ -72,13 +72,13 @@ public:
   /// the overload of the function that accepts an explicit start state
   /// if it is possible to do so.
   ///
-  /// \param _coefficients polynomial coefficients
-  /// \param _duration duration of this segment, must be positive
+  /// \param _coefficients Polynomial coefficients
+  /// \param _duration Duration of this segment, must be positive
   void addSegment(const Eigen::MatrixXd& _coefficients, double _duration);
 
   /// Gets the number of segments in this spline.
   ///
-  /// \return number of segments in this spline
+  /// \return Number of segments in this spline
   std::size_t getNumSegments() const;
 
   // Documentation inherited.
@@ -112,20 +112,20 @@ public:
 
   /// Gets a waypoint.
   ///
-  /// \param _index waypoint index
-  /// \param[out] state of the waypoint at index \c _index
+  /// \param _index Waypoint index
+  /// \param[out] state State of the waypoint at index \c _index
   void getWaypoint(
       std::size_t _index, statespace::StateSpace::State* state) const;
 
   /// Gets the time of a waypoint.
-  /// \param _index waypoint index
-  /// \return time of the waypoint at index \c _index
+  /// \param _index Waypoint index
+  /// \return Time of the waypoint at index \c _index
   double getWaypointTime(std::size_t _index) const;
 
   /// Gets the derivative of a waypoint.
-  /// \param _index waypoint index
-  /// \param _derivative order of derivative
-  /// \param[out] _tangentVector output tangent vector in the local frame at
+  /// \param _index Waypoint index
+  /// \param _derivative Order of derivative
+  /// \param[out] _tangentVector Output tangent vector in the local frame at
   /// index \c _index
   void getWaypointDerivative(
       std::size_t _index,
@@ -133,18 +133,18 @@ public:
       Eigen::VectorXd& _tangentVector) const;
 
   /// Gets the duration of a segment.
-  /// \param _index segment index
-  /// \return duration of the segment at index \c _index
+  /// \param _index Segment index
+  /// \return Duration of the segment at index \c _index
   double getSegmentDuration(std::size_t _index) const;
 
   /// Gets the coefficients of a segment.
-  /// \param _index segment index
-  /// \return coefficients of the segment at index \c _index
+  /// \param _index Segment index
+  /// \return Coefficients of the segment at index \c _index
   const Eigen::MatrixXd& getSegmentCoefficients(std::size_t _index) const;
 
   /// Gets the start state of a segment.
-  /// \param _index segment index
-  /// \return start state of the segment at index \c _index
+  /// \param _index Segment index
+  /// \return Start state of the segment at index \c _index
   const aikido::statespace::StateSpace::State* getSegmentState(
       std::size_t _index) const;
 
--- a/include/aikido/trajectory/Spline.hpp
+++ b/include/aikido/trajectory/Spline.hpp
@@ -132,20 +132,21 @@ public:
       int _derivative,
       Eigen::VectorXd& _tangentVector) const;
 
-  double getSegmentDuration(std::size_t _index) const
-  {
-    return mSegments[_index].mDuration;
-  }
+  /// Gets the duration of a segment.
+  /// \param _index segment index
+  /// \return duration of the segment at index \c _index
+  double getSegmentDuration(std::size_t _index) const;
 
-  const Eigen::MatrixXd& getSegmentCoefficients(std::size_t _index) const
-  {
-    return mSegments[_index].mCoefficients;
-  }
+  /// Gets the coefficients of a segment.
+  /// \param _index segment index
+  /// \return coefficients of the segment at index \c _index
+  const Eigen::MatrixXd& getSegmentCoefficients(std::size_t _index) const;
 
-  aikido::statespace::StateSpace::State* getSegmentState(std::size_t _index) const
-  {
-    return mSegments[_index].mStartState;
-  }
+  /// Gets the start state of a segment.
+  /// \param _index segment index
+  /// \return start state of the segment at index \c _index
+  const aikido::statespace::StateSpace::State* getSegmentState(
+      std::size_t _index) const;
 
 private:
   struct PolynomialSegment
--- a/include/aikido/trajectory/Spline.hpp
+++ b/include/aikido/trajectory/Spline.hpp
@@ -145,7 +145,7 @@ public:
   /// Gets the start state of a segment.
   /// \param _index Segment index
   /// \return Start state of the segment at index \c _index
-  const aikido::statespace::StateSpace::State* getSegmentState(
+  const aikido::statespace::StateSpace::State* getSegmentStartState(
       std::size_t _index) const;
 
 private:
--- a/include/aikido/trajectory/util.hpp
+++ b/include/aikido/trajectory/util.hpp
@@ -1,32 +1,33 @@
 #ifndef AIKIDO_TRAJECTORY_UTIL_HPP_
 #define AIKIDO_TRAJECTORY_UTIL_HPP_
 
-#include <aikido/trajectory/Interpolated.hpp>
-#include <aikido/trajectory/Spline.hpp>
+#include "aikido/trajectory/Interpolated.hpp"
+#include "aikido/trajectory/Spline.hpp"
 
 namespace aikido {
 namespace trajectory {
 
-/// Converts an interpolated trajectory to a piecewise linear spline trajectory
+/// Converts an interpolated trajectory to a piecewise linear spline trajectory.
+///
 /// This function requires the \c _inputTrajectory to use a \c
-/// GeodesicInterpolator.
-/// So the conversion exactly preserves the geometric path
-/// \param[in] _inputTrajectory Interpolated trajectory
-/// \return a spline trajectory
-aikido::trajectory::UniqueSplinePtr convertToSpline(
-    const aikido::trajectory::Interpolated& _inputTrajectory);
+/// GeodesicInterpolator. So the conversion exactly preserves the geometric
+/// path.
+///
+/// \param[in] inputTrajectory Interpolated trajectory
+/// \return A spline trajectory
+UniqueSplinePtr convertToSpline(const Interpolated& inputTrajectory);
 
 /// Converts a piecewise linear spline trajectory to an interpolated trajectory
 /// using a given interpolator.
 /// \param[in] traj Spline trajectory
 /// \param[in] interpolator Interpolator used in connecting two ends of a
 /// segment
-/// \return an interpolated trajectory
-aikido::trajectory::UniqueInterpolatedPtr convertToInterpolated(
-    const aikido::trajectory::Spline& traj,
-    aikido::statespace::ConstInterpolatorPtr& interpolator);
+/// \return An interpolated trajectory
+UniqueInterpolatedPtr convertToInterpolated(
+    const Spline& traj, statespace::ConstInterpolatorPtr& interpolator);
 
-/// Concatenates two interpolated trajectories
+/// Concatenates two interpolated trajectories.
+///
 /// This function concatenates two interpolated trajectories into one
 /// interpolated trajectory. The start state of the last segment in the first
 /// trajectory is connected with the start of the first segment in the second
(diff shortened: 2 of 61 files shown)


src/trajectory/util.cpp
```diff
@@ -0,0 +1,318 @@
...
+  auto outputTrajectory = make_unique<aikido::trajectory::Spline>(
+      stateSpace, traj.getStartTime());
+
+  double currSegmentStartTime = traj.getStartTime();
+  double currSegmentEndTime = currSegmentStartTime;
+  std::size_t currSegmentIdx = 0;
+
+  auto segmentStartState = stateSpace->createState();
+  auto segmentEndState = stateSpace->createState();
+  Eigen::VectorXd segStartPos(dimension), segEndPos(dimension),
```
gilwoolee: Nit: Let's make it three lines, one per variable.

src/trajectory/util.cpp
```diff
@@ -0,0 +1,318 @@
...
+    const aikido::trajectory::Spline& traj, double partialStartTime)
+{
+  if (partialStartTime < traj.getStartTime()
+      || partialStartTime > traj.getEndTime())
+    throw std::runtime_error("Wrong partial start time");
+
+  auto stateSpace = traj.getStateSpace();
+  const std::size_t dimension = stateSpace->getDimension();
+  auto outputTrajectory = make_unique<aikido::trajectory::Spline>(
+      stateSpace, traj.getStartTime());
```
gilwoolee: Why does it use `traj.getStartTime()` instead of `partialStartTime`?

include/aikido/trajectory/util.hpp
```diff
@@ -0,0 +1,66 @@
...
+    const aikido::trajectory::Interpolated& _inputTrajectory);
+
+/// Convert a piecewise linear spline trajectory to an interpolated trajectory
+/// using a given interpolator.
+/// \param[in] spline trajectory
+/// \param[in] interpolator used in connecting two ends of a segment
+/// \return an interpolated trajectory
+std::unique_ptr<aikido::trajectory::Interpolated> convertToInterpolated(
+    const aikido::trajectory::Spline& traj,
+    const aikido::statespace::InterpolatorPtr interpolator);
```
gilwoolee: Can we change this to `ConstInterpolatedPtr&`?

include/aikido/trajectory/util.hpp
```diff
@@ -0,0 +1,66 @@
...
+
+/// Concatenate two spline trajectories
+/// This function converts two spline trajectories into interpolated
+/// trajectories
+/// (geodesic interpolator) for concatenation, then convert the concatenated
```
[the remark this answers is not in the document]
dqyi11: Yes. I originally only said it concatenated two spline trajectories. But later I added more details because I want any future user will be aware of how it works. Because there are a few different strategies in concatenation. This is one of those that works well with the specific need in the servo control.

include/aikido/trajectory/util.hpp
```diff
@@ -0,0 +1,66 @@
...
+/// interpolated to spline. The start state of the last segment in the first
+/// trajectory is connected with the start of the first segment in the second
+/// trajectory in concatenation.
```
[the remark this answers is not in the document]
dqyi11: Not in this implementation. We do not enforce the match. Because in most of the cases, the end of the first does not match the start of the second. Maybe I should not name it `concatenate`. Do you think `merge` is better? Instead of connecting the end of the last segment with the start of the second, it directly connects the start of the last segment.

include/aikido/trajectory/util.hpp
```diff
@@ -0,0 +1,66 @@
...
+/// \param[in] input trajectory
+/// \param[in] reference state
+/// \param[in] time step in finding the closest state
+/// \return the time of the closest state on the input trajectory
+double findTimeOfClosetStateOnTrajectory(
+    const aikido::trajectory::Trajectory* traj,
+    const Eigen::VectorXd& refenceState,
+    double timeStep = 0.01);
+
+/// Create a partial Spline trajectory by a given start time.
```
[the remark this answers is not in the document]
dqyi11: Great catch! There is no exception handling now. I will add validation so that an error will be thrown when the input is not in the interval.

src/trajectory/util.cpp
dqyi11: It is like cutting a partial trajectory from a given start to the end; then shifting it to the same start time.

include/aikido/trajectory/util.hpp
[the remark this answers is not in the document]
gilwoolee: So we concatenate the two (and interpolate), with the time interval between the two points the same as `dt`? Then I think it may be nice to have a tolerance param to check for maximum distance between the two points so that two widely different points do not concatenate. I'm fine with the current name as long as it's explained clearly.

src/trajectory/util.cpp
```diff
@@ -0,0 +1,320 @@
...
+        traj.getSegmentDuration(i),
+        traj.getSegmentState(i));
+  }
+  return outputTrajectory;
+}
+
+//==============================================================================
+aikido::trajectory::UniqueInterpolatedPtr convertToInterpolated(
+    const aikido::trajectory::Spline& traj,
+    aikido::statespace::ConstInterpolatorPtr& interpolator)
```
jslee02: Pass `interpolator` by value and move it to the constructor of `Interpolated` as:
```cpp
  auto outputTrajectory
      = make_unique<aikido::trajectory::Interpolated>(stateSpace, std::move(interpolator));
```

The rationale is as follows:
1. No pass by reference: We don't change the pointer (but share the ownership).
2. Pass by reference: This is helpful to indicate that this function takes the ownership.

FYI, here is the [convention](https://github.com/personalrobotics/aikido/blob/master/STYLE.md#smart-pointers).

I understand this is a bit subtle, but let's try to be consistent with other code.

src/trajectory/util.cpp
```diff
@@ -0,0 +1,320 @@
...
+    outputTrajectory->addWaypoint(t, state);
+  }
+
+  return outputTrajectory;
+}
+
+//==============================================================================
+aikido::trajectory::UniqueInterpolatedPtr concatenate(
+    const aikido::trajectory::Interpolated& traj1,
+    const aikido::trajectory::Interpolated& traj2)
```
jslee02: This function isn't exposed at the header.

src/trajectory/util.cpp
```diff
@@ -0,0 +1,320 @@
...
+  return outputTrajectory;
+}
+
+//==============================================================================
+aikido::trajectory::UniqueSplinePtr concatenate(
+    const aikido::trajectory::Spline& traj1,
+    const aikido::trajectory::Spline& traj2)
+{
+  if (traj1.getStateSpace() != traj2.getStateSpace())
+    throw std::runtime_error("State space mismatch");
```
jslee02: `concatenate()` for `Interpolated` just checks the dimension while this function checks the instances.

@gilwoolee @brianhou Hm, what was the correct way of checking the compatibility of state spaces?

include/aikido/trajectory/util.hpp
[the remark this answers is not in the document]
jslee02: Looking at the implementation, it seems the last waypoint of the first trajectory is ignored and the first waypoint of the second trajectory is used at the end time of the first trajectory.

For example, the following trajecoty
```
traj1: wp1(t=0), wp2(t=1.1), wp3(t=3.0)
traj2: wp4(t=0), wp5(t=5.0)
```
becomes
```
new_traj: wp1(t=0), wp2(t=1.1), wp4(t=3.0), wp5(t=8.0).
```
so that the total time of the new trajectory is the sum of the durations of the first and second trajectories.

We may want to be able to set the policy of which waypoint is used for the "junction waypoint", but the current implementation would be okay as long as it's clearly documented.

tests/trajectory/test_SplineTrajectory.cpp
```diff
@@ -354,3 +355,68 @@ TEST_F(SplineTest, EvaluateDerivative_HigherOrder_ReturnsZero)
   trajectory.evaluateDerivative(7.5, 3, tangentVector);
   EXPECT_TRUE(Vector2d::Zero().isApprox(tangentVector));
 }
+
+TEST_F(SplineTest, findTimeOfClosetStateOnTrajectory)
```
jslee02: Nit: `TEST_F(SplineTest, FindTimeOfClosetStateOnTrajectory)`

tests/trajectory/test_SplineTrajectory.cpp
```diff
@@ -354,3 +355,68 @@ TEST_F(SplineTest, EvaluateDerivative_HigherOrder_ReturnsZero)
...
+  EXPECT_TRUE(Vector2d(1., 2.).isApprox(state.getValue()));
+
+  newTraj->evaluate(1., state);
+  EXPECT_TRUE(Vector2d(3., 6.).isApprox(state.getValue()));
+
+  newTraj->evaluate(2., state);
+  EXPECT_TRUE(Vector2d(6., 10.).isApprox(state.getValue()));
+}
+
+TEST_F(SplineTest, createPartialTrajectory)
```
jslee02: Nit: `TEST_F(SplineTest, CreatePartialTrajectory)`

include/aikido/trajectory/util.hpp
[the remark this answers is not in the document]
dqyi11: Yes. It is exactly like what @jslee02 described. I will use this example to explain.

src/trajectory/util.cpp
gilwoolee: Since we're merging the two trajectories in this case, shouldn't this check for instance match?

jslee02:
@gilwoolee I believe all the comments are addressed. Could you take a look at this PR when you have a chance?

include/aikido/trajectory/Spline.hpp
```diff
@@ -132,6 +132,22 @@ class Spline : public Trajectory
...
+
+  /// Gets the coefficients of a segment.
+  /// \param _index segment index
+  /// \return coefficients of the segment at index \c _index
+  const Eigen::MatrixXd& getSegmentCoefficients(std::size_t _index) const;
+
+  /// Gets the start state of a segment.
+  /// \param _index segment index
+  /// \return start state of the segment at index \c _index
+  const aikido::statespace::StateSpace::State* getSegmentState(
```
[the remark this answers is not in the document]
brianhou: I think this still needs to be renamed?

include/aikido/trajectory/util.hpp
```diff
@@ -0,0 +1,93 @@
...
+/// GeodesicInterpolator. So the conversion exactly preserves the geometric
+/// path.
+///
+/// \param[in] inputTrajectory Interpolated trajectory
+/// \return A spline trajectory
+UniqueSplinePtr convertToSpline(const Interpolated& inputTrajectory);
+
+/// Converts a piecewise linear spline trajectory to an interpolated trajectory
+/// using a given interpolator.
+/// \param[in] traj Spline trajectory
```
brianhou: Nit: newline between the description and parameter list?

src/trajectory/util.cpp
```diff
@@ -0,0 +1,314 @@
...
+        "The interpolator of inputTrajectory should be a GeodesicInterpolator");
+  }
+
+  const auto stateSpace = inputTrajectory.getStateSpace();
+  const auto dimension = stateSpace->getDimension();
+  const auto numWaypoints = inputTrajectory.getNumWaypoints();
+
+  if (!checkStateSpace(stateSpace.get()))
+    throw std::invalid_argument(
+        "computeParabolicTiming only supports Rn, "
```
brianhou:
```suggestion
        "convertToSpline only supports Rn, "
```

src/trajectory/util.cpp
```diff
@@ -0,0 +1,314 @@
...
+  }
+
+  return outputTrajectory;
+}
+
+//==============================================================================
+UniqueInterpolatedPtr concatenate(
+    const Interpolated& traj1, const Interpolated& traj2)
+{
+  if (traj1.getStateSpace()->getDimension()
```
brianhou: Should we check that the trajectories' state spaces and interpolators are the same instance, rather than just dimension?

src/trajectory/util.cpp
```diff
@@ -0,0 +1,314 @@
...
+      != traj2.getStateSpace()->getDimension())
+    throw std::runtime_error("Dimension mismatch");
+
+  auto stateSpace = traj1.getStateSpace();
+
+  auto outputTrajectory
+      = make_unique<Interpolated>(stateSpace, traj1.getInterpolator());
```
brianhou: Is there a reason to define `stateSpace` above rather than just calling it here?

src/trajectory/util.cpp
```diff
@@ -0,0 +1,314 @@
...
+
+  for (std::size_t i = 0; i < traj1.getNumWaypoints() - 1; i++)
```
brianhou:
```suggestion
  for (std::size_t i = 0; i < traj1.getNumWaypoints() - 1; ++i)
```

src/trajectory/util.cpp
```diff
@@ -0,0 +1,314 @@
...
+  {
+    outputTrajectory->addWaypoint(
+        traj1.getWaypointTime(i), traj1.getWaypoint(i));
+  }
+  outputTrajectory->addWaypoint(traj1.getEndTime(), traj2.getWaypoint(0));
```
brianhou: Why is this outside the loop below? What's wrong with starting that loop at 0?

src/trajectory/util.cpp
```diff
@@ -0,0 +1,314 @@
...
+  double endTimeOfTraj1 = traj1.getEndTime();
+  for (std::size_t i = 1; i < traj2.getNumWaypoints(); i++)
```
brianhou:
```suggestion
  for (std::size_t i = 1; i < traj2.getNumWaypoints(); ++i)
```

include/aikido/trajectory/util.hpp
```diff
@@ -0,0 +1,93 @@
...
+    const Spline& traj, statespace::ConstInterpolatorPtr interpolator);
+
+/// Concatenates two interpolated trajectories.
+///
+/// This function concatenates two interpolated trajectories into one
+/// interpolated trajectory. The start state of the last segment in the first
+/// trajectory is connected with the start of the first segment in the second
+/// trajectory in concatenation. For example, concatenating trajectory a:
+/// [wp1(t=0), wp2(t=1.1), wp3(t=3.0)] and trajectory b: [wp4(t=0), wp5(t=5.0)]
+/// gets a new trajectory: [wp1(t=0), wp2(t=1.1), wp3'(t=3.0), wp5(t=8.0)].
```
brianhou: I found this description confusing. Based on the code, it seems like this is actually `[wp1(t=0), wp2(t=1.1), wp4(t=3.0), wp5(t=8.0)]`.

src/trajectory/util.cpp
```diff
@@ -0,0 +1,314 @@
...
+}
+
+//==============================================================================
+UniqueSplinePtr concatenate(const Spline& traj1, const Spline& traj2)
+{
+  if (traj1.getStateSpace() != traj2.getStateSpace())
+    throw std::runtime_error("State space mismatch");
+  auto stateSpace = traj1.getStateSpace();
+  statespace::ConstInterpolatorPtr interpolator
+      = std::make_shared<statespace::GeodesicInterpolator>(stateSpace);
```
brianhou: Same as in `concatenate(Interpolated, Interpolated)`: is there a reason to define `stateSpace` above rather than just calling it here?

src/trajectory/util.cpp
```diff
@@ -0,0 +1,314 @@
...
+    const auto& coefficients = spline.getCoefficients().front();
+    outputTrajectory->addSegment(
+        coefficients, nextTime - currentTime, currentState);
+  }
+
+  return outputTrajectory;
+}
+
+//==============================================================================
+double findTimeOfClosetStateOnTrajectory(
```
brianhou: Nit: should the order of the implementations match the order of declarations?

include/aikido/trajectory/util.hpp
```diff
@@ -0,0 +1,93 @@
...
+/// are the start time and end time of the trajectory, respectively. The two
+/// endpoints are always included in the check. If two states are equidistant to
+/// the reference state, then the first one along the time-dimension will be
+/// returned.
+///
+/// \param[in] traj Input trajectory
+/// \param[in] state Reference state
+/// \param[in] timeStep Time step in finding the closest state
+/// \return The time of the closest state on the input trajectory
+double findTimeOfClosetStateOnTrajectory(
```
brianhou:
```suggestion
double findTimeOfClosestStateOnTrajectory(
```

include/aikido/trajectory/util.hpp
```diff
@@ -0,0 +1,93 @@
...
+    const Trajectory& traj,
+    const Eigen::VectorXd& refenceState,
```
brianhou:
```suggestion
    const Eigen::VectorXd& referenceState,
```

include/aikido/trajectory/util.hpp
```diff
@@ -0,0 +1,93 @@
...
+/// Finds the time of the closest state on a trajectory to a given state.
+///
+/// This function checks discrete states on [t0, t1, ..., tn] where t0 and tn
+/// are the start time and end time of the trajectory, respectively. The two
+/// endpoints are always included in the check. If two states are equidistant to
+/// the reference state, then the first one along the time-dimension will be
+/// returned.
+///
+/// \param[in] traj Input trajectory
+/// \param[in] state Reference state
```
brianhou:
```suggestion
/// \param[in] referenceState Reference state
```

include/aikido/trajectory/util.hpp
```diff
@@ -0,0 +1,93 @@
...
+/// \return The time of the closest state on the input trajectory
+double findTimeOfClosetStateOnTrajectory(
+    const Trajectory& traj,
+    const Eigen::VectorXd& refenceState,
+    double timeStep = 0.01);
+
+/// Retrieves part of a given spline trajectory
+///
+/// Given a spline trajectory \c traj , between [startTime , endTime] and a time
+/// point tPoint, such that startTime <= tPoint <= endTime, retrieves the part
```
brianhou: `tPoint` -> `\c partialStartTime`

include/aikido/trajectory/util.hpp
```diff
@@ -0,0 +1,93 @@
...
+/// of \c traj between tPoint and endTime. The retrieved spline is shifted in
+/// time to begin at startTime instead of at tPoint.
+///
+/// \param[in] traj Original spline trajectory
+/// \param[in] partialStartTime Start time of the new trajectory
+/// \throw If \c partialStartTime in not in the interval defined by the
```
brianhou:
```suggestion
/// \throw If \c partialStartTime is not in the interval defined by the
```

include/aikido/trajectory/util.hpp
```diff
@@ -0,0 +1,93 @@
...
+/// \param[in] timeStep Time step in finding the closest state
+/// \return The time of the closest state on the input trajectory
+double findTimeOfClosetStateOnTrajectory(
+    const Trajectory& traj,
+    const Eigen::VectorXd& refenceState,
+    double timeStep = 0.01);
+
+/// Retrieves part of a given spline trajectory
+///
+/// Given a spline trajectory \c traj , between [startTime , endTime] and a time
```
brianhou:
```suggestion
/// Given a spline trajectory \c traj between [startTime, endTime] and a time
```

include/aikido/trajectory/util.hpp
```diff
@@ -0,0 +1,93 @@
...
+/// \throw If \c partialStartTime in not in the interval defined by the
+/// trajectory start time and end time.
+/// \return The new partial trajectory
+UniqueSplinePtr createPartialTrajectory(
+    const Spline& traj, double partialStartTime);
+
+} // namespace trajectory
+} // namespace aikido
+
+#endif
```
brianhou: Nit: `#endif ...`

src/trajectory/util.cpp
```diff
@@ -0,0 +1,314 @@
...
+            coefficients, segmentDuration, segmentStartState);
+      }
+      break;
+    }
+
+    currSegmentIdx++;
+    currSegmentStartTime = currSegmentEndTime;
+  }
+
+  for (std::size_t i = currSegmentIdx + 1u; i < traj.getNumSegments(); i++)
```
brianhou:
```suggestion
  for (std::size_t i = currSegmentIdx + 1u; i < traj.getNumSegments(); ++i)
```

src/trajectory/util.cpp
```diff
@@ -0,0 +1,314 @@
...
+//==============================================================================
+UniqueInterpolatedPtr convertToInterpolated(
+    const Spline& traj, statespace::ConstInterpolatorPtr interpolator)
+{
+  auto stateSpace = traj.getStateSpace();
+  auto outputTrajectory
+      = make_unique<Interpolated>(stateSpace, std::move(interpolator));
+
+  auto state = stateSpace->createState();
+  for (std::size_t i = 0; i < traj.getNumWaypoints(); i++)
```
brianhou:
```suggestion
  for (std::size_t i = 0; i < traj.getNumWaypoints(); ++i)
```

src/trajectory/util.cpp
jslee02: I don't know why but at least it's not identical if we put this line into the loop with starting from 0 index. More importantly, this logic will not work as expected (traj1.duration + traj2.duration != newTraj.duration) when the start time of traj2 is not zero. Let me fix this and add a corresponding test for it.

jslee02:
Let me know if this PR is good to go!

src/trajectory/util.cpp
brianhou: Yep, I agree that the new fix respects a nonzero start time in traj2 better!

include/aikido/trajectory/util.hpp
```diff
@@ -0,0 +1,95 @@
...
+    const Spline& traj, statespace::ConstInterpolatorPtr interpolator);
+
+/// Concatenates two interpolated trajectories.
+///
+/// This function concatenates two interpolated trajectories into one
+/// interpolated trajectory. The start state of the last segment in the first
+/// trajectory is connected with the start of the first segment in the second
+/// trajectory in concatenation. For example, concatenating trajectory a:
+/// [wp1(t=1), wp2(t=2.1), wp3(t=4)] and trajectory b: [wp4(t=2), wp5(t=7)]
+/// gets a new trajectory: [wp1(t=0), wp2(t=1.1), wp4(t=4), wp5(t=9)].
```
brianhou: I think this implementation keeps the original start time of trajectory a? So I think it should look like

```suggestion
/// gets a new trajectory: [wp1(t=1), wp2(t=2.1), wp4(t=4), wp5(t=9)].
```

jslee02: That's correct. I thought it's updated so but wasn't. :rage: Thanks!

PR #466, opened 2018-10-27
API to use OMPL Planners within the new planner framework

aditya-vk:
This PR enables using the OMPL planners via AIKIDO for any of the `ConfigurationToConfiguration` problems. This should allow us to plan for exact/approximate solutions for configuration(s)/TSR problems.

There are a few TODOs in the PR, one of which would be part of another PR perhaps(?) since they are in relation to statespaces and contraints.

***


include/aikido/planner/ompl/OMPLConfigurationToConfigurationPlanner.hpp
```diff
@@ -0,0 +1,111 @@
...
+#include "aikido/planner/ConfigurationToConfigurationPlanner.hpp"
+#include "aikido/planner/ompl/GeometricStateSpace.hpp"
+#include "aikido/statespace/StateSpace.hpp"
+
+#include <ompl/base/Planner.h>
+#include <ompl/base/ProblemDefinition.h>
+#include <ompl/base/ScopedState.h>
+#include <ompl/base/SpaceInformation.h>
+#include <ompl/base/goals/GoalRegion.h>
+#include <ompl/geometric/PathSimplifier.h>
```
jslee02: Nit: Third party headers should come first and then the project headers (see [this](https://github.com/personalrobotics/aikido/blob/master/STYLE.md#header-style))

include/aikido/planner/ompl/OMPLConfigurationToConfigurationPlanner.hpp
```diff
@@ -0,0 +1,111 @@
...
+  double mMaxDistanceBtwValidityChecks;
+};
+
+} // namespace ompl
+} // namespace planner
+} // namespace aikido
+
+#include "aikido/planner/ompl/detail/OMPLConfigurationToConfigurationPlanner-impl.hpp"
+
+#endif // AIKIDO_PLANNER_SNAPCONFIGURATIONTOCONFIGURATIONPLANNER_HPP_
```
jslee02:
```suggestion
#endif // AIKIDO_PLANNER_OMPL_OMPLCONFIGURATIONTOCONFIGURATIONPLANNER_HPP_
```

src/robot/util.cpp
```diff
@@ -99,18 +104,25 @@ trajectory::TrajectoryPtr planToConfiguration(
...
+  //      goalState,
+  //      space,
+  //      std::make_shared<GeodesicInterpolator>(space),
+  //      createDistanceMetric(space),
+  //      createSampleableBounds(space, rng->clone()),
+  //      collisionTestable,
+  //      createTestableBounds(space),
+  //      createProjectableBounds(space),
+  //      timelimit,
+  //      collisionResolution);
```
jslee02: Nit: Let's remove old code

include/aikido/planner/ompl/detail/OMPLConfigurationToConfigurationPlanner-impl.hpp
```diff
@@ -0,0 +1,200 @@
...
+        = std::move(createProjectableBounds(metaskeletonStatespace));
+
+  // Geometric State space
+  auto sspace = ompl_make_shared<GeometricStateSpace>(
+      mStateSpace,
+      std::make_shared<aikido::statespace::GeodesicInterpolator>(mStateSpace),
+      std::move(dmetric),
+      std::move(sampler),
+      mBoundsConstraint,
+      std::move(mBoundsProjector));
```
jslee02: We don't need to have `mBoundsProjector` as a member variable because it's been moved. It seems we could convert this as a local variable in this constructor implementation.

include/aikido/planner/ompl/detail/OMPLConfigurationToConfigurationPlanner-impl.hpp
```diff
@@ -0,0 +1,200 @@
...
+  si->setStateValidityChecker(vchecker);
+
+  ::ompl::base::MotionValidatorPtr mvalidator
+      = ompl_make_shared<MotionValidator>(si, mMaxDistanceBtwValidityChecks);
+  si->setMotionValidator(mvalidator);
+
+  // Define the OMPL problem
+  auto pdef = ompl_make_shared<::ompl::base::ProblemDefinition>(si);
+  auto sspace
+      = ompl_static_pointer_cast<GeometricStateSpace>(si->getStateSpace());
```
jslee02: Nit: Will the type of `sspace` always be `GeometricStateSpace`? If so, let's add an assertion here for sure something like `assert(ompl_dynamic_pointer_cast<GeometricStateSpace>(si->getStateSpace()))`;

include/aikido/planner/ompl/detail/OMPLConfigurationToConfigurationPlanner-impl.hpp
```diff
@@ -0,0 +1,200 @@
...
+    {
+      throw std::invalid_argument(
+          "Path is not of type PathGeometric. Cannot convert to aikido "
+          "Trajectory.");
+    }
+
+    for (std::size_t idx = 0; idx < path->getStateCount(); ++idx)
+    {
+      const auto* st
+          = static_cast<aikido::planner::ompl::GeometricStateSpace::StateType*>(
```
jslee02: Nit: Let's add an assertion for checking the type using dynamic_cast.

src/robot/util.cpp
```diff
@@ -73,6 +76,8 @@ trajectory::TrajectoryPtr planToConfiguration(
     RNG* rng,
     double timelimit)
 {
+  DART_UNUSED(timelimit);
```
jslee02: Do we still need `timelimit` parameter here?

include/aikido/planner/ompl/OMPLConfigurationToConfigurationPlanner.hpp
```diff
@@ -0,0 +1,111 @@
...
+namespace ompl {
+
+template <class PlannerType>
+class OMPLConfigurationToConfigurationPlanner
+    : public aikido::planner::ConfigurationToConfigurationPlanner
+{
+public:
+  /// Constructor.
+  ///
+  /// \param[in] planner The OMPL planner to use.
```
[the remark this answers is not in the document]
brianhou: I think this line should be deleted now?

include/aikido/planner/ompl/OMPLConfigurationToConfigurationPlanner.hpp
```diff
@@ -0,0 +1,110 @@
...
+      distance::DistanceMetricPtr dmetric = nullptr,
+      constraint::SampleablePtr sampler = nullptr,
+      constraint::TestablePtr boundsConstraint = nullptr,
+      constraint::ProjectablePtr boundsProjector = nullptr,
+      double maxDistanceBtwValidityChecks = 0.1);
+
+  /// Plans a trajectory from start state to goal state by using an interpolator
+  /// to interpolate between them.
+  ///
+  /// The planner returns success if the resulting trajectory satisfies
```
brianhou: I don't think this is accurate: this function always returns the trajectory, which is null if planning fails.

include/aikido/planner/ompl/OMPLConfigurationToConfigurationPlanner.hpp
```diff
@@ -0,0 +1,110 @@
...
+  /// \throw If \c problem is not ConfigurationToConfiguration.
+  /// \throw If \c result is not ConfigurationToConfiguration::Result.
+  trajectory::TrajectoryPtr plan(
+      const SolvableProblem& problem, Result* result = nullptr) override;
+
+  /// Returns the underlying OMPL planner used.
+  ::ompl::base::PlannerPtr getOMPLPlanner();
+
+  /// Sets interpolator used to produce the output trajectory.
+  void setInterpolator(statespace::ConstInterpolatorPtr interpolator);
```
brianhou: Nit: missing documentation for `interpolator`.

brianhou: Is this function being used? What's the use case for wanting to change the interpolator after construction, without changing the other fields?

include/aikido/planner/ompl/OMPLConfigurationToConfigurationPlanner.hpp
```diff
@@ -0,0 +1,110 @@
...
+
+protected:
+  /// Pointer to the underlying OMPL Planner.
+  ::ompl::base::PlannerPtr mPlanner;
+
+  /// Interpolator used to produce the output trajectory.
+  statespace::ConstInterpolatorPtr mInterpolator;
+
+private:
+  // TODO (avk): Should we make the problem take std::vector<TestablePtr>?
```
brianhou: Why would we want a vector of `Testable`s?

include/aikido/planner/ompl/detail/OMPLConfigurationToConfigurationPlanner-impl.hpp
```diff
@@ -0,0 +1,207 @@
+#ifndef AIKIDO_PLANNER_OMPL_DETAIL_OMPLCONFIGURATIONTOCONFIGURATION_IMPL_HPP_
+#define AIKIDO_PLANNER_OMPL_DETAIL_OMPLCONFIGURATIONTOCONFIGURATION_IMPL_HPP_
+
+#include "aikido/planner/ompl/OMPLConfigurationToConfigurationPlanner.hpp"
+
```
brianhou: I think we try to avoid empty lines in the includes?

include/aikido/planner/ompl/detail/OMPLConfigurationToConfigurationPlanner-impl.hpp
```diff
@@ -0,0 +1,207 @@
...
+        statespace::ConstStateSpacePtr stateSpace,
+        common::RNG* rng,
+        statespace::ConstInterpolatorPtr interpolator,
+        distance::DistanceMetricPtr dmetric,
+        constraint::SampleablePtr sampler,
+        constraint::TestablePtr boundsConstraint,
+        constraint::ProjectablePtr boundsProjector,
+        double maxDistanceBtwValidityChecks)
+  : ConfigurationToConfigurationPlanner(std::move(stateSpace), rng)
+  , mInterpolator(std::move(interpolator))
```
brianhou: Why do we move the `interpolator` and `boundsConstraint` here, but not the `dmetric`, `sampler`, and `boundsProjector`? I don't have a particular opinion one way or another, but I think it should at least be consistent.

brianhou: Actually, maybe the root question is why are `interpolator` and `boundsConstraint` the only members? One argument seems that the OMPL planner includes references to all of these, but doesn't the OMPL planner use an interpolator somewhere?

include/aikido/planner/ompl/detail/OMPLConfigurationToConfigurationPlanner-impl.hpp
```diff
@@ -0,0 +1,207 @@
...
+{
+  if (!mInterpolator)
+    mInterpolator = std::make_shared<aikido::statespace::GeodesicInterpolator>(
+        mStateSpace);
+
+  if (!dmetric)
+    dmetric = std::move(createDistanceMetric(mStateSpace));
+
+  // TODO (avk): The constraint namespace functions only work with
+  // metaskeletonstatespace
```
brianhou:
```suggestion
  // MetaSkeletonStateSpaces
```

include/aikido/planner/ompl/detail/OMPLConfigurationToConfigurationPlanner-impl.hpp
```diff
@@ -0,0 +1,207 @@
...
+  const auto metaskeletonStatespace
+      = std::dynamic_pointer_cast<const MetaSkeletonStateSpace>(mStateSpace);
+
+  if (!metaskeletonStatespace
```
brianhou: I found this logic somewhat confusing because it wasn't clear to me when all of the conditions would be satisfied. I think this might be clearer:
```cpp
if (metaSkeletonStateSpace)
{
  if (!sampler) ...
  if (!mBoundsConstraint) ...
  if (!boundsProjector) ...
}
else
{
  if (!sampler || !boundsConstraint || !boundsProjector)
  {
    throw ...
  }
}

auto sspace = ...
```

include/aikido/planner/ompl/detail/OMPLConfigurationToConfigurationPlanner-impl.hpp
```diff
@@ -0,0 +1,207 @@
...
+    mBoundsConstraint = std::move(createTestableBounds(metaskeletonStatespace));
+
+  if (!boundsProjector)
+    boundsProjector
+        = std::move(createProjectableBounds(metaskeletonStatespace));
+
+  // Geometric State space
+  auto sspace = ompl_make_shared<GeometricStateSpace>(
+      mStateSpace,
+      std::make_shared<aikido::statespace::GeodesicInterpolator>(mStateSpace),
```
brianhou: Shouldn't this use `mInterpolator`? (Again, still not clear why that's a field and not just a local variable.)

include/aikido/planner/ompl/detail/OMPLConfigurationToConfigurationPlanner-impl.hpp
```diff
@@ -0,0 +1,207 @@
...
+  // ProblemDefinition clones states and keeps them internally
+  pdef->setStartAndGoalStates(start, goal);
+
+  sspace->freeState(start);
+  sspace->freeState(goal);
+
+  // Plan
+  mPlanner->setProblemDefinition(pdef);
+  mPlanner->setup();
+  auto solved = mPlanner->solve(1);
```
brianhou: Shouldn't `1` be `timelimit` or something?

include/aikido/planner/ompl/detail/OMPLConfigurationToConfigurationPlanner-impl.hpp
```diff
@@ -0,0 +1,207 @@
...
+      throw std::invalid_argument(
+          "Path is not of type PathGeometric. Cannot convert to aikido "
+          "Trajectory.");
+    }
+
+    for (std::size_t idx = 0; idx < path->getStateCount(); ++idx)
+    {
+      // Only states belonging geometric statespaces are supported.
+      assert(
+          dynamic_cast<aikido::planner::ompl::GeometricStateSpace::StateType*>(
```
brianhou: Is it possible for a `PathGeometric` to have non-`GeometricStateSpace::StateType`? If so, is it possible for a `PathGeometric` to have a mixture of `StateType`s (i.e. does this `dynamic_cast` need to happen every time)?

tests/planner/ompl/test_OMPLConfigurationToConfigurationPlanner.cpp
```diff
@@ -0,0 +1,193 @@
...
+          stateSpace,
+          nullptr,
+          interpolator,
+          std::move(dmetric),
+          std::move(sampler),
+          std::move(boundsConstraint),
+          std::move(boundsProjection),
+          0.1);
+  auto traj = planner->plan(problem);
+  EXPECT_TRUE(true);
```
brianhou: What?

include/aikido/planner/ompl/OMPLConfigurationToConfigurationPlanner.hpp
aditya-vk: I was also a little uncomfortable with needing to have an `mBoundsConstraint` to reconstruct the contraint during planning and was wondering if we could have a vector of testables passed in the problem instead of a single testable.

include/aikido/planner/ompl/detail/OMPLConfigurationToConfigurationPlanner-impl.hpp
aditya-vk: sigh

include/aikido/planner/ompl/detail/OMPLConfigurationToConfigurationPlanner-impl.hpp
```diff
@@ -0,0 +1,192 @@
...
+    {
+      throw std::invalid_argument(
+          "Path is not of type PathGeometric. Cannot convert to aikido "
+          "Trajectory.");
+    }
+
+    for (std::size_t idx = 0; idx < path->getStateCount(); ++idx)
+    {
+      const auto* st
+          = static_cast<aikido::planner::ompl::GeometricStateSpace::StateType*>(
```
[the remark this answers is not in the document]
aditya-vk: Done. For now, I don't much of a preference on this, since we are only working with geometric planners/statespaces. But I guess when we later extend our support, we might be dealing with other state types. 

Thanks!

Issue #471, opened 2018-11-30
Increase minimum required ompl version greater than 0.13.0


jslee02:
By #466, AIKIDO uses new API of ompl that are not available in OMPL 0.13. We need to increase the required minimum version of ompl accordingly.

I learned this looking at the build log of PPA: https://launchpadlibrarian.net/399640688/buildlog_ubuntu-trusty-amd64.aikido_0.3.0-378~201811300601~ubuntu14.04.1_BUILDING.txt.gz

aditya-vk:
Addressed in #472

PR #472, opened 2018-11-30
Specify minimum OMPL version

aditya-vk:
Addresses #471 

***

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -150,7 +150,7 @@ int main()
 DART_HAS_VOXELGRIDSHAPE
 )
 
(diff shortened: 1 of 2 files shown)


jslee02:
Unfortunately, OMPL package doesn't support version checking using `find_package()` until 1.4.0. That said, we instead have to check the version using the version variable of OMPL something like:
```cmake
find_package(ompl REQUIRED)
if(OMPL_VERSION VERSION_LESS 0.14.0)
  message(STATUS "warning message...")
  return()
endif()
```

jslee02:
Closing in favor of #473

PR #476, opened 2018-12-24
Demo/nips updates

gilwoolee:
This PR is to incorporate changes made for NIPS demo.
The branch has diverged a few months ago, so the plan is to just get the essential components into master.

It seems that the major additional features are the following:

- `Conversions.hpp` has `toRevoluteJointTrajectory`
- `PoseEstimator` has `getObjInfo`

@aditya-vk please take a pass on the files and clean up as necessary.

***


include/aikido/robot/ConcreteManipulator.hpp
```diff
@@ -42,6 +42,13 @@ class ConcreteManipulator : public Manipulator
       const dart::dynamics::MetaSkeletonPtr& metaSkeleton,
       const aikido::trajectory::Trajectory* path) override;
 
+  // Documentation inherited.
+  virtual std::unique_ptr<aikido::trajectory::Spline> retimePathWithKunzTimer(
```
aditya-vk: We should create an issue stating this is temporary and should be replaced with a commoon Postprocessor class that Hauser, Kunz derive from that can be passed into these functions.

include/aikido/robot/ConcreteRobot.hpp
```diff
@@ -199,7 +207,8 @@ class ConcreteRobot : public Robot
       const aikido::constraint::dart::TSRPtr& tsr,
       const aikido::constraint::dart::CollisionFreePtr& collisionFree,
       double timelimit,
-      std::size_t maxNumTrials);
+      std::size_t maxNumTrials,
+      const Eigen::VectorXd& nominalPosition = Eigen::VectorXd(0));
```
aditya-vk: Are we keeping this argument? I propose to not, and if we are, we need to create an issue to track it and remove it, and add docstring.

src/planner/parabolic/ParabolicUtil.cpp
```diff
@@ -128,11 +135,22 @@ std::unique_ptr<ParabolicRamp::DynamicPath> convertToDynamicPath(
...
+  auto stateSpace = _inputTrajectory.getStateSpace();
+  auto metaSkeletonStateSpace
+      = std::dynamic_pointer_cast<const MetaSkeletonStateSpace>(stateSpace);
+
   const auto numWaypoints = _inputTrajectory.getNumWaypoints();
 
+  ConstSplinePtr r1Trajectory;
+  if (metaSkeletonStateSpace)
+  {
+    r1Trajectory = toR1JointTrajectory(stateSpace, _inputTrajectory);
```
aditya-vk: we should do the same in Kunz timer also

aditya-vk: Should I add a check here to do this only if all joints are 1D and are either SO2 or R1?

gilwoolee:
Closing this PR as all relevant components have been separately merged by #482, #484, #496, #497, #503, #505.

PR #478, opened 2019-01-14
Update gtest version to 1.8.1

jslee02:
This also fixes the [build failure on Ubuntu Bionic](https://travis-ci.org/personalrobotics/aikido/jobs/479173065).

***


brianhou:
Were there any changes to code that we've written, or is the only change to update `gtest` code? If it's the latter, what's the rationale for not just using a `git submodule`?

brianhou:
Hmm, this actually makes me mildly concerned. What if we update gtest again in the future, and don't remember that that change (or other future changes we make) were necessary?

New proposal: what if we fork `gtest` into PRL, and apply our changes. We can then merge in new versions when desired, while maintaining our own changes. I'm somewhat fuzzy on exactly where `libherb` and other libraries that use `aikido` find `gtest`, but maybe that could be useful for those as well?

jslee02:
> Hmm, this actually makes me mildly concerned. What if we update gtest again in the future, and don't remember that that change (or other future changes we make) were necessary?

Agreed.

The proposal sounds reasonable to me.

As a note, `libherb` download gtest (using git) when it gets built instead of having the code inside the repo. I personally don't prefer this way because PPA doesn't allow to download code while building the package.

Anyways, I don't have any strong preference for this. Any method that is convenient to maintain would be great.

PR #480, opened 2019-01-14
Update for DART 6.7

jslee02:
Use recommended APIs instead of deprecated corresponding.

Note that this PR won't pass the CIs (for Ubuntu >= 14.04) until the issues of gtest and `make_unique()`. I breakdown #479 into smaller ones to make the review process, and this is one of them.

***

**Before creating a pull request**

- [x] Format code with `make format`

**Before merging a pull request**

- [x] Set version target by selecting a milestone on the right side
- [x] Summarize this change in `CHANGELOG.md`

PR #481, opened 2019-01-14
Fix compiling dart::common::make_unique() with recent GCC versions

jslee02:
Explicitly specify the namespace as `::dart::common::make_unique()` to avoid the ambiguity with `std::make_unique()`.

Note that this PR won't pass the CIs (for Ubuntu >= 14.04) until the issues of gtest and #480. I breakdown #479 into smaller ones to make the review process, and this is one of them.

***


brianhou:
My guess is that this is a problem when using a compiler that supports C++14 features -- is that correct?

jslee02:
> My guess is that this is a problem when using a compiler that supports C++14 features -- is that correct?

Exactly!

PR #483, opened 2019-01-16
Bugfix/memory leak

gilwoolee:
This PR fixes memory leak in `Interpolated::addWaypoint`, which uses `allocateState` but never frees. This PR adds destructor and calls `freeState` during destruction.

I tested with valgrind in test_ParabolicTimer and it reduces memory leak from 803 bytes to 51 bytes (the 51 bytes are coming from something else at the moment.)
***


src/trajectory/Interpolated.cpp
```diff
@@ -16,6 +16,13 @@ Interpolated::Interpolated(
   // Do nothing
 }
 
+//==============================================================================
+Interpolated::~Interpolated()
+{
+  for (auto waypoint : mWaypoints)
```
jslee02: Nit: to avoid unnecessary copy
```suggestion
  for (auto& waypoint : mWaypoints)
```

src/trajectory/Interpolated.cpp
```diff
@@ -210,5 +217,13 @@ bool Interpolated::Waypoint::operator<(double rhs) const
   return t < rhs;
 }
 
+//==============================================================================
+void Interpolated::Waypoint::free(
+    const statespace::ConstStateSpacePtr& stateSpace)
+{
+  if (state)
+    stateSpace->freeState(state);
```
jslee02: I just realized that this could double-free when this function is called multiple times. Let's prevent it by setting `state` to `nullptr` once it's freed as:
```cpp
if (state)
{
  stateSpace->freeState(state);
  state = nullptr;
}
```

gilwoolee: Oh, good catch! :)

PR #485, opened 2019-01-17
Use own build script for gtest

jslee02:
This PR reverts the customized gtest's build script and uses own build script. This is to simplify the process of updating gtest version without customizing it as discussed in #478.

Depends on #480.

***


tests/statespace/test_CartesianProduct.cpp
```diff
@@ -20,7 +20,11 @@ TEST(CartesianProduct, Clone)
   for (auto i = 0u; i < 5u; ++i)
   {
     auto s1 = space.createState();
+#if DART_VERSION_AT_LEAST(6,7,0)
```
brianhou: I think this needs to merge changes from #480 first.

jslee02: Done!

PR #489, opened 2019-01-22
Silence gcov output.

brianhou:
Some Travis builds error and fail as a result of huge log files. This should fix that problem.

***


brianhou:
It turns out that we were actually running gcov twice: once from `cmake/CodeCoverage.cmake` and once as part of the Codecov bash script. Now, we're skipping the second run (`-X gcov`) and using the output from the `CodeCoverage.cmake` run (`-f ...`).

One detail of this is that the report uploading isn't actually handled by `CodeCoverage.cmake`, so I've hard-coded the path in `.ci/script_linux.sh`. Eventually, we should fix this. There's also still room to improve with quieter gcov output -- I think @jslee02 started investigating a little bit in the `codecov_output` branch.

I think we should merge this branch first, so we can unblock other PRs (and remove at least one failure that we can control).

cmake/CodeCoverage.cmake
```diff
@@ -148,7 +148,6 @@ FUNCTION(SETUP_TARGET_FOR_COVERAGE _targetname _testrunner _outputname)
 		COMMAND ${LCOV_PATH} --directory . --capture --output-file ${coverage_info}
 		COMMAND ${LCOV_PATH} --remove ${coverage_info} 'tests/*' '/usr/*' --output-file ${coverage_cleaned}
 		COMMAND ${GENHTML_PATH} -o ${_outputname} ${coverage_cleaned}
-		COMMAND ${CMAKE_COMMAND} -E remove ${coverage_info} ${coverage_cleaned}
```
jslee02: How does this help? I just would like to learn.

brianhou: This step deletes the coverage output files, which can't be deleted until we upload to Codecov in the next step (after running `CodeCoverage.cmake`). Once we have the upload step here, I think we can restore this clean-up step.

PR #495, opened 2019-01-24
Upload to CodeCov from CodeCoverage.cmake

### LINKED ISSUE #491: Upload to CodeCov from CodeCoverage.cmake
Currently, `cmake/CodeCoverage.cmake` runs `lcov` to generate coverage data, but we rely on another shell command (from `.travis.yml`) to upload to the CodeCov service itself. We should take care of all of this in `CodeCoverage.cmake`.

Tangentially related: it might also be useful to use an updated version of [`CodeCoverage.cmake`](https://github.com/bilke/cmake-modules/blob/master/CodeCoverage.cmake)

brianhou:
Resolves #491.

### DIFF
--- a/cmake/CodeCoverage.cmake
+++ b/cmake/CodeCoverage.cmake
@@ -35,6 +35,9 @@
 # - Added support for Clang.
 # - Some additional usage instructions.
 #
+# 2019-01-24, Brian Hou
+# - Added support for Codecov uploads.
+#
 # USAGE:
 
 # 0. (Mac only) If you use Xcode 5.1 make sure to patch geninfo as described here:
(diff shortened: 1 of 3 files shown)


cmake/CodeCoverage.cmake
```diff
@@ -149,6 +151,14 @@ FUNCTION(SETUP_TARGET_FOR_COVERAGE _targetname _testrunner _outputname)
 		COMMAND ${LCOV_PATH} --remove ${coverage_info} 'tests/*' '/usr/*' --output-file ${coverage_cleaned}
 		COMMAND ${GENHTML_PATH} -o ${_outputname} ${coverage_cleaned}
 
+		# Upload to Codecov (added by @brianhou)
```
jslee02: It would be good to add this change to the [changlog at the top of this file](https://github.com/personalrobotics/aikido/pull/495/files#diff-c647cf0f727c3b16f40757169b38502cL37).

PR #496, opened 2019-01-27
SO2 to R1 conversion for postprocessing

aditya-vk:
The timers assume that the path representation is in R1. When the joints in the skeleton are SO2, we would like them to move along the geodesic and not the other way around. To ensure this happens, we need to appropriately convert the path representation to R1 along the geodesic. THis PR handles this conversion.

***


src/trajectory/util.cpp
```diff
@@ -315,5 +368,72 @@ UniqueSplinePtr createPartialTrajectory(
...
+    subspaces.emplace_back(std::make_shared<const R1>());
+
+  auto rSpace = std::make_shared<CartesianProduct>(subspaces);
+  auto rInterpolator = std::make_shared<GeodesicInterpolator>(rSpace);
+  auto rTrajectory = std::make_shared<Interpolated>(rSpace, rInterpolator);
+
+  Eigen::VectorXd sourceVector(space->getDimension());
+  auto sourceState = rSpace->createState();
+
+  // Add the first waypoint
```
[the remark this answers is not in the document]
gilwoolee: I don't think this is meant to match the actual R1 value of the robot. It produces R1 trajectory such that two waypoints `[pi - alpha, -pi + alpha]` becomes `[pi - alpha, pi + alpha]` so that the smoothers(postprocessors) which assume R1 space can handle it gracefully. In the end, it is converted back to SO2 space before sending it via ROS, and low-level controllers are capable of handling it. @aditya-vk please correct me &/ confirm.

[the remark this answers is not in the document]
aditya-vk: That's right. This has nothing to do with the actual robot joint angle value. As @gilwoolee mentioned, this is only for timing to happen gracefully since the timers assume the values are R1.

src/trajectory/util.cpp
```diff
@@ -83,7 +92,51 @@ bool checkStateSpace(const statespace::StateSpace* _stateSpace)
   }
 }
 
-} // (anonymous) namespace
+void checkValidityOfSpaceAndTrajectory(
```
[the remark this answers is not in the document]
aditya-vk: Under `control/ros` we currently support only trajectories/robots with R1 and SO2 joints. We could potentially support other types of joints but that does not have to be part of this PR.
 https://github.com/personalrobotics/aikido/blob/67ab2db106391a65dcb959787e42e01811bb5257/include/aikido/control/ros/Conversions.hpp#L44

src/trajectory/util.cpp
```diff
@@ -315,5 +368,72 @@ UniqueSplinePtr createPartialTrajectory(
...
+  auto interpolator = std::dynamic_pointer_cast<const GeodesicInterpolator>(
+      trajectory.getInterpolator());
+  if (!interpolator)
+    throw std::invalid_argument(
+        "The interpolator of trajectory should be a GeodesicInterpolator");
+
+  // Create new trajectory space.
+  std::vector<aikido::statespace::ConstStateSpacePtr> subspaces;
+  for (std::size_t i = 0; i < space->getDimension(); ++i)
+    subspaces.emplace_back(std::make_shared<const R1>());
```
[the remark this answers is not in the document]
aditya-vk: Same as above.

src/control/ros/Conversions.cpp
```diff
@@ -374,8 +479,18 @@ std::unique_ptr<SplineTrajectory> toSplineJointTrajectory(
     numCoefficients = 2; // linear
 
   // Convert the ROS trajectory message to an Aikido spline.
-  std::unique_ptr<SplineTrajectory> trajectory{new SplineTrajectory{space}};
-  auto currState = space->createState();
+  // Add a segment to the trajectory.
+  std::vector<aikido::statespace::ConstStateSpacePtr> subspaces;
+  for (std::size_t citer = 0; citer < space->getDimension(); ++citer)
```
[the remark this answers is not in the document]
aditya-vk: Same as above.

src/control/ros/Conversions.cpp
```diff
@@ -173,8 +277,22 @@ void extractTrajectoryPoint(
 
   Eigen::VectorXd tangentVector;
   auto state = space->createState();
+  auto interpolator
```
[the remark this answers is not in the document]
aditya-vk: Interpolated trajectories hold the interpolator but splines do not. So I guess creating it once outside and passing reference to it should be better. Good catch!

aditya-vk:
@brianhou I do not think dealing with multiple DOF joint spaces needs to be a part of this PR given that the rest of the stack, especially trajectory execution does not support it. Other comments have been addressed.

include/aikido/trajectory/util.hpp
```diff
@@ -89,6 +89,26 @@ double findTimeOfClosestStateOnTrajectory(
 UniqueSplinePtr createPartialTrajectory(
     const Spline& traj, double partialStartTime);
 
+/// Converts an interpolated trajectory in the cartesian product space of SO(2)
+/// and R1 joints
+/// to a trajectory in cartesian product space of strictly only R1 joints.
+/// \param[in] space Statespace the input trajectory is in.
```
brianhou: Delete this line.

include/aikido/trajectory/util.hpp
```diff
@@ -89,6 +89,26 @@ double findTimeOfClosestStateOnTrajectory(
 UniqueSplinePtr createPartialTrajectory(
     const Spline& traj, double partialStartTime);
 
+/// Converts an interpolated trajectory in the cartesian product space of SO(2)
+/// and R1 joints
```
brianhou: This is wrapped weirdly, and things aren't capitalized correctly.

include/aikido/trajectory/util.hpp
```diff
@@ -89,6 +89,26 @@ double findTimeOfClosestStateOnTrajectory(
...
+/// to a trajectory in cartesian product space of strictly only R1 joints.
+/// \param[in] space Statespace the input trajectory is in.
+/// \param[in] trajectory Trajectory to be converted.
+/// \return Converted trajectory.
+aikido::trajectory::ConstInterpolatedPtr toR1JointTrajectory(
+    // aikido::statespace::ConstStateSpacePtr& space,
```
brianhou: Delete this line.

include/aikido/trajectory/util.hpp
```diff
@@ -89,6 +89,26 @@ double findTimeOfClosestStateOnTrajectory(
...
+    const aikido::trajectory::Interpolated& trajectory);
+
+/// Converts a spline trajectory in the cartesian product space of SO(2) and R1
+/// joints
+/// to a trajectory in cartesian product space of strictly only R1 joints.
+/// \param[in] space Statespace the input trajectory is in.
```
brianhou: Delete this line.

include/aikido/trajectory/util.hpp
```diff
@@ -89,6 +89,26 @@ double findTimeOfClosestStateOnTrajectory(
...
+/// \param[in] trajectory Trajectory to be converted.
+/// \return Converted trajectory.
+aikido::trajectory::ConstSplinePtr toR1JointTrajectory(
+    // aikido::statespace::ConstStateSpacePtr& space,
```
brianhou: Delete this line.

src/control/ros/Conversions.cpp
```diff
@@ -173,8 +281,26 @@ void extractTrajectoryPoint(
...
   auto state = space->createState();
+
+  auto prevState = space->createState();
+  space->convertPositionsToState(previousPoint, prevState);
+
   trajectory->evaluate(timeAbsolute, state);
-  space->logMap(state, tangentVector);
+  space->convertStateToPositions(state, tangentVector);
+
+  auto geodesicInterpolator = std::dynamic_pointer_cast<aikido::statespace::GeodesicInterpolator>(interpolator);
```
brianhou: Why don't we take in a `GeodesicInterpolator`, rather than always dynamic casting it here?

src/trajectory/util.cpp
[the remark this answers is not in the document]
brianhou: I'm okay with keeping higher DOF joints out of this, but this function still seems really similar to `checkStateSpace` aside from some better error messages. Is the main difference that `checkStateSpace` permits higher DOF joints?

I still don't think the `space` parameter is necessary, since you can take it from the trajectory. I think this function really only looks at the trajectory's state space, so it needs a better name if we decide that it needs to be kept.

src/trajectory/util.cpp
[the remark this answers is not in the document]
brianhou: Got it, thanks for the clarification.

src/control/ros/Conversions.cpp
```diff
@@ -19,11 +21,112 @@ namespace control {
...
+
+//==============================================================================
+void reorder(
+    const std::vector<std::pair<std::size_t, std::size_t>>& indexMap,
+    const Eigen::VectorXd& inVector,
+    Eigen::VectorXd& outVector)
+{
+  assert(indexMap.size() == static_cast<std::size_t>(inVector.size()));
+  outVector.resize(inVector.size());
+  for (auto index : indexMap)
```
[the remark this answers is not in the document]
brianhou: Bump? Or is this wrong?

src/control/ros/Conversions.cpp
aditya-vk: 1. Consistency: Seems to be the norm to pass in the base class and then cast it to the appropriate one.
2. Flexibility: In case we introduce new interpolator classes and want to use the one that's tied to the trajectory.

I don't have a strong preference, but I'd be inclined to go this route.

brianhou: In that case, can we replace this with methods that are defined in the `Interpolator` class?

src/trajectory/util.cpp
```diff
@@ -28,38 +36,16 @@ namespace trajectory {
...
-  else if (dynamic_cast<const R<6>*>(_stateSpace) != nullptr)
-  {
-    return true;
-  }
-  else if (dynamic_cast<const R<Eigen::Dynamic>*>(_stateSpace) != nullptr)
+  SinglularDOF,
+  NonSingularDOF
+};
+
+bool checkStateSpace(const statespace::StateSpace* _stateSpace, AllowJointType type)
```
aditya-vk: @brianhou @gilwoolee :

There are a couple of tests written (planner / trajectory namespaces for example) that instantiate an R2 statespace and use this function during the process. I agree that we should not have two functions doing the same here. This seems to be a reasonable solution for now, especially given we currently have support for multi-DOF joints some places and not others. I would like to have this solution in place with an issue raised to
- Change the tests to be cartesian product spaces with multiple R1 joints.
or
- Extend support for multi-DOF joint spaces uniformly and remove this enum usage.

Since this method is internal, I am okay with using an enum at this point but would liek to get rid of it soon. What do you guys think?

src/control/ros/Conversions.cpp
```diff
@@ -173,8 +282,29 @@ void extractTrajectoryPoint(
...
+  auto geodesicInterpolator
+      = std::dynamic_pointer_cast<aikido::statespace::GeodesicInterpolator>(
+          interpolator);
+  if (!geodesicInterpolator)
+  {
+    throw std::invalid_argument(
+        "The interpolator of trajectory should be a GeodesicInterpolator");
+  }
+
+  auto diff = geodesicInterpolator->getTangentVector(prevState, state);
```
brianhou: I'm not entirely sure I understand why this change is necessary. So far, my understanding of this PR is:
- We plan with SO(2) joints
- We convert SO(2) joints to R1, postprocess (because postprocessors only deal with Rn joints), then convert back to SO(2)

What are we trying to do in this function? Is it what I thought was happening in https://github.com/personalrobotics/aikido/pull/496#discussion_r253279846, i.e. basically converting this into an Rn-joint trajectory to send out via ROS?

My main qualm is that it feels wrong to be requiring a specific type of interpolator here. I'm wondering whether we can rewrite this without using `GeodesicInterpolator`-specific methods, perhaps by actually converting it to an Rn trajectory and then just using `logMap`?

src/trajectory/util.cpp
brianhou: I would prefer to remove the enum and just use a Cartesian product space of multiple R1 joints in these tests, as part of this PR. Then, we can revisit the multi-DOF joint space if/when it becomes useful in the future.

src/control/ros/Conversions.cpp
aditya-vk: To my understanding, we are committing to executing the shortest paths i.e. interpolating via the Geodesic anytime we execute a trajectory.

> What are we trying to do in this function?
We want to extract waypoints along the trajectory "ensuring that the interpolation happens along the geodesic so that SO2 joint space is handled appropriately". 

What is the PR doing?
- We plan with SO2 joints
- We convert SO2 joints to R1 and time it -> this generates a spline that has been timed appropriately.
- When converting to ROS trajectory, each waypoint from this spline is extracted using the trajectory's statespace which is SO2 and hence requiring us to use the previous point to be able to interpolate appropriately, and then use the converted extracted waypoint to construct the ROS trajectory.

Conversion into Rn-joint trajectory and then using logMap would also involve the geodesic interpolator so I am not sure how that would help.

1. The issue is that the base Interpolator class does not have a `getTangentVector()` and I don't see why it shouldn't since every derived class would need to support that. So I can alternatively add a virtual function there and if we take the interpolator by reference or obtain it from the trajectory, it should point to the right implementation? 
2. But the problem is we need to be able to support both Interpolated trajectories and Splines. Spline trajectories do not have an interpolator attached to them so we need to use _something_.

src/control/ros/Conversions.cpp
aditya-vk: See below.

src/trajectory/util.cpp
gilwoolee: I agree with Brian. Can we have an issue to track this in case it becomes necessary?

src/control/ros/Conversions.cpp
gilwoolee: @aditya-vk I didn't fully get the 3rd bullet po int. `trajectory` is already  in `MetaSkeletonStateSpace` so it's a Cartesian product of SO2. In SO2 spline trajectory, why do we need the previous point to interpolate appropriately?

src/control/ros/Conversions.cpp
```diff
@@ -19,11 +22,115 @@ namespace control {
...
+/// \param[in] velocitiesRequired True if velocities are required to be
+/// extracted.
+/// \param[in] accelerations Extracted accelerations corresponding to \c
+/// positions.
+/// \param[in] accelerationsRequired True if accelerations are required to be
+/// extracted.
+/// \param[in] indexMap Map denoting the correct ordering of trajectory data
+/// required.
+/// This is required in case the trajectory's joint indexing is different to
+/// metaskeleton
```
gilwoolee: Merge  84-85.

src/control/ros/Conversions.cpp
aditya-vk: Yes exactly. It is in the Cartesian Product of SO2, call it `space`. So when we evaluate the trajectory, `space` is used to retrieve the `state`. Although the spline polynomial is correct, when we retrieve it using `space`, the representation for SO2 joints is converted back to between `-pi` and `pi`.

aditya-vk: This leaves two options: Either have your trajectory be tied with `CartesianProduct` space of all `R1` joints which can again get messy since we check for `MetaSkeletonStateSpace` during execution or use the interpolation from the previous point.

aditya-vk: I was wondering if we could return `evaluatePolynomial()` instead of `evaluate()` since that returns the value of the polynomial which is what we desire. However, again, since the spline segments are solved from 0, we need to know the start state of the segment which is stores in `mSegments.startState` which is represented as R + SO2. So this route is pretty much useless as well.

brianhou:
After discussing with @aditya-vk, it seems that somewhere lower down in the stack would be a better place to convert to R1 trajectories (when necessary). I think this would make for a cleaner abstraction, since we'd be able to reason with the SO(2) joints all the way from planning until we send it off to the controller.

I've created https://github.com/personalrobotics/rewd_controllers/issues/30 to track this, let's clean up this PR to remove the conversion here.

PR #497, opened 2019-01-27
Enhancement/pose estimator

yskim041:
- Cleaned up hacks for nips demo.
- Added `DetectedObject` class
- `PoseEstimatorModule` gets an object's unique id and additional information in JSON format from the `text` field of a `Marker`. `DetectedObject` is a wrapper for the information and delegates a detected object.

***


src/perception/PoseEstimatorModule.cpp
```diff
@@ -195,5 +207,23 @@ bool PoseEstimatorModule::detectObjects(
   return true;
 }
 
+YAML::Node PoseEstimatorModule::getObjInfo(const std::string& obj_id)
```
brianhou: Can we add the `//===` line here?

src/perception/PoseEstimatorModule.cpp
```diff
@@ -195,5 +207,23 @@ bool PoseEstimatorModule::detectObjects(
...
+  if (!targetNode)
+  {
+    throw std::runtime_error(
+        "[PoseEstimatorModule::getObjInfo] Error: invalid obj_id");
+  }
+
+  return targetNode;
+}
+
+void PoseEstimatorModule::setObjectProjectionHeight(double projectionHeight)
```
brianhou: Can we add the `//===` line here?

src/perception/PoseEstimatorModule.cpp
```diff
@@ -6,6 +6,7 @@
 #include <ros/topic.h>
 #include <visualization_msgs/Marker.h>
 #include <visualization_msgs/MarkerArray.h>
+#include <yaml-cpp/exceptions.h>
```
brianhou: Is this include being used somewhere?

include/aikido/perception/PoseEstimatorModule.hpp
```diff
@@ -75,6 +87,12 @@ class PoseEstimatorModule : public PerceptionModule
 
   /// Listens to the transform attached to the node
   tf::TransformListener mTfListener;
+
+  /// For additional information for each object
+  std::unordered_map<std::string, YAML::Node> mObjInfo;
+
+  bool mProjectObjectToFixedHeight;
```
brianhou: Nit: missing documentation

include/aikido/perception/PoseEstimatorModule.hpp
```diff
@@ -54,6 +55,17 @@ class PoseEstimatorModule : public PerceptionModule
       ros::Duration timeout = ros::Duration(0.0),
       ros::Time timestamp = ros::Time(0.0)) override;
 
+  /// Getter for \c mObjInfo. \c mObjInfo is saving additional information
+  /// from the text field in the MarkerArray message in json format.
+  /// This getter returns a YAML::Node of an object by id.
+  ///
+  /// \param[in] obj_id The key of an object to retrieve its additional
+  /// information
+  YAML::Node getObjInfo(const std::string& obj_id);
```
brianhou: Nit: `objId`

include/aikido/perception/PoseEstimatorModule.hpp
```diff
@@ -54,6 +55,17 @@ class PoseEstimatorModule : public PerceptionModule
       ros::Duration timeout = ros::Duration(0.0),
       ros::Time timestamp = ros::Time(0.0)) override;
 
+  /// Getter for \c mObjInfo. \c mObjInfo is saving additional information
```
brianhou: How will this be used? It'd be really helpful to provide an example of what kind of JSON information would be sent, aside from the object ID.

In addition, does it make sense to return a YAML node here? What if we just return the string and delegate the JSON parsing to the client script? (I think this might also just be unclear to me since I'm not sure what information is being sent.)

include/aikido/perception/PoseEstimatorModule.hpp
```diff
@@ -75,6 +87,12 @@ class PoseEstimatorModule : public PerceptionModule
 
   /// Listens to the transform attached to the node
   tf::TransformListener mTfListener;
+
+  /// For additional information for each object
```
brianhou: Examples here would be helpful!

yskim041: I agree. I will add some examples of new methods.

include/aikido/perception/PoseEstimatorModule.hpp
yskim041: I removed this line.

src/perception/PoseEstimatorModule.cpp
yskim041: I removed this line.

include/aikido/perception/PoseEstimatorModule.hpp
yskim041: I also thought about this point. I just added DetectedObject class that wraps the YAML node and delegates a detected object so that we can pass more general information comparing to the skeleton in the aikido world. By using the JSON format, a third party perception algorithm can send any types of additional information to the application.

include/aikido/perception/PerceptionModule.hpp
```diff
@@ -21,6 +22,7 @@ class PerceptionModule
   /// that represents the world.
   ///
   /// \param[in] env World to add perceived objects to.
+  /// \param[out] detectedObjects An output vector for detected objects.
```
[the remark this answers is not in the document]
brianhou: Quick question -- why is this necessary? The `World` should already have the objects added right?

src/perception/DetectedObject.cpp
```diff
@@ -0,0 +1,64 @@
...
+namespace aikido {
+namespace perception {
+
+//==============================================================================
+DetectedObject::DetectedObject(
+    const std::string& objDBKey,
+    const std::string& detectionFrameID,
+    const std::string& yamlStr)
+  : mObjDBKey(std::move(objDBKey))
+  , mDetectionFrameID(std::move(detectionFrameID))
```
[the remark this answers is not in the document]
egordon: `std::move` doesn't actually mutate the object by itself. It just allows the move constructor to be called. In the `const` case, it should resort to the copy constructor automatically. See https://stackoverflow.com/questions/28595117/why-can-we-use-stdmove-on-a-const-object

include/aikido/perception/PerceptionModule.hpp
```diff
@@ -31,6 +33,7 @@ class PerceptionModule
   /// otherwise.
   virtual bool detectObjects(
       const aikido::planner::WorldPtr& env,
+      std::vector<DetectedObject>& detectedObjects,
       ros::Duration timeout,
```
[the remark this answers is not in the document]
egordon: Non-const reference cannot have a default value (https://stackoverflow.com/questions/1059630/default-value-to-a-parameter-while-passing-by-reference-in-c)

However, I extended the API to allow for passing either by reference or by pointer. And passing by pointer has a default value of `nullptr`.

include/aikido/perception/PerceptionModule.hpp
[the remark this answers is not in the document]
egordon: It's not necessary, only a convenience. Default value is nullptr if you don't want it.

include/aikido/perception/PoseEstimatorModule.hpp
```diff
@@ -40,17 +42,18 @@ class PoseEstimatorModule : public PerceptionModule
...
+      const std::string& referenceFrameId,
       dart::dynamics::Frame* referenceLink);
 
   virtual ~PoseEstimatorModule() = default;
 
   // Documentation inherited
   bool detectObjects(
       const aikido::planner::WorldPtr& env,
+      std::vector<DetectedObject>& detectedObjects,
       ros::Duration timeout = ros::Duration(0.0),
```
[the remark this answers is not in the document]
egordon: Changed to empty Duration

include/aikido/perception/DetectedObject.hpp
```diff
@@ -0,0 +1,80 @@
...
+  std::string getDetectionFrameID();
+
+  /// Get the map of keys to additional informations
+  YAML::Node getYamlNode();
+
+  /// Get a specific value from the information map by a key and the typename
+  /// of the field
+  /// \param[in] key The key (string) of a field in the information map
+  template <typename T>
+  T getInfoByKey(const std::string& key);
```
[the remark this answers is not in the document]
egordon: Added sequence/map types to docstring.

egordon:
@brianhou @gilwoolee Ready for re-review!

include/aikido/perception/PerceptionModule.hpp
```diff
@@ -31,9 +33,18 @@ class PerceptionModule
   /// otherwise.
   virtual bool detectObjects(
       const aikido::planner::WorldPtr& env,
+      std::vector<DetectedObject>* detectedObjects,
       ros::Duration timeout,
       ros::Time timestamp)
       = 0;
+
+  bool detectObjects(
```
gilwoolee: Why are these two near-duplicates necessary? If this is an optional out-param, can we just have `const std::vector<DetectedObject>& detectedObject = std::vector<DetectedObject>()`?

src/perception/PoseEstimatorModule.cpp
```diff
@@ -128,7 +157,7 @@ bool PoseEstimatorModule::detectObjects(
     // Check if skel in World
     // If there is, update its pose
     // If not, add skeleton to env
-    // A pose estimator module should provide the unique obj_id per object
+    // A pose estimator module should provide the unique obj_uid per object
```
gilwoolee: I think this statement is no longer valid.

src/perception/PoseEstimatorModule.cpp
```diff
@@ -63,25 +62,31 @@ bool PoseEstimatorModule::detectObjects(
...
-    // remove a skeleton with obj_id from env
-    // and move to next marker
-    dart::dynamics::SkeletonPtr env_skeleton = env->getSkeleton(obj_id);
-    if (!obj_key.compare("None"))
-    {
-      if (env_skeleton != nullptr)
-      {
-        env->removeSkeleton(env_skeleton);
-      }
+    const std::string obj_uid = obj_ns + "_" + std::to_string(obj_id);
```
gilwoolee: We should mention in the header of this file that this is how we assign uids to objects.

src/perception/PoseEstimatorModule.cpp
```diff
@@ -91,7 +96,31 @@ bool PoseEstimatorModule::detectObjects(
...
+
+    // Add object to output vector, if available
+    if(detectedObjects != nullptr) {
+      detectedObjects->push_back(this_object);
+    }
+    
+
+    // If marker_transform.action is "DELETE",
+    // remove a skeleton with obj_uid from env
+    // and move to next marker
```
gilwoolee: I think we can remove this line

src/perception/PoseEstimatorModule.cpp
```diff
@@ -63,25 +62,31 @@ bool PoseEstimatorModule::detectObjects(
...
-    {
-      if (env_skeleton != nullptr)
-      {
-        env->removeSkeleton(env_skeleton);
-      }
+    const std::string obj_uid = obj_ns + "_" + std::to_string(obj_id);
+
+    // Initialize a DetectedObject class for this object
+    DetectedObject this_object
+        = DetectedObject(obj_uid, obj_ns, detection_frame, marker_transform.text);
```
gilwoolee: I think it's possible that this object already exists in the World, in which case we should just update the existing one. Otherwise wouldn't this generate infinitely many objects of the same uid?

egordon: This check is already done on line 157.

include/aikido/perception/PerceptionModule.hpp
egordon: Doesn't work, then we can't add any new elements.

egordon:
@gilwoolee Ready for re-review!

CHANGELOG.md
```diff
@@ -43,6 +43,7 @@
 
   * Added integrated PoseEstimatorModule: [#336](https://github.com/personalrobotics/aikido/pull/336)
   * Added voxel grid perception module: [#448](https://github.com/personalrobotics/aikido/pull/448)
+  * Added YAML communication between PoseEstimatorModule and third party perception algorithms: [#497](https://github.com/personalrobotics/aikido/pull/497)
```
```suggestion
  * Added YAML communication between PoseEstimatorModule and third-party perception algorithms: [#497](https://github.com/personalrobotics/aikido/pull/497)
```

include/aikido/perception/DetectedObject.hpp
```diff
@@ -0,0 +1,89 @@
+#ifndef AIKIDO_PERCEPTION_DETECTEDOBJECT_HPP_
+#define AIKIDO_PERCEPTION_DETECTEDOBJECT_HPP_
+
+#include <string>
+#include "aikido/io/yaml.hpp"
+
+namespace aikido {
+namespace perception {
+
+/// DetectedObject delegates a detected object from a third party
```
brianhou: Is "delegates" the right word here?
```suggestion
/// DetectedObject delegates a detected object from a third-party
```

include/aikido/perception/DetectedObject.hpp
```diff
@@ -0,0 +1,89 @@
...
+/// perception algorithm.
+
+/// A perception algorithm should send information for an object via ROS
+/// visualization_msgs/Marker message like the following:
+/// Marker.header.frame_id -> detectionFrameID
```
brianhou: The order of these doesn't match the order of the parameters.

include/aikido/perception/DetectedObject.hpp
```diff
@@ -0,0 +1,89 @@
...
+/// Marker.ns + "_" + Marker.id -> objUid (identity in DART world)
+/// Marker.ns -> objAssetDBKey (determines visual asset, see ObjectDatabase)
```
brianhou: We should use Doxygen commands where relevant, e.g. `\see`.

include/aikido/perception/DetectedObject.hpp
brianhou: Here's a stab at new documentation:

DetectedObject stores information about objects detected by a third-party perception system. The perception system should communicate object information via a ROS visualization_msgs/Marker message.

The constructor parameters correspond to the following fields in the Marker message: ...

include/aikido/perception/DetectedObject.hpp
```diff
@@ -0,0 +1,89 @@
...
+/// Marker.text -> yamlStr
+
+/// yamlStr contains any types of additional
```
brianhou: Wrapping is a little funky here?

include/aikido/perception/DetectedObject.hpp
brianhou: Actually, maybe we should be mentioning the fields rather than constructor parameters here, since this is the class documentation.

include/aikido/perception/DetectedObject.hpp
```diff
@@ -0,0 +1,89 @@
...
+/// }
+/// \endcode
+
+/// Clarification: objUid = DART world ID; objDbKey = visual asset key in ObjectDatabase.
+
+class DetectedObject
+{
+public:
+  /// Construct a \c DetectedObject
+  /// \param[in] objUid Unique ID for object in DART world. Same UID -> Same Object
```
brianhou: Would `dartUid` be a reasonable name for this parameter?

include/aikido/perception/DetectedObject.hpp
```diff
@@ -0,0 +1,89 @@
...
+  /// \param[in] objAssetDBKey Key for ObjectDatabase passed into constructor of PoseEstimatorModule. Defines visuals / assets.
```
brianhou: Maybe `assetKey`?

include/aikido/perception/ObjectDatabase.hpp
```diff
@@ -10,8 +10,8 @@
 namespace aikido {
 namespace perception {
 
-/// Instantiation of ObjectDatabase that reads of JSON file containing the
-/// information that maps object keys to the object names and resources.
+/// Instantiation of ObjectDatabase (a.k.a. Asset Database) that reads of JSON file containing the
```
brianhou: Does it make sense to just call this class `AssetDatabase` rather than `ObjectDatabase`?

include/aikido/perception/DetectedObject.hpp
```diff
@@ -0,0 +1,89 @@
...
+  /// \param[in] detectionFrameID Frame ID from ROS Marker
+  /// \param[in] yamlStr String of additional parameters for object. Can override objAssetDBKey with "db_key".
```
brianhou:
```suggestion
  /// \param[in] yamlStr String of additional parameters for object. Can override objAssetDBKey by specifying "db_key".
```

include/aikido/perception/DetectedObject.hpp
```diff
@@ -0,0 +1,89 @@
...
+  DetectedObject(
+      const std::string& objUID,
+      const std::string& objAssetDBKey,
+      const std::string& detectionFrameID,
+      const std::string& yamlStr);
+
+  virtual ~DetectedObject() = default;
+
+  /// Get the unique id of the object
```
brianhou:
```suggestion
  /// Get the unique DART id of the object
```

include/aikido/perception/DetectedObject.hpp
```diff
@@ -0,0 +1,89 @@
...
+  std::string getObjUID();
```
brianhou:
```suggestion
  std::string getDartUid();
```

src/perception/PoseEstimatorModule.cpp
```diff
@@ -63,25 +62,33 @@ bool PoseEstimatorModule::detectObjects(
...
+
+    const std::string obj_db_key = this_object.getObjDBKey();
+
+    // check if valid object
+    if (obj_db_key == "") {
```
brianhou:
```suggestion
    if (obj_db_key.empty()) {
```

src/perception/PoseEstimatorModule.cpp
```diff
@@ -63,25 +62,33 @@ bool PoseEstimatorModule::detectObjects(
...
-      }
+    const std::string obj_uid = obj_ns + "_" + std::to_string(obj_id);
+
+    // Initialize a DetectedObject class for this object
+    DetectedObject this_object
+        = DetectedObject(obj_uid, obj_ns, detection_frame, marker_transform.text);
+
+    const std::string obj_db_key = this_object.getObjDBKey();
+
+    // check if valid object
```
brianhou:
```suggestion
```

src/perception/PoseEstimatorModule.cpp
```diff
@@ -91,7 +98,30 @@ bool PoseEstimatorModule::detectObjects(
...
-    mConfigData->getObjectByKey(obj_key, obj_name, obj_resource, obj_offset);
+    try {
+      mConfigData->getObjectByKey(obj_db_key, obj_name, obj_resource, obj_offset);
+    } catch(std::runtime_error &e) {
+      dtwarn << e.what() << std::endl;
+      continue;
+    }
+
+    // Add object to output vector, if available
+    if(detectedObjects != nullptr) {
```
brianhou:
```suggestion
    if (detectedObjects) {
```

src/perception/PoseEstimatorModule.cpp
```diff
@@ -91,7 +98,30 @@ bool PoseEstimatorModule::detectObjects(
...
+      detectedObjects->push_back(this_object);
```
brianhou: Is there a reason to construct `DetectedObject` and then `push_back`, rather than `emplace_back`? It seems like we have access to all the strings already anyway.

egordon: The constructed object is used later in the function, so I don't believe there is a penalty to using `push_back`.

include/aikido/perception/DetectedObject.hpp
egordon: Yes, it would.

include/aikido/perception/DetectedObject.hpp
egordon: Yep, was trying to avoid confusion with the existing class "ObjectDatabase", but honestly, I think renaming it to "AssetDatabase" makes the most sense.

include/aikido/perception/AssetDatabase.hpp
```diff
@@ -34,34 +34,34 @@ namespace perception {
...
   /// Get the object name, resource, and offset from database by objectKey
-  /// \param[in]  objectKey The key (string) of an object in ObjectDatabase
-  /// \param[out] objectName The retrieved object name from ObjectDatabase
-  /// \param[out] objectResource The retrieved uri of the object
-  /// \param[out] objectOffset The retrieved offset matrix of the object
+  /// \param[in]  assetKey The key (string) of an object in AssetDatabase
+  /// \param[out] assetName The retrieved object name from AssetDatabase
+  /// \param[out] assetResource The retrieved uri of the object
+  /// \param[out] assetOffset The retrieved offset matrix of the object
   ///     i.e. the offset between a tag and the actual origin of an object
```
gilwoolee: Nit: `i.e.` should probably be `e.g.` here since the "tag" refers to AprilTag in this context.

include/aikido/perception/DetectedObject.hpp
```diff
@@ -0,0 +1,91 @@
...
+
+namespace aikido {
+namespace perception {
+
+/// DetectedObject delegates a detected object from a third-party
+/// perception algorithm.
+
+/// A perception algorithm should send information for an object via ROS
+/// visualization_msgs/Marker message like the following:
+/// Marker.ns + "_" + Marker.id -> dartUid (identity in DART world)
```
gilwoolee: I'm a bit confused about calling it `dartUid` because DART does not assign uid to an object afaik. But I think it makes keeps the skeleton names to be unique. Does `dartUid` refer to the name of the DART object, i.e. `dartSkeleton->getName() == dartUid`? Also is it DART world or Aikido world?

include/aikido/perception/PerceptionModule.hpp
```diff
@@ -26,13 +27,15 @@ class PerceptionModule
   /// \param[in] timestamp Only detections more recent than this timestamp will
   /// be accepted. A timestamp of 0 greedily takes the first available message,
   /// and is the default behaviour.
+  /// \param[out] detectedObjects An output vector for detected objects.
   /// \return Returns \c false if no detection observed, or if none of the
```
gilwoolee: Hmm. I think this return parameter is misleading and is redundant. If there is no detection available since `timestamp`, the `detectedObjects` should be empty.

include/aikido/perception/PoseEstimatorModule.hpp
```diff
@@ -40,19 +43,25 @@ class PoseEstimatorModule : public PerceptionModule
...
+      std::shared_ptr<AssetDatabase> assetData,
       std::shared_ptr<aikido::io::CatkinResourceRetriever> resourceRetriever,
-      std::string referenceFrameId,
+      const std::string& referenceFrameId,
       dart::dynamics::Frame* referenceLink);
 
   virtual ~PoseEstimatorModule() = default;
 
   // Documentation inherited
+  /// Looks for the following information sent via ROS (see \c DetectedObject):
```
gilwoolee: If inherited, no need to put this again here. If you're putting additional information specific to this class, you can do 
`/// \copydoc PerceptionModule::detectObjects()` and add more stuff.

include/aikido/perception/PoseEstimatorModule.hpp
```diff
@@ -40,19 +43,25 @@ class PoseEstimatorModule : public PerceptionModule
...
+  /// Marker.header.frame_id -> detectionFrameID
+  /// Marker.ns + "_" + Marker.id -> objUid (identity in DART world)
+  /// Marker.ns -> objAssetDBKey (determines visual asset, see AssetDatabase)
   bool detectObjects(
       const aikido::planner::WorldPtr& env,
-      ros::Duration timeout = ros::Duration(0.0),
-      ros::Time timestamp = ros::Time(0.0)) override;
+      ros::Duration timeout = ros::Duration(),
+      ros::Time timestamp = ros::Time(0.0),
+      std::vector<DetectedObject>* detectedObjects = nullptr) override;
```
gilwoolee: It seems like the default params are specified only here and not in `PerceptionModule`? And `PerceptionModule`'s document seems to suggest otherwise. Could you move these default values to `PerceptionModule`?

src/perception/AssetDatabase.cpp
```diff
@@ -26,54 +26,55 @@ ObjectDatabase::ObjectDatabase(
   }
 
   // Load from string
-  mObjData = YAML::Load(content);
+  mAssetData = YAML::Load(content);
```
gilwoolee: Can we do try/catch in case the content is not YAML-correct?

src/perception/DetectedObject.cpp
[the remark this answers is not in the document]
gilwoolee: Sure, but I still don't see why you'd explicitly state 'move' when you're actually copying.

src/perception/PoseEstimatorModule.cpp
```diff
@@ -63,25 +62,39 @@ bool PoseEstimatorModule::detectObjects(
 
   for (const auto& marker_transform : marker_message->markers)
   {
+    // TODO: Add DELETE_ALL Functionality
+    // TODO: Update when we move over to ROS Kinetic. Indigo is dumb and doesn't
```
gilwoolee: Let's remove the "is dumb"

src/perception/PoseEstimatorModule.cpp
```diff
@@ -90,8 +103,34 @@ bool PoseEstimatorModule::detectObjects(
     Eigen::Isometry3d obj_offset;
     ros::Time t0 = ros::Time(0);
 
-    // get the object name, resource, and offset from database by objectKey
-    mConfigData->getObjectByKey(obj_key, obj_name, obj_resource, obj_offset);
+    // get the object name, resource, and offset from database by assetKey
```
gilwoolee:
```suggestion
    // Get the object name, resource, and offset from database by assetKey
```

src/perception/PoseEstimatorModule.cpp
```diff
@@ -63,25 +62,39 @@ bool PoseEstimatorModule::detectObjects(
...
 
-    // If obj_key is "None",
-    // remove a skeleton with obj_id from env
-    // and move to next marker
-    dart::dynamics::SkeletonPtr env_skeleton = env->getSkeleton(obj_id);
-    if (!obj_key.compare("None"))
+    const std::string obj_uid = obj_ns + "_" + std::to_string(obj_id);
+
+    // Initialize a DetectedObject class for this object
+    DetectedObject this_object = DetectedObject(
```
gilwoolee:
```suggestion
    DetectedObject this_object(
```

src/perception/PoseEstimatorModule.cpp
```diff
@@ -90,8 +103,34 @@ bool PoseEstimatorModule::detectObjects(
...
+
+    // Add object to output vector, if available
+    if (detectedObjects)
+    {
+      detectedObjects->push_back(this_object);
+    }
+
+    // If marker_transform.action is "DELETE",
+    // remove a skeleton with obj_uid from env
+    dart::dynamics::SkeletonPtr env_skeleton = env->getSkeleton(obj_uid);
```
gilwoolee: Seems like the uid is used in `aikido::World`, not DART world

include/aikido/perception/PerceptionModule.hpp
egordon: It is useful for two reasons:
(1) If the user doesn't use detectedObjects (passes in `nullptr` / keeps to the default)
(2) If a user wants to do quick error detection without accessing detectedObjects. e.g.
```
while(!detectedObjects()) {
printf("Waiting for objects!);
}
```

include/aikido/perception/PoseEstimatorModule.hpp
egordon: Defaults need to be present in the subclass that is actually being called.

include/aikido/perception/DetectedObject.hpp
gilwoolee:
```suggestion
/// Marker.ns + "_" + Marker.id -> objectUid (identity in planner::World)
```

include/aikido/perception/PoseEstimatorModule.hpp
```diff
@@ -40,19 +43,25 @@ class PoseEstimatorModule : public PerceptionModule
...
       dart::dynamics::Frame* referenceLink);
 
   virtual ~PoseEstimatorModule() = default;
 
-  // Documentation inherited
+  /// \copydoc PerceptionModule::detectObjects()
+  /// Looks for the following information sent via ROS (see \c DetectedObject):
+  /// visualization_msgs/Marker message like the following:
+  /// Marker.header.frame_id -> detectionFrameID
+  /// Marker.ns + "_" + Marker.id -> objUid (identity in DART world)
```
gilwoolee:
```suggestion
  /// Marker.ns + "_" + Marker.id -> objUid (identity in planner::World)
```

src/perception/PoseEstimatorModule.cpp
```diff
@@ -128,7 +167,6 @@ bool PoseEstimatorModule::detectObjects(
     // Check if skel in World
     // If there is, update its pose
     // If not, add skeleton to env
-    // A pose estimator module should provide the unique obj_id per object
     if (env_skeleton == nullptr)
```
gilwoolee: Let's change this to `if (!env_skeleton)`

src/perception/PoseEstimatorModule.cpp
```diff
@@ -90,8 +103,34 @@ bool PoseEstimatorModule::detectObjects(
...
+      mAssetData->getAssetByKey(asset_key, obj_name, obj_resource, obj_offset);
+    }
+    catch (std::runtime_error& e)
+    {
+      dtwarn << e.what() << std::endl;
+      continue;
+    }
+
+    // Add object to output vector, if available
+    if (detectedObjects)
```
yskim041: I think we need to handle the case when `marker_transform.action` is `DELETE`.

egordon: Good point. I just push it to the back at the end of the for loop.

PR #499, opened 2019-02-03
Pybind11 manual bindings

gilwoolee:
This PR contains minimal class definitions for exposing aikido classes needed for other repositories such as libherb and pr_tsr.

The python binding is built not part of aikido but as a separate project _after_ aikido is built via catkin.

***


python/README.md
```diff
@@ -0,0 +1,56 @@
+# This folder contains manual bindings for aikidopy using pybind11.
```
brianhou: Nit: this header shouldn't be a sentence. Maybe something like
```suggestion
# aikidopy
```
with that sentence as just regular text would make more sense.

python/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+0. The installation guide have been tested on
+
+- sudo apt installed python 2.7 and python 3.4.
```
brianhou: Nit: capitalization (e.g. `python` -> `Python`)

python/README.md
```diff
@@ -0,0 +1,56 @@
...
+- anaconda managed python 3.6
+
+If you use system installed python, ensure you have `python2.7-dev` or `python3-dev` correspondingly, or run `sudo apt-get install python3-dev` to obtain it.
```
brianhou: I think this part about installing the dev headers is the actual step 0. The stuff before this about versions that have been tested is useful information, but aren't actually part of the checklist.

python/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+
+1. Install [pybind11](https://github.com/pybind/pybind11.git) **from source** following this [instruction](https://pybind11.readthedocs.io/en/master/basics.html#compiling-the-test-cases). (Need version >= 2.2.0).
```
brianhou:
```suggestion
1. Install [pybind11](https://github.com/pybind/pybind11.git) **from source** following these [instructions](https://pybind11.readthedocs.io/en/master/basics.html#compiling-the-test-cases). (Need version >= 2.2.0).
```

python/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+```
+git clone https://github.com/pybind/pybind11.git
```
brianhou: Bash commands should be prepended with a `$`, Python commands should be prepended with `>>>`.

python/README.md
```diff
@@ -0,0 +1,56 @@
...
+cd pybind11
+mkdir build
+cd build
+cmake ..
+make -j 4
+```
+
+- For system managed python: `sudo make install`;
```
brianhou: Nit: capitalization.

python/README.md
```diff
@@ -0,0 +1,56 @@
...
+- For anaconda: `pip install -e .`.
+
+You should be able to load `pybind11` in your python.
+
+```
+$ python
+>> import pybind11
```
brianhou:
```suggestion
>>> import pybind11
```

python/README.md
```diff
@@ -0,0 +1,56 @@
...
+```
+
+2. Build aikido and source the setup file.
+```
+catkin build aikido
```
brianhou: It seems that this is assuming that the commands are run from the `src/aikido` directory rather than the workspace directory?

python/README.md
```diff
@@ -0,0 +1,56 @@
...
+source ../../devel.setup.bash
```
brianhou:
```suggestion
source ../../devel/setup.bash
```

python/README.md
```diff
@@ -0,0 +1,56 @@
...
+3. Create a folder `workspace/src/aikido/python/build` (instead of `workspace/src/aikido/build`). Build and install the python binding.
+
+```
+cd build
+cmake .. -DAIKIDOPY_PYTHON_VERSION=3.4
+make -j
+sudo make install
+```
+
+Read the output of `sudo make install` and ensure that `aikidopy.so` get installed in desired location.
```
brianhou: Where is the desired location?

python/README.md
```diff
@@ -0,0 +1,56 @@
...
+```
+catkin build aikido
+source ../../devel.setup.bash
+```
+
+3. Create a folder `workspace/src/aikido/python/build` (instead of `workspace/src/aikido/build`). Build and install the python binding.
+
+```
+cd build
+cmake .. -DAIKIDOPY_PYTHON_VERSION=3.4
```
brianhou: Does this work even for conda 3.6?

python/README.md
```diff
@@ -0,0 +1,56 @@
...
+make -j
```
brianhou: Missing number of threads.

python/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+Read the output of `sudo make install` and ensure that `aikidopy.so` get installed in desired location.
+
+4. Try loading.
+```
+python
+>>> import aikidopy
+```
+
+5. Tips for multiple python versions
```
brianhou: Maybe this should be a `##` header rather than part of the checklist?

python/README.md
Kelym: According to the [man page](https://linux.die.net/man/1/make), I believe that `make -j` will simply "not limit the number of jobs that will run simultaneously". Good enough?

gilwoolee: Yeah I typically just do `make -j`.

gilwoolee:
> Mostly left comments on the README. Can you also explain what the difference is between `aikidopy.cpp` and `aikido_py.cpp`?

Hmm I was following @jslee02 's example code I got for herbpy so @jslee02 might have a better answer (or suggestions). My understanding is that `aikidopy.cpp` defines things (more like hpp files) and `aikido_py.cpp` actually implements them. @jslee02 could you clarify?

python/README.md
brianhou: TIL, cool beans!

jslee02: :skull_and_crossbones: : Use this option at your own risk especially building python bindings. pybind11 (and Boost.Python) utilizes tons of templates, and this consumes your memory a lot, which is proportional to the number of threads. If you don't have enough memory, your system could end up being frozen due to the lack of memory. This happened a lot to me even with 8 threads + 32GB memory (with other programs on memory).

jslee02:
I use `<python_module_name>.cpp` for the main module (`aikidopy.cpp` in this case), which contains `PYBIND11_MODULE`. For the actual bindings, I define them in separate files per namespace and classes for example (see: https://github.com/dartsim/dart/tree/pybind11/python/dartpy).

It seems `aikido_py.cpp` is (currently) a single file for all the bindings of AIKIDO. I think, if necessary, it could be split and renamed when more AIKIDO APIs are bound.

PR #503, opened 2019-02-08
ConcreteRobot uses ranker for PlantoTSR

gilwoolee:
This PR updates `ConcreteRobot` class to use `ConfigurationRanker` when calling `planToTSR`. It changes the `planToTSR` logic a little bit accordingly.

***


gilwoolee:
> Thanks for the PR! We should also probably update the new planner API: `ConfigurationToTSR` and `ConfigurationToConfiguration_to_ConfigurationToTSR` to have a ranker?

They already are. Ranker is one of the constructor arguments in `ConfigurationToConfigurationPlanner` and `ConfigurationToConfiguration_to_ConfigurationToTSR` as this is a property of the planner not the problem. I changed them to `ConstConfigurationRanker` as well.

src/robot/util.cpp
```diff
@@ -217,49 +223,70 @@ trajectory::TrajectoryPtr planToTSR(
...
+    configurations.emplace_back(goalState.clone());
+  }
+
+  {
+    // Set to start state
+    std::lock_guard<std::mutex> lock(robot->getMutex());
+    space->setState(metaSkeleton.get(), startState);
+  }
+
+  if (configurations.size() == 0)
```
brianhou:
```suggestion
  if (configurations.empty())
```

src/robot/util.cpp
```diff
@@ -217,49 +223,70 @@ trajectory::TrajectoryPtr planToTSR(
...
     auto problem = ConfigurationToConfiguration(
-        space, startState, goalState, collisionTestable);
-    auto traj = planner->plan(problem, &pResult);
+        space, startState, configurations[i], collisionTestable);
+
+    auto traj = snapPlanner->plan(problem, &pResult);
 
     if (traj)
+    {
+      space->setState(metaSkeleton.get(), startState);
```
brianhou: Is this necessary? If we do `plan` properly, shouldn't it restore the metaskeleton state?

aditya-vk: Looking at the code, looks like it won't happen. 

This is probably because we are not using `planner::dart::ConfigurationToConfiguration` problem and rightfully so since we are not using a Dart planner, so the `problem` does not hold the metaskeleton for the planner to set the skeleton back. Alternatively, we can 
1. use the adapter `ConfigurationToConfiguration_to_ConfigurationToConfiguration` (let's rename this soon, it's painful to even look at :P) to make the snap planner hold the skeleton
2. In the adapter instead of `return mDelegate->plan()`, we set the skeleton back to original state and return the planned trajectory.

brianhou: Whoops, I think we're both wrong. We've already created a `MetaSkeletonStateSaver` so this should already be taken care of?

https://github.com/personalrobotics/aikido/blob/fac84625aa8fc1a6893a1d581d29dc0a3c52c81d/src/robot/util.cpp#L217

gilwoolee: Ah, yeah I think @brianhou 's last comment is correct. Let me remove that line.

gilwoolee:
> Actually, I think we should update the `dart::ConfigurationToConfiguration_to_ConfigurationToTSR` adapter to use the ranker? Also, we need to update the changelog.

@brianhou it was already using a ranker. I just changed it to use the `Const` version.

brianhou:
I don't think the `plan` method actually uses the ranker anywhere, though?

https://github.com/personalrobotics/aikido/blob/41032c73c79495cd2323fd07f660d6e795cf42fe/src/planner/dart/ConfigurationToConfiguration_to_ConfigurationToTSR.cpp#L88

CHANGELOG.md
```diff
@@ -62,6 +61,7 @@
   * Added a kinodynamic timer that generates a time-optimal smooth trajectory without completely stopping at each waypoint: [#443](https://github.com/personalrobotics/aikido/pull/443)
   * Fixed segmentation fault on 32-bit machines in vector-field planner: [#459](https://github.com/personalrobotics/aikido/pull/459)
   * Updated interface to OMPL planners to follow the style of the new refactored planning API: [#466](https://github.com/personalrobotics/aikido/pull/466)
+  * TSR planners use ConfigurationRanker: [#503](https://github.com/personalrobotics/aikido/pull/503)
```
```suggestion
  * Used ConfigurationRanker in TSR planners: [#503](https://github.com/personalrobotics/aikido/pull/503)
```

gilwoolee:
> I don't think the `plan` method actually uses the ranker anywhere, though?
> 
> [aikido/src/planner/dart/ConfigurationToConfiguration_to_ConfigurationToTSR.cpp](https://github.com/personalrobotics/aikido/blob/41032c73c79495cd2323fd07f660d6e795cf42fe/src/planner/dart/ConfigurationToConfiguration_to_ConfigurationToTSR.cpp#L88)
> 
> Line 88 in [41032c7](/personalrobotics/aikido/commit/41032c73c79495cd2323fd07f660d6e795cf42fe)
> 
>  while (generator->canSample())

lol!!! haha since I didn't make that change in this PR I didn't check... Let me fix that :P

gilwoolee:
@brianhou @aditya-vk please take another look!

src/planner/dart/ConfigurationToConfiguration_to_ConfigurationToTSR.cpp
```diff
@@ -84,24 +87,57 @@ ConfigurationToConfiguration_to_ConfigurationToTSR::plan(
   auto saver = MetaSkeletonStateSaver(mMetaSkeleton);
   DART_UNUSED(saver);
 
+  auto robot = mMetaSkeleton->getBodyNode(0)->getSkeleton();
+
+  std::vector<MetaSkeletonStateSpace::ScopedState> configurations;
+
+  // Use a ranker
+  ConstConfigurationRankerPtr configurationRanker(mConfigurationRanker);
```
aditya-vk: Why do we need to create another pointer here? Can we not do:
```
if (!mConfigurationRanker)
{
    mConfigurationRanker = NominalConfigurationRanker;
}
```
In `robot/utils/planToTSR`, the ranker has been taken with a `const` pointer so I guess that's fine there. Alternatively, we can leave this like this and change the data member to `const` since we are initializing it in the constructor anyway.

gilwoolee: Summarizing my slack discussion with @aditya-vk , I think we should keep it as is, for the following reasons:
1) if the planner is created with `mConfigurationRanker = nullptr`, the expectation is that a NominalConfigurationRanker would be created when this method is called and be initialized with the `startState`. So if we change the member variable as you suggested, the planner now has been mutated and it cannot be reused in the same way. Next time it is called, it'd use this ranker. 
2) To fix, the ideal solution would be the second way you suggested, to have it as `const mConfigurationRanker` so that it never gets changed. Unfortunately, to my best understanding, because this class _becomes_ a TSRPlanner via the templated adapter, it does not have access to `mConfigurationRanker` during the initialization, i.e. I can't do `: mConfigurationRanker(std::move(ranker)) { ... } `. It only has access to the constructor parameters of the PlannerAdapter, which is the `stateSpace` and `metaSkeleton`. See [stackoverflow](https://stackoverflow.com/questions/2290733/initialize-parents-protected-members-with-initialization-list-c) .

Hence, in order to keep the member variable as-is (as if it's a const) _and_ initialize it in constructor if it's passed in, I think we need to keep it the way I have it now. 

@aditya-vk @brianhou  please confirm!

aditya-vk: That's right, there's no way to have it in the initialization list. Hmm.
I think what we have now is good then. Since we don't give an option to change the ranker at all unless a new planner is created, I think it should be fine :+1:

Issue #232, opened 2017-10-07
Make code formatting script as a CMake module


jslee02:
Currently, the code formatting CMake code is included in the main `CMakeLists.txt`. Since we may want to have this functionality for other projects, it would be good to make it as a CMake module file (like [Components.cmake](https://github.com/personalrobotics/aikido/blob/master/cmake/Components.cmake)) and disseminate it to other projects.

jslee02:
[Here](https://github.com/personalrobotics/chimera/blob/master/cmake/ClangFormat.cmake) is an example I did for Chimera. One possible improvement is allowing to configure ClangFormat version instead of using hard-coded version.

brianhou:
Resolved by #509.

PR #509, opened 2019-02-20
Make code formatting script as a CMake module

egordon:
Solve issue #232 

Added CMake module for Clang Format based on the one from Chimera. Can specify requested clang-format version.

***

- [ n/a ] Document new methods and classes
- [ x ] Format code with `make format`

- [ n/a ] Set version target by selecting a milestone on the right side
- [ n/a ] Summarize this change in `CHANGELOG.md`
- [ n/a ] Add unit test(s) for this change

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -49,36 +49,6 @@ if(CODECOV)
 endif()
 
 #==============================================================================
-# Automatic code formatting setup
-#
-define_property(GLOBAL PROPERTY CODE_FORMATTING_SOURCES
-  BRIEF_DOCS "Source files that are automatically formatted by clang-format."
-  FULL_DOCS "Source files that are automatically formatted by clang-format."
-)
-
-function(format_add_sources)
-  foreach(source ${ARGN})
-    if(IS_ABSOLUTE "${source}")
-      set(source_abs "${source}")
-    else()
-      get_filename_component(source_abs
-        "${CMAKE_CURRENT_LIST_DIR}/${source}" ABSOLUTE)
-    endif()
-    if(EXISTS "${source_abs}")
-      set_property(GLOBAL APPEND PROPERTY CODE_FORMATTING_SOURCES "${source_abs}")
-    else()
-      message(FATAL_ERROR
-        "Source file '${source}' does not exist at absolute path"
-        " '${source_abs}'. This should never happen. Did you recently delete"
-        " this file or modify 'CMAKE_CURRENT_LIST_DIR'")
-    endif()
-  endforeach()
-endfunction()
-
-file(GLOB_RECURSE aikido_headers "${CMAKE_SOURCE_DIR}/include/*.hpp")
-format_add_sources(${aikido_headers})
-
-#==============================================================================
 # Helper functions.
 #
 function(aikido_space_delimit output_variable input_list)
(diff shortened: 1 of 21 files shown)


cmake/ClangFormat.cmake
```diff
@@ -0,0 +1,79 @@
...
+  find_program(CLANG_FORMAT_EXECUTABLE NAMES clang-format-${clang_version})
+  if(NOT CLANG_FORMAT_EXECUTABLE)
+    message(STATUS "Looking for clang-format-${clang_version} - NOT found,"
+      "please install to enable automatic code formatting."
+    )
+    return()
+  endif()
+
+  message(STATUS "Found clang-format-${clang_version}.")
+endfunction()
```
jslee02: 7Suggestion: `clang_format_setup` and `clang_format_setup` can be merged by using [`cmake_parse_arguments`](https://cmake.org/cmake/help/v3.12/command/cmake_parse_arguments.html).

cmake/ClangFormat.cmake
```diff
@@ -0,0 +1,79 @@
...
+    message(STATUS "Looking for clang-format - NOT found, please install "
+      "clang-format to enable automatic code formatting."
+    )
+    return()
+  endif()
+
+  message(STATUS "Found clang-format.")
+endfunction()
+
+function(clang_format_setup_version clang_version)
```
jslee02: Nit: Add the line separator for consistency.

CMakeLists.txt
```diff
@@ -144,6 +115,7 @@ find_package(ompl REQUIRED)
 # TREAT_WARNINGS_AS_ERRORS.
 include(CompilerSettings)
 
+
```
jslee02: Nit: Remove whitespace

cmake/ClangFormat.cmake
```diff
@@ -0,0 +1,79 @@
...
+    WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+  )
+  add_custom_target(check-format
+    COMMAND ${CMAKE_COMMAND} -E echo "Checking ${formatting_files_length} files..."
+    COMMAND ${CMAKE_SOURCE_DIR}/tools/check_format.sh ${CLANG_FORMAT_EXECUTABLE} ${formatting_files}
+    COMMAND ${CMAKE_COMMAND} -E echo "Done."
+    DEPENDS ${CLANG_FORMAT_EXECUTABLE}
+    WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+  )
+endfunction()
```
jslee02: This function wouldn't work when `formatting_files` is empty. Consider having [this logic](https://github.com/personalrobotics/aikido/pull/509/files#diff-af3b638bc2a3e6c650974192a53c7291L314).

cmake/ClangFormat.cmake
```diff
@@ -0,0 +1,90 @@
+function(clang_format_setup)
+  cmake_parse_arguments(
+        CF_ARG # prefix
```
brianhou: It seems a little inconsistent to me that the prefix is `CF_ARG` but the exported executable is `CLANG_FORMAT_EXECUTABLE`. I'm having a hard time finding useful examples on the internet, does this reasonably follow CMake conventions @jslee02?

jslee02: CMake is not a good tool to find a good convention of using it. See this [list](https://github.com/onqtam/awesome-cmake). It's horrific. :scream:

Back to the question, it seems people uses lower caess for the prefix (e.g., here and here). Also, there is tendency of using upper cases for global variables and lower cases for local variables. By the way, `CLANG_FORMAT_EXECUTABLE` is set by `find_program` not by `cmake_parse_arguments`.

PR #517, opened 2019-03-06
Change NominalConfigurationRanker to store ScopedState

### LINKED ISSUE #515: PlanToTSR: Replace raw pointers with ScopedStates in Ranker
The current ranker uses the zero vector as the nominal configuration since the nominal config is taken as raw pointer in the ranker. Change this to ScopedState until we a) require the efficiency without introducing bugs and b) understand scoped state fully.

gilwoolee:
This fixes the bug in NominalConfigurationRanker, which keeps a pointer to a state that may become cleared out.

I tested this in `ada_demos` and printed `mNominalConfiguration` in `evaluateConfiguration` to see that it matches the value passed to the constructor.

Resolves #515.

***


gilwoolee:
@brianhou I added a test under `GivenNominalConfigurationOrderTest` to change the originally passed nominal state and check for the ranker to stay the same. Could you take another look?

PR #518, opened 2019-04-17
Fix convertShape to use getMeshPath

gilwoolee:
This PR fixes `rviz::convertShape(MeshShape...)` method to use `getMeshPath` instead of `getMeshUri` for `meshUri`.
When a mesh is loaded via MeshShape, `convertShape(MeshShape...)` gets called. Since `meshUri` is used by Rviz, it has to be a raw filepath, which is not the case when using `getMeshUri`.

This was never an issue with Herb or Ada because we use `urdfLoader` and it constructs AssimpMesh, which doesn't call this function. It was discovered while @vinitha910  was trying to load cozmo mesh directly using raw filepaths. 

It has been tested by constructing cozmo with MeshShapes and visualizing it in Rviz through aikido InteractiveViewer.

***

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,6 +76,7 @@
 * RViz
 
   * Fixed bug of not joining Viewer threads when stopping auto-update: [#463](https://github.com/personalrobotics/aikido/pull/463)
+  * Fixed bug of not passing raw file path to RViz when MeshShape is used: [#518](https://github.com/personalrobotics/aikido/pull/518)
 
 * IO
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,7 +76,7 @@
 * RViz
 
   * Fixed bug of not joining Viewer threads when stopping auto-update: [#463](https://github.com/personalrobotics/aikido/pull/463)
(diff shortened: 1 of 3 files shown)


src/rviz/shape_conversions.cpp
```diff
@@ -193,7 +193,7 @@ bool convertShape(
   marker->scale = convertEigenToROSVector3(shape.getScale());
 
   const aiScene* scene = shape.getMesh();
-  const std::string& meshUri = shape.getMeshUri();
+  const std::string& meshUri = shape.getMeshPath();
```
brianhou: Can we rename the variable to be `meshPath`?

CHANGELOG.md
```diff
@@ -76,6 +76,7 @@
 * RViz
 
   * Fixed bug of not joining Viewer threads when stopping auto-update: [#463](https://github.com/personalrobotics/aikido/pull/463)
+  * Fixed bug of not passing raw file path to RViz when MeshShape is used: [#518](https://github.com/personalrobotics/aikido/pull/518)
```
```suggestion
  * Fixed bug of not passing full file path to RViz when MeshShape is used: [#518](https://github.com/personalrobotics/aikido/pull/518)
```

PR #523, opened 2019-04-29
Support state snapshots with only named skeletons.

brianhou:
When taking a snapshot of the world, we might only care about certain skeletons. For example, when constructing a dataset of possible robot environments, we wouldn't want to preserve the robot's state since that will change during planning.

Efficiency concerns: this implementation uses a lot of `getSkeleton(name)` calls, each of which I think is linear in the number of Skeletons in the world.

***


include/aikido/planner/World.hpp
```diff
@@ -88,10 +88,21 @@ class World
   /// \return State
   World::State getState() const;
 
+  /// Returns the state of named skeletons in this World.
+  /// \param names Names of skeletons that should be in the state
+  /// \return State
+  World::State getState(std::vector<std::string>& names) const;
```
gilwoolee:
```suggestion
  World::State getState(const std::vector<std::string>& names) const;
```

include/aikido/planner/World.hpp
```diff
@@ -88,10 +88,21 @@ class World
...
-  /// \param State State to set this world to.
-  void setState(const World::State& State);
+  /// \param state State to set this world to.
+  void setState(const World::State& state);
+
+  /// Sets the state of named skeletons in this World to match State.
+  /// The caller of this method MUST LOCK the mutex of this World.
+  /// \param state State to set this world to.
+  /// \param names Names of skeletons in the state
+  void setState(const World::State& state, std::vector<std::string>& names);
```
gilwoolee:
```suggestion
  void setState(const World::State& state, const std::vector<std::string>& names);
```

src/planner/World.cpp
```diff
@@ -171,10 +171,27 @@ bool World::State::operator!=(const State& other) const
...
+World::State World::getState(std::vector<std::string>& names) const
+{
+  using ConfigFlags = dart::dynamics::Skeleton::ConfigFlags;
+
+  World::State state;
+
+  for (const auto& name : names)
+  {
+    state.configurations[name]
+        = getSkeleton(name)->getConfiguration(ConfigFlags::CONFIG_POSITIONS);
```
jslee02: This can segfault when the world doesn't contain `name` skeleton. It'd be safer to insert the configuration only when `getSkeleton(name)` is not `nullptr`.

src/planner/World.cpp
```diff
@@ -171,10 +171,27 @@ bool World::State::operator!=(const State& other) const
...
 
   for (const auto& skeleton : mSkeletons)
-    state.configurations[skeleton->getName()] = skeleton->getConfiguration();
+    names.push_back(skeleton->getName());
+
+  return getState(names);
+}
+
+//==============================================================================
+World::State World::getState(std::vector<std::string>& names) const
```
brianhou:
```suggestion
World::State World::getState(const std::vector<std::string>& names) const
```

src/planner/World.cpp
```diff
@@ -187,19 +204,28 @@ void World::setState(const World::State& state)
...
+  names.reserve(getNumSkeletons());
+
   for (const auto& skeleton : mSkeletons)
+    names.push_back(skeleton->getName());
+
+  setState(state, names);
+}
+
+//==============================================================================
+void World::setState(const World::State& state, std::vector<std::string>& names)
```
brianhou:
```suggestion
void World::setState(const World::State& state, const std::vector<std::string>& names)
```

Issue #363, opened 2018-03-22
Segfaults with OMPL (>=1.2.0) + GCC


jslee02:
[All of AIKIDO's tests for OMPL segfault with OMPL 1.2.0](https://ci.appveyor.com/project/jslee02/aikido/build/91/job/37puse8agadeecb5#L5845) or higher. Here is the build configuration:
* OS: Ubuntu Xenial
* Compiler: gcc 5.4.0 (xenial default)
* Boost: 1.59
* OMPL: 
  * working: 1.0.0 (xenial default), 1.1.1
  * not-working: 1.2.0, 1.2.1 (artful default), 1.3.0, 1.3.1, 1.3.2

I have tested different versions and binary packages (`libompl-dev` and `ros-indigo/kinetic/lunar-ompl`) and ones built from source, but only OMPL 1.2.0 or higher have the issue.

I couldn't pinpoint a line causes the segfault using a debugger. The debugger just points assembly code as
```asm
0x7ffff7de7620                   41 56                    push   %r14
0x7ffff7de7622  <+0x0002>        41 55                    push   %r13
0x7ffff7de7624  <+0x0004>        41 54                    push   %r12
0x7ffff7de7626  <+0x0006>        55                       push   %rbp
0x7ffff7de7627  <+0x0007>        53                       push   %rbx
        36 [1]	in dl-init.c
0x7ffff7de7628  <+0x0008>        0f b6 87 14 03 00 00     movzbl 0x314(%rdi),%eax
        39 [1]	in dl-init.c
0x7ffff7de762f  <+0x000f>        4c 8b 47 08              mov    0x8(%rdi),%r8
        36 [2]	in dl-init.c
0x7ffff7de7633  <+0x0013>        83 c8 08                 or     $0x8,%eax
0x7ffff7de7636  <+0x0016>        88 87 14 03 00 00        mov    %al,0x314(%rdi)
        39 [2]	in dl-init.c
0x7ffff7de763c  <+0x001c>        45 0f b6 08              movzbl (%r8),%r9d
0x7ffff7de7640  <+0x0020>        45 84 c9                 test   %r9b,%r9b
0x7ffff7de7643  <+0x0023>        0f 84 cf 00 00 00        je     0x7ffff7de7718 <call_init+248>
        44 [1]	in dl-init.c
0x7ffff7de7649  <+0x0029>        48 8b 87 a0 00 00 00     mov    0xa0(%rdi),%rax
[...]

PR #524, opened 2019-04-30
Fix Travis CI builds

egordon:
Update DART version to match PPA (unblocks other repos like `libada` that are dealing with an interface change).

***

- [ N/A ] Set version target by selecting a milestone on the right side


gilwoolee:
LGTM! @egordon  could you update CHANGELOG.md as well?

egordon:
@gilwoolee Note that this update breaks Travis (since trusty doesn't have DART v6.8.2). I'm going to see if Travis plays nice with xenial.

src/constraint/dart/InverseKinematicsSampleable.cpp
```diff
@@ -221,7 +221,7 @@ bool IkSampleGenerator::sample(statespace::StateSpace::State* _state)
     mInverseKinematics->getTarget()->setTransform(poseState.getIsometry());
 
     // Run the IK solver. If it succeeds, return the solution.
```
aditya-vk:
```suggestion
    // Run the IK solver. If an exact solution is computed, apply it to the skeleton.
```

src/constraint/dart/InverseKinematicsSampleable.cpp
```diff
@@ -221,7 +221,7 @@ bool IkSampleGenerator::sample(statespace::StateSpace::State* _state)
...
-    if (mInverseKinematics->solve(true))
+    if (mInverseKinematics->solveAndApply(false))
```
jslee02: The behavior with this change will be different from the original behavior. What is the expected behavior here?

egordon: Wait, @jslee02 , so in the original behavior, even incomplete solutions were applied to the skeleton?

jslee02: Yes, that's the motivation of the new APIs to be more clear on the behavior.

.travis.yml
```diff
@@ -1,4 +1,4 @@
-dist: trusty
+dist: xenial
```
jslee02: Xenial build still has an issue: https://github.com/personalrobotics/aikido/issues/363

src/constraint/dart/InverseKinematicsSampleable.cpp
```diff
@@ -220,8 +220,8 @@ bool IkSampleGenerator::sample(statespace::StateSpace::State* _state)
 
     mInverseKinematics->getTarget()->setTransform(poseState.getIsometry());
 
-    // Run the IK solver. If it succeeds, return the solution.
-    if (mInverseKinematics->solve(true))
+     // Run the IK solver. If an exact solution is computed, apply it to the skeleton.
```
jslee02:
```suggestion
    // Run the IK solver. If an exact solution is computed, apply it to the skeleton.
```

src/constraint/dart/InverseKinematicsSampleable.cpp
```diff
@@ -220,8 +220,8 @@ bool IkSampleGenerator::sample(statespace::StateSpace::State* _state)
...
+    if (mInverseKinematics->solveAndApply(true))
```
jslee02: It seems AIKIDO still support Ubuntu 14.04. The latest DART version available in Ubuntu 14.04 is 6.6.2. I would change this to be backward compatible something like:
```cpp
#if DART_VERSION_AT_LEAST(6,8,0)
    if (mInverseKinematics->solveAndApply(true))
#else
    if (mInverseKinematics->solve(true))
```

.travis.yml
egordon: Double-check that this is still an issue in CI, since I believe in the launch scripts I forced it to pull OMPL 1.0 rather than 1.2 to get around this.

jslee02:
The CI tests are failing because it uses deprecated `::dart::common::make_unique`. It's deprecated as DART increased the minimum required C++ standard to C++14 since DART 6.9.0.

I suggest following possible solutions:
(1) increasing the minimum required C++ standard to C++14 for AIKIDO as well and replacing all the use of `::dart::common::make_unique` with `::std::make_unique`
(2) adding `::aikido::common::make_unique` that is equivalent to `::dart::common::make_unique` and use it
(3) suppressing warnings of using deprecated APIs

jslee02:
@egordon @brianhou The `make_unique` issue is resolved by #532, and Travis CI is happy now. Please take a look this CR when you guys get a chance.

.ci/install_linux_cmake.sh
```diff
@@ -0,0 +1,87 @@
...
+$SUDO apt-get -y install \
+  libtinyxml2-dev \
+  libyaml-cpp-dev
+
+if [ "$BUILD_AIKIDOPY" = "ON" ]; then
+  $SUDO apt-get -y install python3-dev python3-numpy
+  $SUDO apt-get -y install python3-pip -y
+  $SUDO pip3 install pytest -U
+
+  if [ $(lsb_release -sc) = "trusty" ]; then
```
brianhou: Is there a reason that the `trusty`, `xenial`, and `bionic` cases are duplicated? (And `cosmic` and `disco`.)

jslee02: Nope. Just for the case that they require different scripts. Feel free to merge them into two groups.

brianhou:
Just to be clear, it looks like this leaves

- TRUSTY_CATKIN_RELEASE
- TRUSTY_CATKIN_FULL_DEBUG
- TRUSTY_CATKIN_FULL_RELEASE
- XENIAL_CATKIN_FULL_RELEASE
- BIONIC_CATKIN_FULL_RELEASE

in the allowed failures list. That's fine for now, but we should continue to discuss which ones we care about.

.travis.yml
brianhou: Looks like it's still failing? https://travis-ci.org/personalrobotics/aikido/jobs/559769715

jslee02:
It seems the new macOS image is unstable. The last commit is just to retrigger with a new instance.

PR #527, opened 2019-05-26
Fix multiple CI issues

jslee02:
Currently, all the CI tests are failing due to several issues, which may need to be addressed one by one. It seems mostly they're related to building with ROS packages. 

This PR is just to add CMake build option (without using Catkin) so that we can test ROS agnostic AIKIDO components at least. All the catkin builds are disabled (put them to the `allowed_failure` section). Only building with CMake and building on macOS are working fine.

This PR is based on and targetting to #524. Feel free to modify this PR directly.

***


egordon:
@jslee02 We can do this testing inside of egordon-dart-6.8.2, since I don't plan on making any other changes to that branch.

PR #532, opened 2019-07-02
Remove use of deprecated dart::common::make_unique

jslee02:
***


jslee02:
It seems two build tests are failing:
- The macOS build fails due to failure on installing dependencies.
- The documentation build fails for unknown issue.

I'd move the failing builds to the allowed failures list for now and fix them later.

.ci/script_cmake.sh
```diff
@@ -8,4 +8,4 @@ mkdir build
 cd build
 cmake -DCMAKE_BUILD_TYPE=${BUILD_TYPE} -DTREAT_WARNINGS_AS_ERRORS=ON ..
 make -j4 tests
-make test
+ctest
```
brianhou: Why?

jslee02: Just to see if it makes any difference that turned out not. Feel free to revert it.

jslee02:
Yes, exactly!

PR #538, opened 2019-08-06
Fix compilation error (maybe only on Trusty?)

brianhou:
@planc509 was having this compilation error, which may have been introduced in [#499](https://github.com/personalrobotics/aikido/pull/499/files#diff-aa512357e2a323b6efdea83791638706L442). I have the same issue on my Trusty machine.

@gilwoolee @jslee02 do you remember why this change was made? (We should be deprecating this function anyway, so maybe it doesn't matter...)

***


brianhou:
Good enough for me, thanks!

Issue #534, opened 2019-07-23
Caching a trajectory


gilwoolee:
We need to design a way to cache and load various types of trajectories so that we can save redundant planning/retiming.

For example, OpenRAVE supports [serialization/deserialization](http://openrave.org/docs/0.8.2/architecture/trajectory/) which we utilized in [prpy](https://github.com/personalrobotics/prpy/blob/master/src/prpy/rave.py#L163) and had "waving" trajectories in [herbpy](https://github.com/personalrobotics/herbpy/blob/ab48e9190b061759b31bc9c879a7f96a51d975f5/src/herbpy/action/rogue.py#L303) that could be directly executed.

I think we should discuss which features to support or improve from the OpenRave API. 

1. Planner fills the trajectory class with timestamps, interpolation methods, and waypoints. It should be up to user to select trajectory class that will best execute this data.
2. Users can create, append, retime, and serialize trajectories both in C++ and Python.
3. Store arbitrary animations of an environment. Support for any configuration space similar to PlannerParameters so that multiple bodies and affine transformations are supported. For example, a robot opening door requires the robot joints and the door to move together.
4. Sample a trajectory at any time. Easily set the scene as the trajectory dictates it. How would this affect robot controllers?
5. It should be possible to play back the trajectory classes without loading any robots or any environment data. A real controller, on robot side, can link with openrave-core and use the trajectory class for interpolating incoming trajectory data.

Here are potential discussion points:
1. Does it make sense for us to give flexibility to the user to select trajectory class?
2. I think we should support the C++ version. Shall the serialization be in yaml format? 
3. Maybe we can defer this for now and support only single robot, and leave it as future feature.
4. Not sure what it means by "easily set the scene as the trajectory dictates it"
5. This may be a neat feature. We should be able to directly send a cached timed trajectory to a real-robot through ros controller. 

@siddhss5 @brianhou @aditya-vk @sniyaz  Please join the discussion.

siddhss5:
Is there any reason why we should not copy-exact the OpenRAVE format? Do we need to invent something new?

gilwoolee:
We could, it supports features that we may not need or re-think, which is why I raised questions above. I'd like to make the first version as minimal as possible instead of supporting potentially unnecessary (or incorrect) things.

siddhss5:
I do like the concept of (5), of being able to send _trajectories_ straight to `ros control`. Combined with easy recording of teleop trajectories, this can allow us to quickly experiment with things.

[...]

PR #541, opened 2019-08-16
Add methods for caching and retrieving trajectories

madan96:
This PR addresses #534 and adds methods to save and load a spline trajectory using YAML file. The methods are implemented as utility functions under `src/trajectory`.

This is an example of how a saved trajectory file looks like:
```
start_time: 0
dofs: 7
traj_order: 3
num_segments: 2
seg_0:
  coefficients: [4.07281e-20,0,-0.00101443,-2.925e-14,3.69597e-18,3.90809e-17,-1,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]
  duration: 0.1875
  start_state: [3.14159,7.38502e-17,0,0,0,0,0]
seg_1:
  coefficients: [3.35408e-20,-0.000380411,-0.00101443,4.69822e-14,1.47196e-17,-0.375,-1,2.39233e-14,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]
  duration: 0.1875
  start_state: [3.14156,-0.0351562,0,0,0,0,0]
```
I have added a unit test to check trajectory loaded from a saved YAML file has the same member variables as that of the original trajectory. Additionally, as a sanity check, also tested the methods with [herb3_demos/simple_trajectories](https://github.com/personalrobotics/herb3_demos/tree/master/simple_trajectories) by loading saved trajectories for execution and visualization it in RViz.

***


src/trajectory/util.cpp
```diff
@@ -308,5 +309,72 @@ UniqueInterpolatedPtr toR1JointTrajectory(const Interpolated& trajectory)
...
+  std::size_t numSegments = trajFile["num_segments"].as<std::size_t>();
+
+  auto Trajectory
+      = ::aikido::common::make_unique<Spline>(stateSpace, startTime);
+
+  for (std::size_t i = 0; i < numSegments; ++i)
+  {
+    auto segment_id = "seg_" + std::to_string(i);
+    Eigen::MatrixXd coeffVec
+        = trajFile[segment_id]["coefficients"].as<Eigen::MatrixXd>();
```
brianhou: Why `MatrixXd` rather than `VectorXd`? It seems to just be a vector in the yaml file?

madan96: No particular reason, while implementing I may have associated `MatrixXd` with all coefficient related variables I used. But yes, having this as `VectorXd` definitely makes more sense.

madan96:
* I thought to have this implementation in `aikido::trajectory` since it is something very specific to trajectories and sounded more like a utility function provided along with trajectory. Also, I'm not too familiar with `aikido::io` but I was of the impression that it has classes/methods that are generic and used by multiple submodules in aikido.

* I didn't know about YAML emitters but looks really helpful. I think I can definitely rewrite the YAML related part using this and helper function implemented in `aikido/io/detail/yaml_extension.hpp`

The format that you've suggested actually looks great and is more clearer. Would the entries under `data` block be indexed? This was one of the reasons why I used `num_segments` and segment names since I couldn't figure out a *simple* way to parse all the segments without knowing the exact number while loading.

brianhou:
It seems that `aikido::io` mainly defines convenience functions for robot library or demo-level scripts, not other AIKIDO modules. The current [dependencies](https://github.com/personalrobotics/aikido/blob/master/src/CMakeLists.txt#L8) seem to suggest that only `perception` and `robot` depend on `io` at the moment, neither of which are dependencies for the core planning modules.

So, I think I'm still inclined to put this with other deserialization code (for raw YAML, kinbody, URDF formats) in `io`. It means that `io` is the main component that depends on yaml-cpp. @gilwoolee any thoughts?

I think after yaml-cpp parses the sequence for you, you should be able to then query for the size of that sequence just like any other C++ container :)

include/aikido/io/trajectory.hpp
```diff
@@ -0,0 +1,33 @@
...
+
+/// Saves a timed trajectory
+///
+/// Given a spline trajectory \c_traj, object state space and trajectory
+/// file path, saves the trajectory as a yaml file for reuse later.
+/// \param[in] trajectory Spline trajectory
+/// \param[in] skelSpace Metaskeleton of the object
+/// \param[in] savePath save path for the trajectory yaml file
+void saveTrajectory(const aikido::trajectory::Spline& trajectory,
+    const aikido::statespace::dart::MetaSkeletonStateSpacePtr& skelSpace,
```
brianhou: Is this parameter necessary? I think it should be read from the trajectory.

include/aikido/io/trajectory.hpp
```diff
@@ -0,0 +1,33 @@
+#ifndef AIKIDO_IO_TRAJECTORY_HPP_
+#define AIKIDO_IO_TRAJECTORY_HPP_
+
+#include "aikido/trajectory/Spline.hpp"
+#include "aikido/statespace/dart/MetaSkeletonStateSpace.hpp"
+
+namespace aikido {
+namespace io {
+
+/// Saves a timed trajectory
```
brianhou: I don't think we need this full summary. I'd just rephrase as "Serializes a spline trajectory to YAML." before listing the parameters.

include/aikido/io/trajectory.hpp
```diff
@@ -0,0 +1,33 @@
...
+/// Given a spline trajectory \c_traj, object state space and trajectory
+/// file path, saves the trajectory as a yaml file for reuse later.
+/// \param[in] trajectory Spline trajectory
+/// \param[in] skelSpace Metaskeleton of the object
+/// \param[in] savePath save path for the trajectory yaml file
+void saveTrajectory(const aikido::trajectory::Spline& trajectory,
+    const aikido::statespace::dart::MetaSkeletonStateSpacePtr& skelSpace,
+    const std::string& savePath);
+
+/// Load spline trajectory from yaml file
```
brianhou: Similarly, I'd just say that this "Deserializes a spline trajectory from YAML.".

src/io/trajectory.cpp
```diff
@@ -0,0 +1,98 @@
...
+using aikido::statespace::StateSpacePtr;
+using aikido::statespace::dart::MetaSkeletonStateSpacePtr;
+using aikido::trajectory::Spline;
+using aikido::trajectory::UniqueSplinePtr;
+
+namespace aikido {
+namespace io {
+
+void saveTrajectory(const aikido::trajectory::Spline& trajectory,
+    const MetaSkeletonStateSpacePtr& skelSpace, const std::string& savePath)
```
brianhou: As above, I think you should be able to just pull the statespace from the trajectory. I would just try to cast it to a `MetaSkeletonStateSpace` to get the properties you need.

src/io/trajectory.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+//==============================================================================
+UniqueSplinePtr loadSplineTrajectory(
+    const std::string& trajPath, const MetaSkeletonStateSpacePtr& stateSpace)
+{
+  YAML::Node trajFile = YAML::LoadFile(trajPath);
+  const YAML::Node& config = trajFile["configuration"];
+
+  double startTime = config["start_time"].as<double>();
+  std::vector<std::string> dofs = config["dofs"].as<std::vector<std::string>>();
```
brianhou: Should check that this matches the dof names from the statespace.

src/io/trajectory.cpp
```diff
@@ -0,0 +1,98 @@
...
+  std::string trajType = config["type"].as<std::string>();
```
brianhou: I'd check that this is `spline`.

brianhou: Since we're assuming the spline order is fixed, we should probably check that as well.

src/io/trajectory.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+  auto Trajectory
```
brianhou:
```suggestion
  auto trajectory
```

src/io/trajectory.cpp
```diff
@@ -0,0 +1,98 @@
...
+      = ::aikido::common::make_unique<Spline>(stateSpace, startTime);
+
+  const YAML::Node& segments = trajFile["data"];
+  for (std::size_t i = 0; i < segments.size(); i++) {
```
brianhou: The open brace should be on the next line.

brianhou: Can we iterate directly over `segments` rather than by index?

src/io/trajectory.cpp
```diff
@@ -0,0 +1,98 @@
...
+    const YAML::Node& segment = segments[i];
+    Eigen::MatrixXd coefficients = segment["coefficients"].as<Eigen::MatrixXd>();
+    double duration = segment["duration"].as<double>();
+    Eigen::VectorXd position = segment["start_state"].as<Eigen::VectorXd>();
+
+    // Convert position Eigen vector to StateSpace::State*
+    aikido::statespace::StateSpace::State* startState
```
brianhou: @aditya-vk is this correct usage? Do we need to free the state?

src/io/trajectory.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+void saveTrajectory(const aikido::trajectory::Spline& trajectory,
+    const MetaSkeletonStateSpacePtr& skelSpace, const std::string& savePath)
+{
+  std::ofstream file(savePath);
+  YAML::Emitter emitter;
+
+  statespace::ConstStateSpacePtr trajSpace = trajectory.getStateSpace();
+  emitter << YAML::BeginMap;
+  emitter << YAML::Key << "configuration" << YAML::BeginMap;
```
brianhou: Put the `YAML::BeginMap` on a new line?

src/io/trajectory.cpp
```diff
@@ -0,0 +1,98 @@
...
+  emitter << YAML::Key << "start_time" << YAML::Value << trajectory.getStartTime();
+  emitter << YAML::Key << "dofs" << YAML::Flow << skelSpace->getProperties().getDofNames();
+  emitter << YAML::Key << "type" << YAML::Value << "spline";
+  emitter << YAML::Key << "spline";
+  emitter << YAML::BeginMap;
+  emitter << YAML::Key << "order" << YAML::Value << trajectory.getNumDerivatives();
+  emitter << YAML::EndMap;
+  emitter << YAML::EndMap << YAML::Key << "data" << YAML::BeginSeq;
```
brianhou: Put the `YAML::Key << "data" << YAML::BeginSeq` on two more lines? I'd also move this right before the for loop.

include/aikido/io/trajectory.hpp
```diff
@@ -0,0 +1,33 @@
+#ifndef AIKIDO_IO_TRAJECTORY_HPP_
+#define AIKIDO_IO_TRAJECTORY_HPP_
+
+#include "aikido/trajectory/Spline.hpp"
+#include "aikido/statespace/dart/MetaSkeletonStateSpace.hpp"
+
```
brianhou: Maybe this would be a good place to document the YAML format that we're serializing to and that we're expecting when deserializing.

include/aikido/io/trajectory.hpp
```diff
@@ -0,0 +1,33 @@
...
+void saveTrajectory(const aikido::trajectory::Spline& trajectory,
+    const aikido::statespace::dart::MetaSkeletonStateSpacePtr& skelSpace,
+    const std::string& savePath);
+
+/// Load spline trajectory from yaml file
+///
+/// Given trajectory file and trajectory state space, this method parses
+/// the trajectory file and loads a timed trajectory for direct execution.
+/// \param[in] trajPath Spline trajectory
+/// \param[in] stateSpace Metaskeleton for the trajectory
```
aditya-vk: If we expect the stateSpace to be associated with a MetaSkeleton, I would rename it to `metaSkeletonStateSpace`. Additionally, `MetaSkeleton` -> `MetaSkeletonStateSpace`.

include/aikido/io/trajectory.hpp
```diff
@@ -0,0 +1,33 @@
...
+/// \return Loaded spline trajectory
+aikido::trajectory::UniqueSplinePtr loadSplineTrajectory(const std::string& trajPath,
+    const aikido::statespace::dart::MetaSkeletonStateSpacePtr& stateSpace);
+} // namespace io
+} // namespace aikido
+
+#endif // AIKIDO_IO_TRAJECTORY_HPP_
```
aditya-vk: Add EOF line.

src/CMakeLists.txt
```diff
@@ -5,14 +5,14 @@ add_subdirectory("external/hauser_parabolic_smoother")
...
-add_subdirectory("io")         # [common], boost, dart, tinyxml2, yaml-cpp
 add_subdirectory("statespace") # dart
-add_subdirectory("perception") # [io], boost, dart, yaml-cpp, geometry_msgs, roscpp, std_msgs, visualization_msgs
 add_subdirectory("distance")   # [statespace], dart
 add_subdirectory("trajectory") # [common], [statespace]
 add_subdirectory("constraint") # [common], [statespace]
 add_subdirectory("planner")    # [external], [common], [statespace], [trajectory], [constraint], [distance], dart, ompl
 add_subdirectory("rviz")       # [constraint], [planner], boost, dart, roscpp, geometry_msgs, interactive_markers, std_msgs, visualization_msgs, libmicrohttpd
 add_subdirectory("control")    # [statespace], [trajectory]
+add_subdirectory("io")         # [common], boost, dart, tinyxml2, yaml-cpp
```
aditya-vk: Specify `[trajectory]` component

src/io/trajectory.cpp
aditya-vk: If you do the above, I suggest using `MetaSkeletonStateSpace` specific functions to make the log/exp conversions.

src/io/trajectory.cpp
aditya-vk: Definitely. Needs a `freeState()` at the end of the function.

src/io/trajectory.cpp
```diff
@@ -0,0 +1,98 @@
...
+        = stateSpace->allocateState();
+    stateSpace->expMap(position, startState);
+
+    Trajectory->addSegment(coefficients, duration, startState);
+  }
+  return Trajectory;
+}
+
+} // namespace io
+} // namespace aikido
```
aditya-vk: Add EOF line.

tests/io/test_trajectory_io.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+    Eigen::VectorXd originalPosition(stateSpace->getDimension());
+    Eigen::VectorXd loadedPosition(stateSpace->getDimension());
+    stateSpace->logMap(
+        originalSmoothTrajectory->getSegmentStartState(i), originalPosition);
+    stateSpace->logMap(
+        loadedSmoothTrajectory->getSegmentStartState(i), loadedPosition);
+    EXPECT_TRUE(loadedPosition.isApprox(originalPosition));
+  }
+}
```
aditya-vk: Add EOF line.

src/io/trajectory.cpp
madan96: Yes I guess we can use an iterator (with `for` loop) over the YAML node. Would that be better?

src/io/trajectory.cpp
madan96: In the current version it isn't really fixed, but we can definitely compare the order in yaml with the order of initialized trajectory. Since they both should match based on the stateSpace param we supply.

src/io/trajectory.cpp
madan96: @aditya-vk  Is there any particular reason for that? Because I think using MetaSkeletonStateSpace specific function would first require type-casting the state that we pass to these functions.

src/io/trajectory.cpp
brianhou: Yep!

madan96:
I think this is ready for the final review! I haven't run `make format` since it might make unwanted changes as I've a different version of `clang-format` on my (bionic) system.

PR #546, opened 2019-08-28
Added proper SO2 handling for Spline conversion

egordon:
Tested on entire feeding demo, both in simulation and on the real robot. Will run unit tests before merging.

To clarify the issue:

Since the Spline conversion was happening in R1, during wrap, the interpolation would actually add waypoints going the long-way around (i.e. if we send 2 waypoints pi and -pi as R1 over ROS, the reconstructed spline would go from -pi through 0 to pi, rather than just leaving them be).

This would trip the velocity constraints in `rewd_controllers`.

While there a a fix en route in `rewd_controllers` (https://github.com/personalrobotics/rewd_controllers/pull/34), it makes sense to add a fix here too, in case the controller isn't as intelligent as, say, `ros_trajectory_controller`.

***


src/control/ros/Conversions.cpp
```diff
@@ -459,7 +460,16 @@ std::unique_ptr<SplineTrajectory> toSplineJointTrajectory(
...
+    auto jointSpace = space->getJointSpace(citer);
+    auto r1Joint = std::dynamic_pointer_cast<const R1Joint>(jointSpace);
+    if (!r1Joint)
+    {
+      subspaces.emplace_back(std::make_shared<aikido::statespace::SO2>());
+    }
+    else
+    {
+      subspaces.emplace_back(std::make_shared<aikido::statespace::R1>());
+    }
```
aditya-vk: Why do we need to construct another space when we have the metaskeleton statespace which is R + SO(2)?

src/control/ros/Conversions.cpp
```diff
@@ -1,5 +1,6 @@
 #include "aikido/control/ros/Conversions.hpp"
 
+#include <iostream>
```
aditya-vk: This and the print statements can be removed now.

src/control/ros/Conversions.cpp
egordon: @aditya-vk @brianhou Since the Metaskeleton statespaces are JointSpaces, I believe, they need to be converted / cast. Is there a better way to do this?

egordon:
@aditya-vk The point of this PR is the reverse. I don't touch the code that *sends* trajectories. Instead, this code takes an incoming R^n trajectory and re-interprets it as R^n+SO(2)^m based on the MetaSkeleton. It makes receiving more robust while leaving sending unaffected.

egordon:
@gilwoolee updated PR description.

src/control/ros/Conversions.cpp
```diff
@@ -488,6 +501,15 @@ std::unique_ptr<SplineTrajectory> toSplineJointTrajectory(
         unspecifiedMetaSkeletonJoints,
         startPositions);
 
+    // Calculate diffVec for Spline conversion
+    Eigen::VectorXd diffPosition;
```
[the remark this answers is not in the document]
egordon: Added #562

PR #547, opened 2019-08-29
Upgrade clang-format version

madan96:
Switch to clang-format-6.0 

***

### DIFF
--- a/include/aikido/common/StepSequence.hpp
+++ b/include/aikido/common/StepSequence.hpp
@@ -92,11 +92,12 @@ private:
   std::size_t mNumSteps;
 };
 
-class StepSequence::const_iterator : public boost::iterator_facade<
-                                         StepSequence::const_iterator,
-                                         double,
-                                         boost::forward_traversal_tag,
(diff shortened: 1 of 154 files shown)


include/aikido/common/StepSequence.hpp
```diff
@@ -92,11 +92,11 @@ class StepSequence
   std::size_t mNumSteps;
 };
 
-class StepSequence::const_iterator
-    : public boost::iterator_facade<StepSequence::const_iterator,
```
brianhou: One possible additional option to set: `BreakBeforeInheritanceComma: true`. I think that would bring the code back closer to this, although then we'd have commas before all the classes that are being inherited (similar to what it looks like with initializer lists).

madan96: With `BreakBeforeInheritanceComma: true` it looks better in terms of readability. So should I add go ahead with adding it?

brianhou: Sounds good, let's go for it.

PR #551, opened 2019-09-04
Remove Trusty and Xenial builds.

brianhou:
Trusty is no longer supported and we're skipping Xenial, so there's no point in wasting Travis VMs to test on them.


brianhou:
The `BIONIC_CATKIN_FULL_RELEASE_AIKIDOPY` job seems to be running up against the 50 minute Travis limit. Maybe it makes sense to migrate this job to Azure @colinxs @jslee02?

jslee02:
:+1:  I've been using Azure (in parallel with Travis CI) for DART, and it's very satisfying.

PR #552, opened 2019-09-13
BugFix: Van der Corput - consider span for resolution check

aditya-vk:
Expected behavior for VDC with resolution `0.5`:
1. If an edge is of length(span) 1, output is `0, 0.5, 1`.
2. If the edge is of length(span) 2, output is `0, 0.5, 1.0, 1.5, 2.0`.

Current behavior generates sequence for [0,1] and multiplies with length thus giving
1. `0, 0.5, 1.0`
2. `0, 1.0, 2.0`

We have been using Van der Corput sequence only for interpolated trajectories with [0, 1] time so it was not an issue.

This PR introduces a test that fails with the current master and proposes a fix.

***

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
   * Cleaned up doxygen errors: [#357](https://github.com/personalrobotics/aikido/pull/357)
   * Fixed bug in compiling with Boost 1.58 on Kinetic + Xenial: [#490](https://github.com/personalrobotics/aikido/pull/490)
   * Fixed bug in Interpolated::addWaypoint(): [#483](https://github.com/personalrobotics/aikido/pull/483)
+  * Fixed bug in VanDerCorput sequence generator in handling non-unit Span: [#552](https://github.com/personalrobotics/aikido/pull/552)
 
 * Distance
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
   * Cleaned up doxygen errors: [#357](https://github.com/personalrobotics/aikido/pull/357)
   * Fixed bug in compiling with Boost 1.58 on Kinetic + Xenial: [#490](https://github.com/personalrobotics/aikido/pull/490)
   * Fixed bug in Interpolated::addWaypoint(): [#483](https://github.com/personalrobotics/aikido/pull/483)
-  * Fixed bug in VanDerCorput sequence generator in handling non-unit Span: [#552](https://github.com/personalrobotics/aikido/pull/552)
+  * Fixed bug in VanDerCorput sequence generator to handle non-unit span: [#552](https://github.com/personalrobotics/aikido/pull/552)
 
 * Distance
 
--- a/tests/common/test_VanDerCorput.cpp
+++ b/tests/common/test_VanDerCorput.cpp
@@ -99,16 +99,13 @@ TEST(VanDerCorput, FirstNineIteratorValuesWithEndpoints)
   EXPECT_DOUBLE_EQ(7. / 8, *itr);
 }
 
-TEST(VanDerCorput, ScaleProperlyOverSpan)
+TEST(VanDerCorput, ScaleProperlyOverSpanWithResolution)
 {
-  VanDerCorput vdc{2};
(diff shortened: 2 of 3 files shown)


tests/common/test_VanDerCorput.cpp
```diff
@@ -99,16 +99,13 @@ TEST(VanDerCorput, FirstNineIteratorValuesWithEndpoints)
...
-TEST(VanDerCorput, ScaleProperlyOverSpan)
-{
-  VanDerCorput vdc{2};
-  EXPECT_DOUBLE_EQ(2.0 / 2, vdc[0].first);
-  EXPECT_DOUBLE_EQ(2.0 / 4, vdc[1].first);
-  EXPECT_DOUBLE_EQ(6.0 / 4, vdc[2].first);
-  EXPECT_DOUBLE_EQ(2.0 / 8, vdc[3].first);
-  EXPECT_DOUBLE_EQ(10. / 8, vdc[4].first);
-  EXPECT_DOUBLE_EQ(6.0 / 8, vdc[5].first);
-  EXPECT_DOUBLE_EQ(14. / 8, vdc[6].first);
```
jslee02: Why not keeping this test? Isn't this still a valid test?

aditya-vk: I figured the replaced test would suffice for both. I brought this back and added a separate test for this bugfix.

aditya-vk:
@jslee02 fixed both the value as well as the resolution.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
   * Cleaned up doxygen errors: [#357](https://github.com/personalrobotics/aikido/pull/357)
   * Fixed bug in compiling with Boost 1.58 on Kinetic + Xenial: [#490](https://github.com/personalrobotics/aikido/pull/490)
   * Fixed bug in Interpolated::addWaypoint(): [#483](https://github.com/personalrobotics/aikido/pull/483)
+  * Fixed bug in VanDerCorput sequence generator in handling non-unit Span: [#552](https://github.com/personalrobotics/aikido/pull/552)
```
```suggestion
  * Fixed bug in VanDerCorput sequence generator to handle non-unit span: [#552](https://github.com/personalrobotics/aikido/pull/552)
```

tests/common/test_VanDerCorput.cpp
```diff
@@ -216,3 +216,37 @@ TEST(VanDerCorput, IterationEndsWhenMinimumResolutionReached)
...
+  });
+  EXPECT_EQ(7, iterations);
+  EXPECT_EQ(7, vdc.getLength());
+
+  // Check that with a non-unit span, the vdc values are correct.
+  EXPECT_DOUBLE_EQ(1.00, vdc[0].first);
+  EXPECT_DOUBLE_EQ(0.50, vdc[1].first);
+  EXPECT_DOUBLE_EQ(1.50, vdc[2].first);
+  EXPECT_DOUBLE_EQ(0.25, vdc[3].first);
+  EXPECT_DOUBLE_EQ(1.25, vdc[4].first);
```
brianhou: This isn't especially critical, but I would've expected the third level of the tree to be 0.25, 0.75, 1.25, 1.75 from the documentation. Why isn't that what we return?

aditya-vk: Where does it say that? That would be wrong, sequence jumps around: https://en.wikipedia.org/wiki/Van_der_Corput_sequence

aditya-vk: I was expecting the same or for it to be 0.25, 1.75, 0.75, 1.25 until I wrote the test and learned my lesson.

brianhou: Ah, I see. I was slightly mislead by [level-order traversal](https://github.com/personalrobotics/aikido/blob/84d78563360a86457bc220682c10fec1937b2f12/include/aikido/common/VanDerCorput.hpp#L14). It's technically correct (in that we're processing the whole level before advancing to the next one), but I've previously seen it as also processing each level from left to right.

I think it's fine as is to provide intuition for what the sequence does (and not a critical distinction that needs to be fixed).

Issue #477, opened 2019-01-03
Clean up the retimer methods in the robot namespace


aditya-vk:
Currently, it seems like we need a different method for each of the timers we have (kunz and parabolic). It'd be neater to have a single `retimePath` function that takes in timer as an argument and calls the corresponding `postprocess()` function.

PR #554, opened 2019-10-02
Use KunzRetimer by default.

brianhou:
This is a temporary hack to make HERB demos work on master. We need to be able to properly dispatch to different postprocessors from librobot-level code.

I'm not sure whether ADA is using a similar hack, so please check. This is related to #477.

***


aditya-vk:
Are you looking to have this merged? I think we should solve this issue correctly to keep master clean and let this be on a working branch if it's for demos.

brianhou:
Agreed, although I don't think we even have an API designed to solve this issue properly yet.

PR #560, opened 2019-12-30
Bugfix in InteractiveMarkerViewer

aditya-vk:
If the viewer has an underlying world, skeleton markers corresponding to skeletons in the world that have been removed continued to be visualized. When the world removes the skeleton, the `SkeletonPtr` is non-nullptr and therefore the control did not reach the line that erases the marker from the map. This PR reorganizes the code for correction.

Tested via `herb3_demos` scripts for addition, deletion and updates to skeletons.

@brianhou: As a side: I wonder if we should re-introduce the `WorldInteractiveMarkerViewer`, not in its previous state but just to reimplement the `updateSkeletonMarkers()` function to have fewer if conditions.  

***

### DIFF
--- a/src/rviz/InteractiveMarkerViewer.cpp
+++ b/src/rviz/InteractiveMarkerViewer.cpp
@@ -85,7 +85,7 @@ void InteractiveMarkerViewer::updateSkeletonMarkers()
   {
     // If there is no underlying world and the skeleton exists,
     // update the corresponding marker.
-    if (!mWorld && it->first) 
+    if (!mWorld && it->first)
     {
       std::unique_lock<std::mutex> skeleton_lock(
           it->first->getMutex(), std::try_to_lock);
--- a/src/rviz/InteractiveMarkerViewer.cpp
+++ b/src/rviz/InteractiveMarkerViewer.cpp
@@ -54,6 +54,63 @@ SkeletonMarkerPtr InteractiveMarkerViewer::addSkeletonMarker(
 }
 
 //==============================================================================
+void InteractiveMarkerViewer::updateSkeletonMarkers()
+{
+  // Update SkeletonMarkers from the world.
+  if (mWorld)
+  {
+    for (std::size_t i = 0; i < mWorld->getNumSkeletons(); ++i)
+    {
+      const dart::dynamics::SkeletonPtr skeleton = mWorld->getSkeleton(i);
+
+      if (skeleton)
+      {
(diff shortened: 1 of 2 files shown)


src/rviz/InteractiveMarkerViewer.cpp
```diff
@@ -64,6 +121,13 @@ FrameMarkerPtr InteractiveMarkerViewer::addFrameMarker(
   return marker;
 }
 
+//==============================================================================
+void InteractiveMarkerViewer::updateFrameMarkers()
+{
+  for (const FrameMarkerPtr& marker : mFrameMarkers)
```
brianhou:
```suggestion
  for (const auto& marker : mFrameMarkers)
```

src/rviz/InteractiveMarkerViewer.cpp
```diff
@@ -53,6 +53,63 @@ SkeletonMarkerPtr InteractiveMarkerViewer::addSkeletonMarker(
...
+
+  // Clear removed skeletons
+  auto it = std::begin(mSkeletonMarkers);
+  while (it != std::end(mSkeletonMarkers))
+  {
+    // If there is no underlying world and the skeleton exists,
+    // update the corresponding marker.
+    if (!mWorld && it->first)
+    {
+      std::unique_lock<std::mutex> skeleton_lock(
```
brianhou: As we discussed offline, I think this function might be clearer if we have the insert loop above (without updating), then do all the updating in this loop over `mSkeletonMarkers` since we're iterating over all of them anyway. I think this could also allow us to merge the increments/erases into a single if condition.

egordon:
@brianhou @aditya-vk This introduces a warning (if statement ambiguity) that breaks all downstream Travis CI calls (that treat warnings as errors):
```
/root/workspace/src/aikido/src/rviz/InteractiveMarkerViewer.cpp: In member function 'void aikido::rviz::InteractiveMarkerViewer::updateSkeletonMarkers()':
/root/workspace/src/aikido/src/rviz/InteractiveMarkerViewer.cpp:81:30: error: suggest parentheses around '&&' within '||' [-Werror=parentheses]
     if (!it->first || mWorld && !mWorld->hasSkeleton(it->first))
                       ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cc1plus: all warnings being treated as errors
```

Issue #467, opened 2018-11-01
Deprecate robot planning utils


brianhou:
Rather than maintaining two versions of our planning calls, we should deprecate https://github.com/personalrobotics/aikido/blob/master/src/robot/util.cpp in favor of the planner API.

PR #564, opened 2020-02-26
Add SequenceConfigurationToConfigurationMetaPlanner

gilwoolee:
This PR adds `SequenceConfigurationToConfigurationMetaPlanner`, in preparation for #467.

SequenceConfigurationToConfigurationMetaPlanner takes a list of `ConfigurationToConfigurationPlanner` and use them in sequence to solve a `ConfigurationToConfiguration` problem.

While a generic  `SequenceMetaPlanner` can be used to serve this purpose, we need this planner to be of type `ConfigurationToConfigurationPlanner` in order to be used by PlannerAdapter `ConfigurationToConfiguration_to_ConfigurationToTSR`. 

Later, we may want to make a templated `SequenceMetaPlanner<T>`, but I'd leave it as a future TODO as it adds complexity.

TODO. I will test after the initial PR review. 

***

### DIFF
--- a/include/aikido/planner/SequenceConfigurationToConfigurationMetaPlanner.hpp
+++ b/include/aikido/planner/SequenceConfigurationToConfigurationMetaPlanner.hpp
@@ -1,19 +1,20 @@
 #ifndef AIKIDO_PLANNER_SEQUENCECONFIGURATIONTOCONFIGURATIONMETAPLANNER_HPP_
 #define AIKIDO_PLANNER_SEQUENCECONFIGURATIONTOCONFIGURATIONMETAPLANNER_HPP_
 
-#include "aikido/planner/SequenceMetaPlanner.hpp"
 #include "aikido/planner/ConfigurationToConfigurationPlanner.hpp"
+#include "aikido/planner/SequenceMetaPlanner.hpp"
 
 namespace aikido {
 namespace planner {
 
-/// A meta planner that solves ConfigurationToConfiguration using the sub planners one-by-one
-/// sequentially and returns the first successfully planned trajectory.
-class SequenceConfigurationToConfigurationMetaPlanner :
-	public SequenceMetaPlanner, public ConfigurationToConfigurationPlanner
+/// A meta planner that solves ConfigurationToConfiguration using the sub
+/// planners one-by-one sequentially and returns the first successfully planned
+/// trajectory.
+class SequenceConfigurationToConfigurationMetaPlanner
+  : public SequenceMetaPlanner
+  , public ConfigurationToConfigurationPlanner
 {
 public:
-
   /// Constructs given list of planners.
   ///
   /// \param[in] stateSpace State space that this planner associated with.
@@ -28,9 +29,8 @@ public:
   // Documentation inherited.
   // trajectory::TrajectoryPtr plan(
   //     const Problem& problem, Result* result = nullptr) override;
-   trajectory::TrajectoryPtr plan(
-      const SolvableProblem& problem, Result* result = nullptr)
-      override;
+  trajectory::TrajectoryPtr plan(
+      const SolvableProblem& problem, Result* result = nullptr) override;
 };
 
 } // namespace planner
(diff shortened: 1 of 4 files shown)


brianhou:
I think the current way that we've been using this in `libherb` is to have a sequence of `[ConfigToConfig, ConfigToTSR]` where `ConfigToTSR` is an explicit adaptation of the `ConfigToConfig`.

It seems like the new way would be a sequence of `[[ConfigToConfig]s, [ConfigToTSR]s]`, where the first is of type `SequenceConfigurationToConfigurationMetaPlanner` and the second is an explicit adaption. Is that correct?

src/planner/SequenceConfigurationToConfigurationMetaPlanner.cpp
```diff
@@ -0,0 +1,35 @@
...
+        = std::dynamic_pointer_cast<ConfigurationToConfigurationPlannerPtr>(
+            planner);
+
+    if (!castedPlanner)
+      throw std::invalid_argument(
+          "One of the planners is not ConfigurationToConfigurationPlanner.");
+  }
+}
+
+// //==============================================================================
```
brianhou:
```suggestion
//==============================================================================
```

src/planner/SequenceConfigurationToConfigurationMetaPlanner.cpp
```diff
@@ -0,0 +1,35 @@
...
+//==============================================================================
+SequenceConfigurationToConfigurationMetaPlanner::
+    SequenceConfigurationToConfigurationMetaPlanner(
+        statespace::ConstStateSpacePtr stateSpace,
+        const std::vector<PlannerPtr>& planners,
+        common::RNG* rng)
+  : SequenceMetaPlanner(stateSpace, std::move(planners))
+  , ConfigurationToConfigurationPlanner(stateSpace, std::move(rng))
+{
+  for (auto planner : mPlanners)
```
brianhou:
```suggestion
  for (const auto& planner : mPlanners)
```

include/aikido/planner/SequenceConfigurationToConfigurationMetaPlanner.hpp
```diff
@@ -0,0 +1,39 @@
...
+  /// \param[in] rng RNG that planner uses. If nullptr, a default is created.
+  /// \throw If any of \c planners are not ConfigurationToConfigurationPlanner.
+  SequenceConfigurationToConfigurationMetaPlanner(
+      statespace::ConstStateSpacePtr stateSpace,
+      const std::vector<PlannerPtr>& planners = std::vector<PlannerPtr>(),
+      common::RNG* rng = nullptr);
+
+  // Documentation inherited.
+  // trajectory::TrajectoryPtr plan(
+  //     const Problem& problem, Result* result = nullptr) override;
```
brianhou: Why is this commented out?

gilwoolee: By mistake..

gilwoolee:
@brianhou I'm not sure I fully followed.

>  the current way we've been using this in libherb is to have a sequence of [ConfigToConfig, ConfigToTSR] where ConfigToTSR is an explicit adaptation of the ConfigToConfig.

Let me just share how I was initially hoping to use this. `ConcreteRobot` currently has two potential places for `SequenceMetaPlanner`. One is `PlanToConfiguration`, where we currently call `Snap` and `RRTConnect` in sequence. This can be replaced by a generic `SequenceMetaPlanner` ([code example](https://github.com/personalrobotics/aikido/blob/gilwoo/robot_util_deprecation/src/robot/ConcreteRobot.cpp#L425)).

The other place is for `PlanToTSR` (I think this is what you're referring to), where we (1) sample a bunch of configs (2) try all of them w/ Snap (3) try all of them w/ RRTConnect. This is where we _might_ need this, if we choose to follow the spec of PlanToTSR regarding the ranking. That is, we want to actually try our best w/ all possible planners per configuration before trying the next one, we'd need this planner. [Here](https://github.com/personalrobotics/aikido/blob/gilwoo/robot_util_deprecation/src/robot/ConcreteRobot.cpp#L513 
)'s how I initially thought we'd use it.

That said, it's against the fail-fast policy, so I don't think it'd be wise to do it this way for `PlanToTSR`. I still don't quite like the long code for PlanToTSR, so how about calling `SnapConfigurationToConfigurations` and `RRTConfigurationToConfigurations` in sequence?

Regardless, we'd need a similar solution to this if we want to pass any single problem meta planner into a `dart::PlannerAdapter`, which I thought was one of the main use cases meta planners. I'm happy to table this PR if unnecessary. :)

Issue #555, opened 2019-10-02
Formatting checks disabled (possible erroneous?)


aditya-vk:
Two issues:

1. Format checks are currently disabled. I think this was intentional during the shift to bionic and clang-format-6.0. We need to bring it back.

I am not sure if the following issue is local to me (@madan96 can you please verify)
2. On running the formatting command, #521 seems to still fail. For instance `src/rviz/InteractiveMarkerViewer.cpp` Line 206 is never corrected on running clang-format. I have clang-format-6.0 on bionic.

madan96:
Yeah, I just checked and there seem to be some src files in `rviz`, `perception` and `control` subdirectories that still require formatting changes. Not sure how they got left out, there is no PR after #549 that was merged into master after this so I think that's where the problem is.

madan96:
@brianhou Should we run [check-format](https://github.com/personalrobotics/aikido/blob/master/.ci/script_catkin.sh#L17) for all builds or at least `BIONIC_CATKIN_FULL_RELEASE`?

brianhou:
We only need to check the formatting on one build.

PR #565, opened 2020-02-26
Clang format

gilwoolee:
Somehow I see bunch of clang format changes so I'm creating a PR to address this. 
I'm on clang-format 6.0 so these seem to be correct? 

***


aditya-vk:
We can merge this but we do not have Travis setup for formatting checks. 

@madan96 can you take this over [see #555 for the related issue]. I think the check was disabled during the shift to clang-6.0.

PR #566, opened 2020-02-26
Add ConfigurationToTSRwithTrajectoryConstraint

gilwoolee:
In preparation for #467, this PR contains the following update:

- Add `planner::dart::ConfigurationToTSRwithTrajectoryConstraint`
- Add `planner::dart::CRRTConfigurationToTSRwithTrajectoryConstraintPlanner`
- Add `planner_dart` component and make relevant changes in CMakeLists files.

The last change is necessary because `CRRTConfigurationToTSRwithTrajectoryConstraintPlanner` needs to call `planner_ompl`, which needs to be built before this planner is compiled.

**Testing**
This PR passes existing unit tests. I'll add unit tests for this planner after the initial review.

***

### DIFF
--- a/include/aikido/planner/dart/CRRTConfigurationToTSRwithTrajectoryConstraintPlanner.hpp
+++ b/include/aikido/planner/dart/CRRTConfigurationToTSRwithTrajectoryConstraintPlanner.hpp
@@ -3,9 +3,9 @@
 
 #include "aikido/planner/dart/ConfigurationToTSRPlanner.hpp"
 #include "aikido/planner/dart/ConfigurationToTSRwithTrajectoryConstraint.hpp"
+#include "aikido/robot/util.hpp"
 #include "aikido/statespace/dart/MetaSkeletonStateSpace.hpp"
 #include "aikido/trajectory/Trajectory.hpp"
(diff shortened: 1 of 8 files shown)


brianhou:
I haven't thought about this very much, but I initially thought it might make sense to add trajectory constraints as a part of all `Problem`s rather than as a specific planner that can deal with them. What do you think? (I guess part of the challenge is that we don't have an reusable component that can plan with those constraints, so I could see an argument that we should just be pragmatic and call this a separate planner type until we do.)

Also, I think it would be good to separate the new `planner_dart` component into its own PR.

gilwoolee:
Thanks for the comments @brianhou !
I created another PR #567 as you suggested.

> it might make sense to add trajectory constraints as a part of all Problems rather than as a specific planner that can deal with them. What do you think? (I guess part of the challenge is that we don't have an reusable component that can plan with those constraints, so I could see an argument that we should just be pragmatic and call this a separate planner type until we do.)

Yep I agree on both of these. I mainly chose this way for the latter challenge you mentioned and also because this is the only constrained planning problem type we use now. I think it's worth revisiting for a more general solution once we have more of these constrained planning problem types.

PR #567, opened 2020-03-03
Add planner_dart component for DART planners

gilwoolee:
This PR creates `planner_dart` component and builds all DART planners as `planner_dart` component.
***

### DIFF
--- a/src/planner/dart/CMakeLists.txt
+++ b/src/planner/dart/CMakeLists.txt
@@ -1,0 +1,62 @@
+if(CMAKE_COMPILER_IS_GNUCXX)
+  if(OMPL_VERSION VERSION_GREATER 1.2.0 OR OMPL_VERSION VERSION_EQUAL 1.2.0)
+    if(Boost_VERSION VERSION_LESS 106501)
+      message(STATUS "OMPL planners are disabled for OMPL (>=1.2.0) + GCC + "
+          "Boost (< 1.65.1). For details, please see: "
+          " https://github.com/personalrobotics/aikido/issues/363"
+      )
+      return()
+    endif()
+  endif()
+endif()
+
+#==============================================================================
+# Libraries
+#
+set(sources
+  ConfigurationToConfiguration.cpp
+  ConfigurationToConfigurationPlanner.cpp
+  ConfigurationToConfiguration_to_ConfigurationToConfiguration.cpp
+  ConfigurationToConfiguration_to_ConfigurationToTSR.cpp
+  ConfigurationToEndEffectorOffset.cpp
+  ConfigurationToEndEffectorPose.cpp
+  ConfigurationToEndEffectorOffsetPlanner.cpp
+  ConfigurationToTSR.cpp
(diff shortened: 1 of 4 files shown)


sniyaz:
Hey @brianhou, can you take another look? Reviving this now, I've made some fixes so Travis should be satisfied.

src/planner/dart/CMakeLists.txt
```diff
@@ -0,0 +1,62 @@
+if(CMAKE_COMPILER_IS_GNUCXX)
+  if(OMPL_VERSION VERSION_GREATER 1.2.0 OR OMPL_VERSION VERSION_EQUAL 1.2.0)
+    if(Boost_VERSION VERSION_LESS 106501)
```
aditya-vk, line 3: We do not need to check boost version here since this subdirectory does not have ompl code.

src/planner/dart/CMakeLists.txt
```diff
@@ -0,0 +1,62 @@
...
+)
+target_link_libraries("${PROJECT_NAME}_planner_dart"
+  PUBLIC
+    "${PROJECT_NAME}_common"
+    "${PROJECT_NAME}_constraint"
+    "${PROJECT_NAME}_distance"
+    "${PROJECT_NAME}_trajectory"
+    "${PROJECT_NAME}_statespace"
+    "${PROJECT_NAME}_planner"
+    "${PROJECT_NAME}_planner_ompl"
```
aditya-vk, line 43: do we need ompl here?

aditya-vk (changes requested):
Looks mostly good! However, what is our current policy on creating a new component? Is it mostly organization dependent (in which case it does not have to be a component that can be compiled optionally) or is based on the dependencies involved? Whichever it is, we should document this. Could you create an issue to discuss, track and add to the our code guidelines?

src/planner/dart/CMakeLists.txt
sniyaz: This is related to the below comment as well: this PR is set up for https://github.com/personalrobotics/aikido/pull/566. Sorry, I totally should have made that more clear: because the ultimate goal here is deprecating a lot of the old `robot/util` stuff, we need to move some of the `planTo` methods that use ompl into this new `dart_planner` module. That's what https://github.com/personalrobotics/aikido/pull/566 starts to do, but we set up the ompl dependency here 😄

src/planner/dart/CMakeLists.txt
sniyaz: See previous comment! 😸

sniyaz:
@aditya-vk You have a good question about when to split off a new component. I think the motivation here is a combination of:

1. The amount of code in this component has grown large enough that splitting it off makes sense.
2. The dependency here on OMPL means that **none** of our planners would build if OMPL was not installed correctly (assuming we were using a single component for all planners like we have now).

If these motivations sound good, I'm happy to create an issue to track this and update our AIKIDO guidelines.

PR #568, opened 2020-05-15
Updated GTest to v1.11 and make submodule

egordon:
Our current version of gtest is v1.7, from 2013. We also bypass all OS checks in the gtest CMakeLists.txt by linking it manually.

This change updates gtest, switches it to using a git submodule (so we don't pollute our git tree), and uses its built-in CMakeLists.txt.

***


brianhou:
Bump -- still curious whether gtest can be an externally-configured submodule or something rather than explicitly including all the files here.

Issue #445, opened 2018-06-22
Change planning problem classes to hold ScopedState instead of State*


brianhou:
Otherwise, states in a `Problem` can become dangling pointers.

aditya-vk:
Addressed by #569

PR #569, opened 2020-05-15
Check that all planning problems only hold ScopedStates.

sniyaz:
See: https://github.com/personalrobotics/aikido/issues/445

I've gone through and checked that all planner problem classes hold only `ScopedState` instead of raw pointers (to prevent the dangling pointer issue mentioned above). 

It turns out the only class that needed fixing was `ConfigurationToConfigurations`. At some point, it looks like the other problems were moved over to holding `ScopedState`. This PR addresses this.

***

### DIFF
--- a/include/aikido/planner/ConfigurationToConfigurations.hpp
+++ b/include/aikido/planner/ConfigurationToConfigurations.hpp
@@ -51,7 +51,7 @@ public:
 
 protected:
   /// Start state.
-  const statespace::StateSpace::State* mStartState;
+  statespace::StateSpace::ScopedState mStartState;
 
   /// Goal States.
   const GoalStates mGoalStates;
--- a/include/aikido/planner/ConfigurationToConfigurations.hpp
+++ b/include/aikido/planner/ConfigurationToConfigurations.hpp
@@ -47,7 +47,7 @@ public:
   std::size_t getNumGoalStates() const;
 
   /// Returns goal states.
-  const GoalStates& getGoalStates() const;
+  const GoalStates getGoalStates() const;
 
 protected:
   /// Start state.
--- a/include/aikido/planner/ConfigurationToConfigurations.hpp
+++ b/include/aikido/planner/ConfigurationToConfigurations.hpp
@@ -18,7 +18,7 @@ namespace planner {
 class ConfigurationToConfigurations : public Problem
 {
 public:
-  using GoalStates = std::unordered_set<const statespace::StateSpace::State*>;
+  using GoalStates = std::vector<const statespace::StateSpace::State*>;
 
   /// Constructor.
   ///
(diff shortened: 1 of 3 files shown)


include/aikido/planner/ConfigurationToConfigurations.hpp
```diff
@@ -47,14 +47,14 @@ class ConfigurationToConfigurations : public Problem
   std::size_t getNumGoalStates() const;
 
   /// Returns goal states.
-  const GoalStates& getGoalStates() const;
+  const GoalStates getGoalStates() const;
 
 protected:
   /// Start state.
-  const statespace::StateSpace::State* mStartState;
+  statespace::StateSpace::ScopedState mStartState;
```
aditya-vk, line 54: Why remove the `const`? Looks like we are updating the member in the initializer list, so should be fine? 

That said, looks like the other `problem`s have it non-const, we can also address all the const-ness in another PR. Could you make an issue?

include/aikido/planner/ConfigurationToConfigurations.hpp
```diff
@@ -47,14 +47,14 @@ class ConfigurationToConfigurations : public Problem
   std::size_t getNumGoalStates() const;
 
   /// Returns goal states.
-  const GoalStates& getGoalStates() const;
+  const GoalStates getGoalStates() const;
```
aditya-vk, line 50: Since we are returning a copy, I do not think we need to return `const GoalStates` but if others think the intention is clear from `const` we can let it be.

src/planner/ConfigurationToConfigurations.cpp
```diff
@@ -45,10 +47,16 @@ std::size_t ConfigurationToConfigurations::getNumGoalStates() const
 }
 
 //==============================================================================
-const ConfigurationToConfigurations::GoalStates&
+const ConfigurationToConfigurations::GoalStates
 ConfigurationToConfigurations::getGoalStates() const
 {
-  return mGoalStates;
+  std::vector<const statespace::StateSpace::State*> pointerStates;
+  for (size_t idx = 0; idx < mGoalStates.size(); idx++)
```
aditya-vk, line 54: Just to be consistent with what you have in the constructor:
```
for (const auto& goal : mGoalStates) {
    pointerStates.push_back(goal.getState());
}
```

src/planner/ConfigurationToConfigurations.cpp
```diff
@@ -45,10 +47,16 @@ std::size_t ConfigurationToConfigurations::getNumGoalStates() const
...
+  {
+    pointerStates.push_back(mGoalStates.at(idx).getState());
+  }
+
```
aditya-vk, line 58: Nit: remove empty line

include/aikido/planner/ConfigurationToConfigurations.hpp
sniyaz: Yeah, for some reason there are issues with the `ScopedState` class that make compilation fail if we do that (I think that's why all of the other classes have it as non-const as well). Fixing that looks like it's a big task (probably outside the scope of this PR).

Once we merge I'm happy to make an issue for us to go back and clean this up later! :smile:

PR #570, opened 2020-05-15
Fix OS X dependencies, remove ubuntu disco, update C++ standard

egordon:
Currently, all tests segfault after main on OS X. This was due to a missing dart dependency: `tinyxml`.

In addition, Ubuntu disco is no longer supported, so that has been updated to the latest LTS release (focal).

Finally, due to defaults in both bionic and focal, this updates the standard to C++14. This matches the DART standard, allows for now-idiomatic function (e.g. `std::make_unique`), and its where we should be heading for a stable release.

***


.ci/install_linux_catkin.sh
```diff
@@ -24,7 +24,7 @@ if [ "$BUILD_AIKIDOPY" = "ON" ]; then
     make -j4
     $SUDO make install
     cd ../..
-  elif [ $(lsb_release -sc) = "cosmic" ] || [ $(lsb_release -sc) = "disco" ]; then
+  elif [ $(lsb_release -sc) = "cosmic" ] || [ $(lsb_release -sc) = "disco" ] || [ $(lsb_release -sc) = "focal" ]; then
```
brianhou, line 27: Let's delete cosmic and disco from these.

.ci/install_linux_cmake.sh
```diff
@@ -48,7 +48,7 @@ if [ "$BUILD_AIKIDOPY" = "ON" ]; then
     make -j4
     $SUDO make install
     cd ../..
-  elif [ $(lsb_release -sc) = "cosmic" ] || [ $(lsb_release -sc) = "disco" ]; then
+  elif [ $(lsb_release -sc) = "cosmic" ] || [ $(lsb_release -sc) = "disco" ] || [ $(lsb_release -sc) = "focal" ]; then
```
brianhou, line 51: Same as above.

Brewfile
```diff
@@ -3,7 +3,7 @@ brew 'dartsim'
 brew 'eigen'
 brew 'libmicrohttpd'
 brew 'ompl'
-brew 'tinyxml2'
+brew 'tinyxml'
 brew 'yaml-cpp'
 
 # Install dartsim dependencies explicitly because, for some unknown reasons,
```
brianhou, line 9: Not being able to automatically pull in transitive dependencies is incredibly annoying. It really seems like there should be an obvious way to fix this. Can we add an issue to do so?

@jslee02 do you remember whether this problem was just with Travis CI + Homebrew? Or whether it's a plain Homebrew issue?

In the worst case, it seems like we should be able to do something like `brew deps dartsim` from Travis to generate this list of dependencies that should be installed.

egordon: Added issue #571 

Maybe we can unblock PRs by pushing this and continue discussion there?

brianhou: Yep, absolutely.

jslee02: > @jslee02 do you remember whether this problem was just with Travis CI + Homebrew? Or whether it's a plain Homebrew issue?

I didn't investigate further but just this workaround. Probably it was just Homebrew. Not sure if it's fixed or now. :disappointed:

PR #574, opened 2020-05-29
Fix bugs in CRRTs.

sniyaz:
This is a "spring cleaning" task I've wanted to get to for a while. Our existing `CRRT` and `CRRTConnect` implementations have a few bugs, and after diving in this PR should fix the major ones:

1) Use of hardcoded zero to check for a reach in `constrainedExtend`.

2) Not enforcing a resolution on the constraint manifold (just during interpolation, which can cause huge violations of the constraint after projection).

3) Force connection distance to be bounded by coll-check resolution in `CRRTConnect` (so connection is always collision-free).

***


src/planner/ompl/CRRT.cpp
```diff
@@ -335,12 +335,24 @@ CRRT::Motion* CRRT::constrainedExtend(
...
       // reached target or not making progress
       break;
     }
 
+    // NOTE: (sniyaz) This check makes sure the resolution of the path on the
+    // constraint manifold and the resolution used for interpolation (*before*
+    // projection) do not differ wildly.
+    // TODO: Should we expose the factor to scale `mMaxStepsize` by?
+    double manifoldResolution = si_->distance(xstate, cmotion->state);
+    if (manifoldResolution > 2.0 * mMaxStepsize)
```
aditya-vk, line 349: Even if it isn't exposed, it might be better to pull it out of here and make it `constexpr double k<Name>` in the header or in the beginning of this file.

src/planner/ompl/CRRT.cpp
```diff
@@ -335,12 +335,24 @@ CRRT::Motion* CRRT::constrainedExtend(
...
+    {
+      // The resolution on the manifold is violted, so treat the projection as
```
aditya-vk, line 351: typo: violated

src/planner/ompl/CRRTConnect.cpp
```diff
@@ -207,8 +207,12 @@ ::ompl::base::PlannerStatus CRRTConnect::solve(
     Motion* startMotion = startTree ? newmotion : lastmotion;
     Motion* goalMotion = startTree ? lastmotion : newmotion;
 
+    // NOTE: (sniyaz): Also check that the conenction motion is collision-free.
```
aditya-vk, line 210: typo: connection

sniyaz:
Awesome, thanks for the feedback all! Responded, and removing the WIP flag. Somehow it looks like I've broken a test on Travis, so I'm gonna look into that.

sniyaz:
@brianhou @aditya-vk Things should be all set now. Can y'all approve if things look OK?

src/planner/ompl/CRRT.cpp
sniyaz: Done!

src/planner/ompl/CRRT.cpp
sniyaz: Done!

src/planner/ompl/CRRTConnect.cpp
sniyaz: Done!

src/planner/ompl/CRRT.cpp
```diff
@@ -358,6 +362,18 @@ CRRT::Motion* CRRT::constrainedExtend(
...
 
+    // NOTE: (sniyaz) This check makes sure the resolution of the path on the
+    // constraint manifold and the resolution used for interpolation (*before*
+    // projection) do not differ wildly.
+    // TODO: Should we expose the factor to scale `mMaxStepsize` by?
+    double manifoldResolution = si_->distance(xstate, cmotion->state);
+    if (manifoldResolution > kLambda * mMaxStepsize)
+    {
+      // The resolution on the manifold is violated, so treat the projection as
+      // if it failed. This emulates the original CBiRRT paper correctly.
```
aditya-vk, line 373: 1. Why have a `kLambda` at all? Why not just compare with `mMaxStepsize`?
2. Resolution of the path and resolution for interpolation are confusing terms in this context. Replace with: "Ensure that the random state projected onto the constraint manifold is not more than `mMaxStepsize` distance away from the nearest neighbor on the tree."

aditya-vk (changes requested):
Good catch on the projection possibly being far away from the nearest neighbor. Left a few comments to address.

src/planner/ompl/CRRT.cpp
aditya-vk: I have not seen the original paper talk about this particular case actually. Could you point me to it? If not, best remove the reference to the paper and just explain the reason to add this block of code here.

src/planner/ompl/CRRTConnect.cpp
```diff
@@ -207,8 +207,12 @@ ::ompl::base::PlannerStatus CRRTConnect::solve(
     Motion* startMotion = startTree ? newmotion : lastmotion;
     Motion* goalMotion = startTree ? lastmotion : newmotion;
 
+    // NOTE: (sniyaz): Also ensure that the connection motion is collision-free.
+    bool validConnection
```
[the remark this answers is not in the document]
sniyaz: Yeah, my concern here is that `mConnectionRadius` is an independent parameter that the user gets to set. There are cases where the user could set  `mConnectionRadius` to a large value, which would possible allow collisions (exactly as you said).

IMO this is a better solution than forcing `mConnectionRadius` to be smaller than the collision-check resolution: that's a little cumbersome since the user has to be aware of the relationship between all of these parameters, rather than handling it gracefully here.

That said, I'm willing to change if we really want: just wanted to explain my thinking behind this 😄

src/planner/ompl/CRRT.cpp
```diff
@@ -335,7 +339,7 @@ CRRT::Motion* CRRT::constrainedExtend(
   while (ptc == false)
   {
 
-    if (distToTarget == 0 || distToTarget - prevDistToTarget >= -mMinStepsize)
+    if (si_->equalStates(cmotion->state, gstate) || distToTarget - prevDistToTarget >= -mMinStepsize)
```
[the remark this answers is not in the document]
sniyaz: Agree, might as well fix that while we're here! 👍

src/planner/ompl/CRRT.cpp
sniyaz: Nice catch on the original CBiRRT paper: I just re-read it and I don't think this is mentioned anywhere, so I must have been confused. I talked with Sidd a *long* time ago when I was writing my own CBiRRT code, and he mentioned that this is an important thing to do. I'll update the comment though, since it's wrong right now 😛 

The usage of `kLambda` is inspired by how OMPL's constrained statespaces handle this problem: their logic is that you want some "slack" between the resolution used during interpolation and the final resolution of the path on the constraint manifold. If you set `kLambda = 1`, any projection that increases the distance from the NN (even a little bit) is rejected, and that's probably too strict.

LMK your thoughts!

aditya-vk: Sure, a slack factor sounds good. Let's expose it, and change the name to denote what it implies. Maybe `mConstrainedExtensionMaxStepsize`?

src/planner/ompl/CRRTConnect.cpp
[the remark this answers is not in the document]
aditya-vk: The other issue with `mConnectionRadius` being too large is that the step sizes do not matter anymore at that point. I think the way to _correctly_ handle this is to force the user to choose a connection radius to be smaller than or equal to the collision checking resolution. From the docstring for connection radius, it does imply that a collision check between the two states in unnecessary since they are considered to be the same. 

If you think the user might not know the parameters correctly, how about adding a dosctring to mConnectionRadius that it needs to be <= collision resolution. If it is greater, issue an OMPL_WARN() and clamp it to the `min(mConnectionRadius, collision resolution)`.

src/planner/ompl/CRRT.cpp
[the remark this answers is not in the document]
sniyaz: Done!

src/planner/ompl/CRRTConnect.cpp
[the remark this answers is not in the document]
sniyaz: Agreed, that makes more sense. Done!

src/planner/ompl/CRRT.cpp
sniyaz: Done!

sniyaz:
@aditya-vk I think this is ready for another pass! I also renamed `mMaxStepsize ` -> `mProjectionResolution` since I realized the old name we had was super unclear.

include/aikido/planner/ompl/CRRT.hpp
```diff
@@ -207,7 +221,12 @@ class CRRT : public ::ompl::base::Planner
   constraint::ProjectablePtr mCons;
 
   /// The maximum length of a step before projecting
-  double mMaxStepsize;
+  double mProjectionResolution;
+
+  /// We multiply this with `mProjectionResolution` to get the maximum length of
+  /// a step *after* projecting: in other words, the resolution of the final
+  /// path on the constraint manifold.
+  double mFinalResolutionSlackFactor;
```
aditya-vk, line 229: nit: Consider `mProjectionResolutionSlackFactor` or `maxProjectedStepsizeSlackFactor`

src/planner/ompl/CRRT.cpp
```diff
@@ -358,6 +378,18 @@ CRRT::Motion* CRRT::constrainedExtend(
       }
     }
 
+    // NOTE: (sniyaz) This check makes sure the resolution of the path on the
```
aditya-vk, line 381: nit: can remove the name tag now

src/planner/ompl/CRRT.cpp
```diff
@@ -358,6 +378,18 @@ CRRT::Motion* CRRT::constrainedExtend(
...
+    // constraint manifold and the resolution used for interpolation (*before*
+    // projection) do not differ wildly. In other words, after projection
+    // the distance between the nearest-neighbor and the projected state should
+    // lie within a scalar multiple of `mProjectionResolution`.
+    double manifoldResolution = si_->distance(xstate, cmotion->state);
+    if (manifoldResolution
+        > mFinalResolutionSlackFactor * mProjectionResolution)
+    {
+      break;
```
aditya-vk, line 390: Should it `break` or should it `continue` to the next random sample?

src/planner/ompl/CRRTConnect.cpp
```diff
@@ -87,6 +87,18 @@ void CRRTConnect::clear()
 //==============================================================================
 void CRRTConnect::setConnectionRadius(double radius)
 {
+  if (radius > si_->getStateValidityCheckingResolution())
```
aditya-vk, line 90: Thanks for adding this!

sniyaz: Of course!

include/aikido/planner/ompl/CRRT.hpp
```diff
@@ -207,7 +221,12 @@ class CRRT : public ::ompl::base::Planner
   constraint::ProjectablePtr mCons;
 
   /// The maximum length of a step before projecting
-  double mMaxStepsize;
+  double mProjectionResolution;
```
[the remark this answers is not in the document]
sniyaz: Done!

include/aikido/planner/ompl/CRRT.hpp
sniyaz: Done!

src/planner/ompl/CRRT.cpp
sniyaz: Done!

src/planner/ompl/CRRT.cpp
sniyaz: I think the break is correct: we don't actually sample `gstate` in `constrainedExtend`, we just move towards this state that has already been sampled before `constrainedExtend` is called (see line `260` of `CRRT.cpp`).

I think this is the same reason we break on line `377`.

sniyaz:
@aditya-vk @brianhou Think the most recent nits are handled, can y'all take another pass?

src/planner/ompl/CRRT.cpp
aditya-vk: Ah yes, my bad. I mistook this to be a change in `solve()`.

aditya-vk (approved):
Looks good. Thanks for the changes!

PR #575, opened 2020-06-01
Update README to include DOI

aditya-vk:
Adds a DOI badge generated for the latest release.

***

- [NA] Document new methods and classes
- [NA] Format code with `make format`

- [NA] Set version target by selecting a milestone on the right side
- [NA] Summarize this change in `CHANGELOG.md`
- [NA] Add unit test(s) for this change


README.md
```diff
@@ -1,4 +1,5 @@
-# AIKIDO - AI for KIDO [![Build Status](https://travis-ci.org/personalrobotics/aikido.svg?branch=master)](https://travis-ci.org/personalrobotics/aikido) [![codecov](https://codecov.io/gh/personalrobotics/aikido/branch/master/graph/badge.svg)](https://codecov.io/gh/personalrobotics/aikido) [![Codacy Badge](https://api.codacy.com/project/badge/Grade/6d2da97f57904c96a01a5993cbbc4f51)](https://www.codacy.com/app/personalrobotics/aikido?utm_source=github.com&amp;utm_medium=referral&amp;utm_content=personalrobotics/aikido&amp;utm_campaign=Badge_Grade)
+# AIKIDO - AI for KIDO [![Build Status](https://travis-ci.org/personalrobotics/aikido.svg?branch=master)](https://travis-ci.org/personalrobotics/aikido) [![codecov](https://codecov.io/gh/personalrobotics/aikido/branch/master/graph/badge.svg)](https://codecov.io/gh/personalrobotics/aikido) [![Codacy Badge](https://api.codacy.com/project/badge/Grade/6d2da97f57904c96a01a5993cbbc4f51)](https://www.codacy.com/app/personalrobotics/aikido?utm_source=github.com&amp;utm_medium=referral&amp;utm_content=personalrobotics/aikido&amp;utm_campaign=Badge_Grade)[![DOI](https://zenodo.org/badge/34424077.svg)](https://zenodo.org/badge/latestdoi/34424077)
```
jslee02, line 1: This DOI is for a specific version (v0.3.1 in this case). Consider using https://doi.org/10.5281/zenodo.3871155 that represents all version and will always resolve to the latest one. Otherwise, you might need to update this link for every new release

brianhou: Bump -- does this still need to be fixed?

aditya-vk: I don't think so. The DOI already always points to the latest release.

aditya-vk: I don't think so. The DOI is updated automatically to resolve to the latest release (as this branch updated to now resolve to 0.4.0) -- does not have to be manual.

jslee02: Yeah, both links resolve to the latest release. Either link should be okay.

PR #586, opened 2020-09-26
Deprecate robot/util -> planToConfiguration() [WIP]

sniyaz:
This PR removes `planToConfiguration()` from `robot/util`, and updates the corresponding function in `ConcreteRobot` to use a snap planner.

We actually don't remove the method completely from `robot/util`, just the header: it's needed elsewhere in the `.cpp` file.

**TODO/WIP**:

1. Should we use the normal `ConfigurationToConfiguration` planner, or the DART version? Both work fine here and the normal version requires less cruft, so I've gone with that. LMK your thoughts!
2. Should we construct the planner in `ConcreteRobot -> planToConfiguration`, or have a planner member var in `ConcreteRobot`? The reason I didn't do this is because the `MetaSkeleton` passed to the planner constructor(s) may not match the `metaSkeleton` argument of `planToConfiguration()`.

***

### DIFF
--- a/include/aikido/robot/util.hpp
+++ b/include/aikido/robot/util.hpp
@@ -93,22 +93,6 @@ struct CRRTPlannerParameters
   double projectionTolerance;
 };
 
-/// Plan the robot to a specific configuration.
-/// Restores the robot to its initial configuration after planning.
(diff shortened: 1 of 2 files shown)


brianhou:
Should this be closed in favor of #588?

Issue #447, opened 2018-07-14
Locking the skeleton while planning


aditya-vk:
Previously the skeleton was locked in `planTo...` methods before calling any of the planners to plan. 

I think it should now reside in `<Planner><Problem> -> plan()` methods. @brianhou, can you confirm?

brianhou:
Hmm, I could be wrong but I was under the impression that @sniyaz tried to preserve the locking in the lowest level `planTo` functions (e.g. https://github.com/personalrobotics/aikido/blob/master/src/planner/vectorfield/VectorFieldPlanner.cpp#L151).

Regardless, I think this sounds reasonable. We might also want to consider adding an argument `bool lockSkeleton` to `Planner::plan` that controls whether the planner needs to lock the skeleton. This would be useful for wrapper/delegate planners: if the wrapper has already acquired the skeleton's lock in order to do something (e.g. sample configurations from a TSR), then the delegate doesn't need to lock the skeleton.

brianhou:
We should use DART's `MetaSkeleton::getLockableReference()` method rather than locking `Skeleton`s (available since DART 6.5).

sniyaz:
I believe that https://github.com/personalrobotics/aikido/pull/588 handles and documents this, at least for DART planners (I'm not sure what to do with non-DART planners, considering that those don't even know they're operating on a skeleton).

@brianhou @aditya-vk Does this seem like enough to close this issue, or are there other things we want to look at?

PR #588, opened 2020-10-05
Clean up robot/util.

sniyaz:
This PR cleans up `robot/util` by deleting all of the `planTo*` methods. This requires us to rewrite the corresponding methods in `ConcreteRobot` and `ConcreteManipulator` to reroute through the new Planner API.

This PR also *removes* all CRRT functionality in both `ConcreteRobot` and `ConcreteManipulator`. This will be added back later once we actually decide how to integrate CRRT and AIKIDO. None of our demo code should rely on CRRT, so this is fine.

**NOTE:** This PR depends on https://github.com/personalrobotics/aikido/pull/587 by @brianhou, due to the new planner adapter.

***


src/robot/ConcreteRobot.cpp
```diff
@@ -245,16 +264,34 @@ TrajectoryPtr ConcreteRobot::planToConfiguration(
...
+          stateSpace, std::make_shared<GeodesicInterpolator>(stateSpace));
+
+  // Convert to DART planner.
+  auto dartSnapConfigToConfigPlanner = std::make_shared<
+      ConfigurationToConfiguration_to_ConfigurationToConfiguration>(
+      snapConfigToConfigPlanner, metaSkeleton);
+
+  // Create planning problem.
+  auto castedGoalState
+      = static_cast<const statespace::dart::MetaSkeletonStateSpace::State*>(
```
brianhou, line 286: For some reason, I thought that we take the goal state as an `Eigen::VectorXd` instead. Is there a reason we take a `State`? What happens if this isn't actually a `MetaSkeletonStateSpace::State`?

src/robot/ConcreteRobot.cpp
```diff
@@ -245,16 +264,34 @@ TrajectoryPtr ConcreteRobot::planToConfiguration(
     const CollisionFreePtr& collisionFree,
     double timelimit)
 {
+  DART_UNUSED(timelimit);
+
+  auto robot = metaSkeleton->getBodyNode(0)->getSkeleton();
```
brianhou, line 269: This PR seems like a good time to address https://github.com/personalrobotics/aikido/issues/447.

brianhou (changes requested):
Found some time for a quick review! Broad strokes seem solid, although I'd also like to see a corresponding libada PR before merging.

src/robot/ConcreteRobot.cpp
sniyaz: Yeah, this is something I noticed as well. I'm not sure why we decided to take `StateSpace::State*` instead of `MetaSkeletonStateSpace::State*`. Since we're working with `ConcreteRobot` we must be working with `MetaSkeletonStateSpace::State*` (unless the user is doing something very stupid).

I agree that this design decision is a little funky (especially considering that we take a `metaSkeleton` as another argument), but changing it would break a lot of things just because of type issues. Maybe we can file an issue to clean this up later?

sniyaz:
@brianhou I tried to handle https://github.com/personalrobotics/aikido/issues/447 by enforcing that any **DART** planner (and only a DART planner) has to satisfy these two invariants when `plan()` is called:

1. The MetaSkeleton is locked during planning.
2. The MetaSkeleton is reset after planning.

The only DART planner class that didn't need to be updated was the VFP offset planner (since the core VFP code does both of these anyway). LMK what you think of this solution: if we're happy with it, I can document this new invariant for DART planners somewhere.

Also, for what it's worth I tested `simple_trajectories` with all of these changes and things seem to work fine (so at least `planToConfiguration` isn't broken).

src/robot/ConcreteRobot.cpp
brianhou: @aditya-vk does this seem ok for now?

brianhou (approved):
Seems generally fine to me, but would like @aditya-vk's eyes on this too.

aditya-vk:
Looks good to me. Has it been tested through `libherb` and `herb3_demos` to ensure everything works fine?

sniyaz:
@aditya-vk Yup! The only other thing we're waiting on are the corresponding ADA changes. I'll chat with @egordon about those.

aditya-vk (approved):
LGTM

aditya-vk (approved):
Thanks for the checks with ADA and HERB!

PR #589, opened 2020-10-05
Set the interactive marker viewer to autoupdate by default.

aditya-vk:
The user is currently required to do the following:
```
  aikido::rviz::InteractiveMarkerViewer viewer(topicName, frameName, world);
  viewer.setAutoUpdate(true);
```
The default expectation on introducing a viewer is to see the frames being updated in simulation. The user shouldn't need to explicitly set updates to be true.

***

- [NA] Document new methods and classes

- [NA] Summarize this change in `CHANGELOG.md`
- [NA] Add unit test(s) for this change


brianhou (approved):
Seems fine to me.

sniyaz (approved):
LGTM, FWIW. Only other thing I can maybe think of is documenting this default somewhere.

PR #593, opened 2020-10-30
Simplify Robot API

aditya-vk:
This PR simplifies the API to construct and plan with new robots. This changes the software design as well as moves a significant amount of code to setup robots from librobots to aikido in a general fashion. [Link to design document.](https://docs.google.com/document/d/1HAADie7oqY79yu6CV0-bSkeSIVaE2GdhC4A35AuaNlc/edit).

The code has been tested downstream with modified libherb/herb3_demos when registering subrobots with a factory pattern. The WIP is to replace with templates.

***


aditya-vk:
Assigning to @egordon as parts of this PR need to be modified (or even a complete rewrite might be warranted) to reflect the requirements from food manipulation project. @egordon please feel free to assign to someone else if needed? I might not have cycles to see this through. Thank you!

egordon:
Note: needed a DART version update for `cloneMetaskeleton`. This version should still be well behind for both `bionic` and `focal`.

Should be ready for a review concurrent with me adding this interface to `libada`.

Some issues that should be addressed in later PRs:
* Why the distinction between the DART planners and the non-DART planners? In general, I think the planning interface could be simplified.
* Ensure the TrajectoryExecutor joints actually match the managed degrees of freedom in the Robot's Metaskeleton. This fix can be implemented fairly easily after merging the Executor overhaul in https://github.com/personalrobotics/aikido/pull/602
* In the same vein, we should add the ability to register arbitrary executors. And RosRobot should gain the ability to manage controllers.

CMakeLists.txt
```diff
@@ -87,7 +87,7 @@ include(ClangFormat)
 
 find_package(Boost REQUIRED COMPONENTS filesystem)
 
-find_package(DART 6.6.2 REQUIRED
+find_package(DART 6.8.5 REQUIRED
```
amalnanavati, line 90: The README references DART 6.6.2, and should be updated

amalnanavati:
Does this need a corresponding libada PR? [Libada currently uses ConcreteRobot](https://github.com/personalrobotics/libada/blob/bd82b3487eae263f40bddc8de7926c7a9b5913c0/include/libada/Ada.hpp#L429), so as soon as this code gets merged in, libada will break.

include/aikido/robot/Robot.hpp
```diff
@@ -2,93 +2,345 @@
...
+  ///
+  /// Plan a specific body node of the robot to a
+  /// sample configuraiton within a Task Space Region.
+  ///
+  /// \param[in] bodyNodeName Bodynode (usually the end effector) whose frame
+  /// should end up in the TSR.
+  /// \param[in] tsr \see constraint::dart::TSR
+  /// \param[in] testableConstraint Planning (e.g. collision) constraints,
+  /// set to nullptr for no constraints (not recommended)
+  /// \param[in] maxSamples Maximum number of TSR samples
```
amalnanavati, line 182: Just flagging that this will be a merge conflict with https://github.com/personalrobotics/aikido/pull/607 . No worries -- whoever merges second can adapt to whoever merges first, but I just wanted to flag this for our records.

include/aikido/robot/Robot.hpp
```diff
@@ -2,93 +2,345 @@
...
+  }
+
+  // Sets the root robot.
+  void setRootRobot(RobotPtr root)
+  {
+    mRootRobot = root;
+  }
+
+  // Sets the Trajectory Executor
+  // TODO(egordon) Later: ensure trajectory executor joints match managed DoFs
```
amalnanavati, line 281: Just flagging this in case it got missed

amalnanavati:
> Does this need a corresponding libada PR? [Libada currently uses ConcreteRobot](https://github.com/personalrobotics/libada/blob/bd82b3487eae263f40bddc8de7926c7a9b5913c0/include/libada/Ada.hpp#L429), so as soon as this code gets merged in, libada will break.

EDIT: NVM, I see the corresponding libada issue: https://github.com/personalrobotics/libada/pull/64

src/robot/ros/CMakeLists.txt
```diff
@@ -0,0 +1,62 @@
+#==============================================================================
```
amalnanavati, line 1: I might have missed where this is called from, but why do we need a separate CMakeLists over here?

egordon: See control/ros for an example. This is here to define a new component robot::Ros, which holds all the Ros-dependent functionality. Right now, that's just URDF/srdfdom, but will contain controller switching in the future.

This should be called from robot CMakeLists as add_subdirectory("ros")

src/robot/Robot.cpp
```diff
@@ -1,21 +1,454 @@
...
+  if (!mTrajectoryExecutor)
+  {
+    throw std::runtime_error(
+        "TrajectoryExecutor is null, cannot execute trajectory.");
+  }
+
+  // Ensure all other executors that could contain these joints are cancelled
+  if (mRootRobot)
+  {
+    mRootRobot->cancelAllTrajectories(false, true);
```
amalnanavati, line 72: Why don't you just run this with true, true? It seems like you are replicating the logic since you're looping over all subrobots anyway

src/robot/Robot.cpp
```diff
@@ -1,21 +1,454 @@
...
+  auto selfCollisionFree = getSelfCollisionConstraint();
+
+  if (!collisionFree)
+    return selfCollisionFree;
+
+  // Make testable constraints for collision check
+  std::vector<aikido::constraint::ConstTestablePtr> constraints;
+  constraints.reserve(2);
+  constraints.emplace_back(selfCollisionFree);
+  if (collisionFree)
```
amalnanavati, line 183: Small nit here -- shouldn't this by default be true because you already return if it is not true? I'm not sure if this adds functionality or readability by being here

src/robot/Robot.cpp
```diff
@@ -1,21 +1,454 @@
...
+  {
+    if (collisionFree->getStateSpace() != mStateSpace)
```
amalnanavati, line 185: Another small nit: You allocate and reserve two spots in `constraints`, but then may not use it if this conditional is true. Perhaps move this conditional before the allocation of `constraints` just to avoid reserving and clearing memory unnecesarily?

src/robot/Robot.cpp
```diff
@@ -1,21 +1,454 @@
...
+        return nullptr;
+      }
+    }
+  }
+
+  // Create the subrobot.
+  auto subRobot = std::make_shared<Robot>(metaSkeleton, name);
+  const auto& root = mRootRobot ? mRootRobot : RobotPtr(this);
+  subRobot->setRootRobot(root);
+  subRobot->setWorld(nullptr); // inherit root robot world
```
amalnanavati, line 274: This comment was a bit confusing to be because it seems to imply that because the world ptr is nullptr, the subrobot will inherit the root robot's world. But I couldn't find logic for that. Looking back in the code, I saw on line 202 that the subrobot *would* inherit the root robot's world, but that is not because of the nullptr, but rather because the world is set. So if you can change this comment to something like "The subrobot will inherit the root robot's world, hence its world is unnecessary" that might make it clearer

src/robot/Robot.cpp
```diff
@@ -1,21 +1,454 @@
...
+
+  // Convert to DART Planner
+  using planner::dart::
+      ConfigurationToConfiguration_to_ConfigurationToConfiguration;
+  auto dartPlanner = std::make_shared<
+      ConfigurationToConfiguration_to_ConfigurationToConfiguration>(
+      basePlanner, mMetaSkeleton);
+
+  // Solve the problem with the DART planner.
+  auto rawPlan = dartPlanner->plan(problem, /* result */ nullptr);
```
amalnanavati, line 342: I've always been confused as to why result is commented out (even in earlier versions of AIKIDO). Will we ever have to revert to it? If not, perhaps we should delete it (in this line and others?)

src/robot/Robot.cpp
```diff
@@ -1,21 +1,454 @@
...
+  // Postprocess if enabled or provided
+  auto postprocessor
+      = (trajPostProcessor)
+            ? trajPostProcessor
+            : ((mEnablePostProcessing) ? mDefaultPostProcessor : nullptr);
+  if (rawPlan && postprocessor)
+  {
+    // Cast to interpolated or spline:
+    auto interpolated
+        = dynamic_cast<const aikido::trajectory::Interpolated*>(rawPlan.get());
```
amalnanavati, line 353: It seems like rawPlan.get() will either return Interpolated or Spline, and you're not sure which, so you try both? I'm wondering if there might be a cleaner way to do this (i.e., where you store what type rawPlan.get() will return)? No need to do it if it is too much work, I just wanted to mention that this current system of trying two casts to see which succeeds seems a bit strange imo

src/robot/Robot.cpp
```diff
@@ -1,21 +1,454 @@
...
+    // Else return raw path
+  }
+  return rawPlan;
 }
 
 } // namespace robot
 } // namespace aikido
+
+// TODO: Switch to PRIMITIVE once this is fixed in DART.
+// mCollisionDetector->setPrimitiveShapeType(FCLCollisionDetector::PRIMITIVE);
```
amalnanavati, line 454: ? This TODO and commented line are unclear to me. If it's important to keep, perhaps add some more context so that future generations of developers can pick up on this todo?

amalnanavati (approved):
Overall looks good. There are comments to address but none are major enough to prevent merging.

include/aikido/planner/World.hpp
```diff
@@ -21,7 +21,7 @@ class World
   // Encapsulates the state of the World.
   struct State
   {
-    std::unordered_map<std::string, dart::dynamics::Skeleton::Configuration>
+    std::unordered_map<std::string, ::dart::dynamics::Skeleton::Configuration>
```
amalnanavati, line 24: Just wondering, what is the conflicting namespace that requires pre-pended colons here? (As I understand it, this is chiefly useful if the class name exists in the global namespace and a sub-namespace that you're in. But I don't think either the aikido or planner namespaces have conflicting class names with this.)

amalnanavati:
Looked at the new commits, and they look good

include/aikido/planner/World.hpp
egordon: The namespace aikido::planner::dart exists (for planners that rely on DART body nodes, e.g. planToTSR). Honestly, I think it needs to be reorganized, but that's a discussion for a different PR.

I got a compile error when including both World and a planner::dart header file in Robot.hpp saying that `dynamics` isn't a namespace in `aikido:: planner:: dart`, so here we are.

CMakeLists.txt
egordon: Good catch, updated.

PR #596, opened 2021-03-12
Fixed logic in trajectory executor callback, ensure promise is only s…

egordon:
If trajectory aborts, mPromise exception is set twice. This raises an std::future exception that is ignored in RosTrajectoryExecutionException.cpp

Bug found while testing Python bindings. This logic ensures that mPromise exceptions are only raised once.

***


amalnanavati (approved):
Overall this looks good to me. I had a slight concern with it (see below), but I realized the concern doesn't matter in practice (also explained below), which is why I'm approving it.

Let's say the code gets into the first if statement (`terminalState != TerminalState::SUCCEEDED`) and not the second (`result->error_code != Result::SUCCESSFUL`). In the earlier version of the code, the exception would have included the terminal state, which might be useful information to have. Now, the exception will only include -1, which is a loss of information.

However, this doesn't matter in practice, for two reasons. First,  [RosTrajectoryExecutionException](https://github.com/personalrobotics/aikido/blob/master/src/control/ros/RosTrajectoryExecutionException.cpp) currently doesn't do anything with the second parameter. I can imagine a future where we change that, in which case this loss of `terminalState` information in the exception might matter. However, that brings me to the second point, which is that the `terminalState` is included in the message of the exception. So in practice, with or without this change, the user will know the terminal state for the actionlib. Therefore, I don't think the information loss affects anything in practice, and this PR can be merged.

egordon:
This is a good thing to flag for any future changes to RosTrajectoryExecutionException, so any future code reviews will catch this.

I think in the future, it may make sense for it to accept both the terminal state [which is always available] and the error code [available only if result exists, otherwise -1 suffices] as separate arguments (if they are both needed).

Issue #597, opened 2021-03-15
Migrate CI from Travis-CI.org to Github Actions


egordon:
Travis is going through some... [changes](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing)

It appears to be shutting down travis-ci.org (see yellow notice at [our repo page](https://travis-ci.org/github/personalrobotics/aikido)), migrating to travis-ci.com and reducing resources. The osx build now continuously errors via time-out. We can't merge any new PRs right now.

Options are:
(1) Request OSS minutes from Travis directly
(2) Migrate to Github Actions
(3) Pay for CI or switch to something else

Per discussion in Slack, we are likely going with option (2). Use this issue for further discussion.

PR #521, opened 2019-04-22
Combine InteractiveMarkerViewer classes.

aditya-vk:
Simplifies the code in InteractiveMarkerViewer. Refer to #520 for details (followed suggestion 2).

Resolves #519. Resolves #520.

***

**Before creating a pull request**

- [x] Document new methods and classes
- [x] Format code with `make format`

**Before merging a pull request**

- [x] Set version target by selecting a milestone on the right side
- [x] Summarize this change in `CHANGELOG.md`
- [ ] Add unit test(s) for this change

PR #598, opened 2021-04-02
Migrate Aikido to Github Actions

egordon:
Responding to Issue #597 .

This should run the exact same jobs as Travis but on Github Actions instead.


egordon:
I'm explicitly minimizing the change between these tests and the Travis tests.

However, the AIKIDOPY seems to fail to build because of the WorldInteractiveMarker issue from #521 , but only when building with Catkin... hmm

Should probably investigate that.

egordon:
Will create another PR based on the branch egordon/rviz-world after this is pushed.

egordon:
Issue with Bionic (unclear how we got around this in Travis). Tests are passing, but reporting as failure due to a segfault that happens during cleanup afterwards.

This is due to a mis-match between liboctomap-dev in bionic (version 1.8) and ros-melodic-octomap (version 1.9). I am not sure why catkin seems to get these two version confused.

That said, since both PRL and ICAROS are now updated to focal, we should switch these tests to be focal-centric.

Therefore, this PR now depends on https://github.com/personalrobotics/pr-cleanroom/pull/26

egordon:
Pushed and updated, so ready for review!

sniyaz:
So I've read through and (from what little I understand of GH actions) this seems sane.

One question though- I see we have one test failing currently. Do you know why? I think it's important to fix that before merging.

sniyaz:
> So I've read through and (from what little I understand of GH actions) this seems sane.
> 
> One question though- I see we have one test failing currently. Do you know why? I think it's important to fix that before merging.

Ah, ignore me- I just saw your other PR https://github.com/personalrobotics/aikido/pull/599.

In that case this seems fine- the only feedback is that it would nice to document somewhere how this all works (or even just link to some good documentation from our wiki). I always felt like I had to guess and check with Travis.

That's an issue we've always had, so I don't think it's fair to block this PR on that. Can we make an issue to add this documentation though?

sniyaz (approved):

egordon:
New wiki page created: https://github.com/personalrobotics/pr_docs/wiki/Continuous-Integration-and-Branch-Setup

PR #599, opened 2021-04-09
Aikidopy uses outdated WorldInteractiveMarker

egordon:
Aikidopy relies on the deprecated WorldInteractiveMarkerViewer which was removed in https://github.com/personalrobotics/aikido/pull/521 

This updates that interface.

Note: this depends on #598 , as this should fix the one failing (not required) test.

***


.ci/script_catkin.sh
```diff
@@ -20,7 +20,7 @@ fi
 # Build aikidopy and pytest
 if [ $BUILD_AIKIDOPY = ON ]; then
   ./scripts/internal-run.sh catkin build --no-status --no-deps -p 1 -i --cmake-args -DCMAKE_BUILD_TYPE=$BUILD_TYPE -DTREAT_WARNINGS_AS_ERRORS=OFF -DCODECOV=OFF --make-args aikidopy -- aikido
-  ./scripts/internal-run.sh make -C build/aikido pytest
+  $SUDO ./scripts/internal-run.sh make -C build/aikido pytest
```
sniyaz, line 23: What's the reason for this change? This may be dumb, seems unrelated to the interface change.

python/aikidopy/CMakeLists.txt
```diff
@@ -94,6 +88,12 @@ set_target_properties(aikidopy
     DEBUG_POSTFIX ""
 )
 
+execute_process(COMMAND ${PYTHON_EXECUTABLE} -c
```
sniyaz, line 91: Ditto: What's the reason for this change? This may be dumb, seems unrelated to the interface change.

sniyaz:
Heads up: I changed the base branch to your GH actions one so we can see a better diff.

This seems mostly fine, I just left two comments: I'm a bit confused since those changes don't seem to relate to `WorldInteractiveMarkerViewer -> InteractiveMarkerViewer`

python/aikidopy/CMakeLists.txt
egordon: Just a small readability change, since this variable is only used in the below lines.

.ci/script_catkin.sh
egordon: CI fails to run without this addition, since the installation step requires sudo.

egordon:
@sniyaz Responded for review comments and updated CHANGELOG. Ready for re-review

hejia-zhang (approved):

PR #602, opened 2021-08-12
Add base and pass-through Executor classes

egordon:
***


hejia-zhang (approved):
Looks good to me!

include/aikido/control/EffortExecutor.hpp
```diff
@@ -0,0 +1,52 @@
...
+  /// \note Future should return 0 on success or timeout.
+  ///
+  /// \param command Vector of joint commands, parallel with joints vector
+  /// \param timeout How long until command expires
+  virtual std::future<int> execute(
+      const std::vector<double> command,
+      const std::chrono::duration<double>& timeout)
+      = 0;
+
+  /// Documentation inherited.
```
brianhou, line 41: I think since this is `///`, Doxygen might interpret "Documentation inherited." as the new documentation. Usually this is with `//`, but it could be better with an explicit `\copydoc`.

include/aikido/control/Executor.hpp
```diff
@@ -0,0 +1,148 @@
...
+#include "aikido/common/pointers.hpp"
+
+namespace aikido {
+namespace control {
+
+AIKIDO_DECLARE_POINTERS(Executor)
+
+enum class ExecutorType
+{
+  kSTATE = -1,
```
brianhou, line 22: Since this is already an enum, I think adding the `k` prefix to all these names is redundant.

include/aikido/control/Executor.hpp
```diff
@@ -0,0 +1,148 @@
...
+
+// inline function to get joint names from skeleton
+inline std::vector<std::string> skeletonToJointNames(
+    dart::dynamics::SkeletonPtr skeleton)
+{
+  std::vector<std::string> ret;
+  if (!skeleton)
+    return ret;
+  ret.reserve(skeleton->getNumDofs());
+  for (auto dof : skeleton->getDofs())
```
brianhou, line 38:
```suggestion
  for (const auto& dof : skeleton->getDofs())
```

include/aikido/control/Executor.hpp
```diff
@@ -0,0 +1,148 @@
...
+  {
+    ret.push_back(dof->getName());
+  }
+  return ret;
+}
+
+// Parameter defaults
+constexpr std::chrono::milliseconds cDefaultThreadRate{10};
```
brianhou, line 46: I don't think we've previously used Hungarian notation for variable names, so we should be consistent.

include/aikido/control/Executor.hpp
```diff
@@ -0,0 +1,148 @@
...
+  {
+  }
+
+  /// Get primary ExecutorType
+  ExecutorType getType()
+  {
+    return mTypes[0];
+  }
+
+  // Get all ExecutorTypes
```
brianhou, line 83: Should these be `///`?

brianhou (changes requested):
I left some inline comments, but wanted to have a longer conversation about code duplication before reviewing in more depth.

It looks like there are three-ish versions of each executor for position/velocity/effort that share a majority of logic. Would it be possible to template these classes and implement the small bits of different logic in `*-impl.hpp` headers instead? This would make it easier for us to maintain, although I'm not sure whether it'd be harder to use from Python.

For example, I think that the vector of joint names that are taken in by the executors should be taken as `const std::vector<std::string>& joints` rather than `std::vector<std::string> joints`. Because the implementations are spread out across multiple files, this becomes really annoying to change. I think this current approach also makes us more susceptible to copy and paste bugs. (It also makes this PR much more difficult to review 😅 )

include/aikido/control.hpp
```diff
@@ -1,13 +1,19 @@
+#include "aikido/control/EffortExecutor.hpp"
```
brianhou, line 1: I think some of these may no longer exist?

egordon: Fixed, thanks!

egordon:
@brianhou this is probably the lowest-hanging, highest impact re-review.

include/aikido/control/detail/KinematicSimulationJointCommandExecutor-impl.hpp
```diff
@@ -0,0 +1,211 @@
...
+    mStartPosition.clear();
+    mTimeout = timeout;
+    switch (T)
+    {
+      case ExecutorType::VELOCITY:
+
+        for (size_t i = 0; i < mSkeleton->getDofs().size(); i++)
+        {
+          auto dof = mSkeleton->getDof(i);
+          dof->setVelocity(command[i]);
```
brianhou, line 90: Is `setVelocity` necessary if we're manually setting the position in `step`?

include/aikido/control/detail/KinematicSimulationJointCommandExecutor-impl.hpp
```diff
@@ -0,0 +1,211 @@
...
+
+        break;
+
+      case ExecutorType::POSITION:
+
+        for (size_t i = 0; i < mSkeleton->getDofs().size(); i++)
+        {
+          auto dof = mSkeleton->getDof(i);
+          mStartPosition.push_back(dof->getPosition());
+          if (mTimeout.count() > 1E-8)
```
brianhou, line 102: What is this for?

include/aikido/control/detail/KinematicSimulationJointCommandExecutor-impl.hpp
```diff
@@ -0,0 +1,211 @@
...
+    mTimeout = timeout;
+    switch (T)
+    {
+      case ExecutorType::VELOCITY:
+
+        for (size_t i = 0; i < mSkeleton->getDofs().size(); i++)
+        {
+          auto dof = mSkeleton->getDof(i);
+          dof->setVelocity(command[i]);
+          mStartPosition.push_back(dof->getPosition());
```
brianhou, line 91: This seems to happen for both types; consider pulling this out of the `switch`?

include/aikido/control/detail/KinematicSimulationJointCommandExecutor-impl.hpp
```diff
@@ -0,0 +1,211 @@
...
+  const auto executionTime
+      = std::chrono::duration<double>(timeSinceBeginning).count();
+
+  // executionTime may be negative if the thread calling \c step is queued
+  // before and dequeued after \c execute is called.
+  if (executionTime < 0)
+    return;
+
+  // Set joint states
+  double interpTime;
```
brianhou, line 143: Should this be inside the position case?

include/aikido/control/detail/KinematicSimulationJointCommandExecutor-impl.hpp
```diff
@@ -0,0 +1,211 @@
...
+      break;
+
+    default:
+      // SHOULD NEVER REACH
+      throw std::logic_error(
+          "Other KinematicSimulationExecutors not implemented.");
+  }
+
+  // Check if command has timed out
+  if (mTimeout.count() > 0.0 && executionTime >= mTimeout.count())
```
brianhou, line 182: What if `mTimeout` is 0?

include/aikido/control/detail/KinematicSimulationJointCommandExecutor-impl.hpp
```diff
@@ -0,0 +1,211 @@
...
+        auto dof = mSkeleton->getDof(i);
+        double pose
+            = (1.0 - interpTime) * mStartPosition[i] + interpTime * mCommand[i];
+        dof->setPosition(pose);
+      }
+
+      break;
+
+    default:
+      // SHOULD NEVER REACH
```
brianhou, line 176: I think this comment should either be removed or elaborated on.

include/aikido/control/util.hpp
```diff
@@ -0,0 +1,30 @@
...
+
+#include <dart/dynamics/DegreeOfFreedom.hpp>
+#include <dart/dynamics/Skeleton.hpp>
+
+namespace aikido {
+namespace control {
+
+/// Get joint names from skeleton for Executor intiialization
+static std::vector<std::string> skeletonToJointNames(
+    const dart::dynamics::SkeletonPtr skeleton)
```
brianhou: Should this be `ConstSkeletonPtr`?

include/aikido/control/util.hpp
```diff
@@ -0,0 +1,30 @@
+#ifndef AIKIDO_CONTROL_COMMON_HPP_
+#define AIKIDO_CONTROL_COMMON_HPP_
+
+#include <chrono>
```
brianhou, line 4: Is this used?

include/aikido/control/util.hpp
```diff
@@ -0,0 +1,30 @@
...
+#include <dart/dynamics/Skeleton.hpp>
+
+namespace aikido {
+namespace control {
+
+/// Get joint names from skeleton for Executor intiialization
+static std::vector<std::string> skeletonToJointNames(
+    const dart::dynamics::SkeletonPtr skeleton)
+{
+  std::vector<std::string> ret;
```
brianhou, line 16: Rename this to `names`?

include/aikido/control/util.hpp
```diff
@@ -0,0 +1,30 @@
...
+#include <chrono>
+
+#include <dart/dynamics/DegreeOfFreedom.hpp>
+#include <dart/dynamics/Skeleton.hpp>
+
+namespace aikido {
+namespace control {
+
+/// Get joint names from skeleton for Executor intiialization
+static std::vector<std::string> skeletonToJointNames(
```
brianhou: Maybe this could just be `getSkeletonDofNames`?

include/aikido/control/util.hpp
```diff
@@ -0,0 +1,30 @@
+#ifndef AIKIDO_CONTROL_COMMON_HPP_
+#define AIKIDO_CONTROL_COMMON_HPP_
```
brianhou, line 2: common -> util?

include/aikido/control/TrajectoryExecutor.hpp
```diff
@@ -14,9 +15,13 @@ namespace control {
 AIKIDO_DECLARE_POINTERS(TrajectoryExecutor)
 
 /// Abstract class for executing trajectories.
-class TrajectoryExecutor
+class TrajectoryExecutor : public Executor
 {
 public:
+  TrajectoryExecutor(std::vector<std::string> joints)
```
brianhou, line 21: Should this be `const` to match `Executor`?

include/aikido/control/ros/RosJointCommandExecutor.hpp
```diff
@@ -0,0 +1,67 @@
...
+#include <future>
+
+#include "aikido/control/JointCommandExecutor.hpp"
+#include "aikido/control/ros/RosJointGroupCommandClient.hpp"
+
+namespace aikido {
+namespace control {
+namespace ros {
+
+/// This Executor uses pr_control_msgs/JointGrouCommandAction to
```
brianhou, line 14:
```suggestion
/// This Executor uses pr_control_msgs/JointGroupCommandAction to
```

include/aikido/control/ros/RosJointCommandExecutor.hpp
```diff
@@ -0,0 +1,67 @@
...
+/// execute joint-wise velocity commands.
```
brianhou, line 15: Update this.

src/control/ros/RosJointGroupCommandClient.cpp
```diff
@@ -30,97 +30,128 @@ RosPositionCommandExecutor::RosPositionCommandExecutor(
...
+RosJointGroupCommandClient::~RosJointGroupCommandClient()
 {
   // Do nothing.
 }
 
 //==============================================================================
-std::future<void> RosPositionCommandExecutor::execute(
-    const Eigen::VectorXd& goalPositions)
+std::future<int> RosJointGroupCommandClient::execute(
+    ExecutorType type, const std::vector<double> goal, ::ros::Duration timeout)
```
brianhou, line 40: Pass `goal` by reference?

include/aikido/control/ros/RosJointCommandExecutor.hpp
```diff
@@ -0,0 +1,67 @@
...
+  /// Constructor.
+  /// \param[in] node ROS node handle for action client.
+  /// \param[in] controllerName Name of the controller to send command to.
+  /// \param[in] jointNames The names of the joints to set goal targets for
+  /// \param[in] connectionTimeout Timeout for server connection.
+  /// \param[in] connectionPollingPeriod Polling period for server connection.
+  RosJointCommandExecutor(
+      ::ros::NodeHandle node,
+      const std::string& controllerName,
+      std::vector<std::string> jointNames,
```
brianhou, line 30: Pass by const reference?

include/aikido/control/ros/detail/RosJointCommandExecutor-impl.hpp
```diff
@@ -0,0 +1,70 @@
...
+std::future<int> RosJointCommandExecutor<T>::execute(
+    const std::vector<double> command,
+    const std::chrono::duration<double>& timeout)
+{
+  ::ros::Duration duration;
+  duration.sec
+      = std::chrono::duration_cast<std::chrono::seconds>(timeout).count();
+  duration.nsec
+      = std::chrono::duration_cast<std::chrono::nanoseconds>(timeout).count()
+        % 1000000000UL;
```
brianhou, line 49: Nit: is there a way to implement this without having to count zeros? Maybe subtracting the whole seconds component from `timeout` before counting the nanoseconds remaining?

include/aikido/control/ros/RosTrajectoryExecutor.hpp
```diff
@@ -33,6 +33,7 @@ class RosTrajectoryExecutor : public aikido::control::TrajectoryExecutor
       const std::string& serverName,
       double waypointTimestep,
       double goalTimeTolerance,
+      std::vector<std::string> jointNames,
```
brianhou, line 36: Pass by const reference?

src/control/ros/RosTrajectoryExecutor.cpp
```diff
@@ -107,7 +109,9 @@ void RosTrajectoryExecutor::validate(const trajectory::Trajectory* traj)
     throw std::invalid_argument(
         "Trajectory is not in a MetaSkeletonStateSpace.");
 
-  // TODO: No check's happening here. Check if that's correct.
+  // TODO(egordon): check joints in traj against this->getJoints()
```
brianhou, line 112: Should this be implemented in this PR or a future one?

brianhou (changes requested):

include/aikido/control/Executor.hpp
```diff
@@ -0,0 +1,130 @@
...
+  {
+    if (mThread)
+    {
+      mThread->stop();
+      mThread.reset();
+    }
+  }
+
+private:
+  /// Call to spin first to pass current time to step
```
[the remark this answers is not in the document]
egordon: Actually, the reason this function is necessary is that, in order to enable real-time execution using an Executor's ExecutorThread, it needs a way to pass the current time into step(). This can't be done with std::bind (which would only pass the time point of construction) or a default argument (which also only passes the time point of construction). It also acts as a check to see if a joint has been requested of the thread, and if so, to not bother executing step().

include/aikido/control/Executor.hpp
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  /// Get primary ExecutorType
+  ExecutorType getType()
+  {
+    return mTypes[0];
+  }
+
+  /// Get all of this Executor's ExecutorTypes
+  std::vector<ExecutorType> getTypes()
```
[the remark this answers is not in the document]
egordon: Main use case: consider a `RosRobot` that handles controller switching. When you give it an executor, it will check the types and joint names to see which controller it needs to spin up.

include/aikido/control/Executor.hpp
```diff
@@ -0,0 +1,130 @@
+#ifndef AIKIDO_CONTROL_EXECUTOR_HPP_
+#define AIKIDO_CONTROL_EXECUTOR_HPP_
+
+#include <chrono>
+#include <future>
+#include <set>
```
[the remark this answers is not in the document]
egordon: Nope

include/aikido/control/JointCommandExecutor.hpp
```diff
@@ -0,0 +1,55 @@
...
+  /// \param[in] joints Vector of joint names this Executor acts upon
+  JointCommandExecutor(std::vector<std::string> joints) : Executor(T, joints)
+  {
+  }
+
+  virtual ~JointCommandExecutor()
+  {
+  }
+
+  /// Execute a Joint Command, setting future upon completion
```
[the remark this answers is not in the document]
egordon: Implementation-dependent. In general, completion could mean error or success. A velocity command could return an error (e.g. reach joint limits, other controller fault), but by default, it will run until `timeout` completes.

include/aikido/control/detail/KinematicSimulationJointCommandExecutor-impl.hpp
```diff
@@ -0,0 +1,211 @@
...
+extern template class KinematicSimulationJointCommandExecutor<
+    ExecutorType::POSITION>;
+
+extern template class KinematicSimulationJointCommandExecutor<
+    ExecutorType::VELOCITY>;
+
+//==============================================================================
+template <ExecutorType T>
+KinematicSimulationJointCommandExecutor<
+    T>::KinematicSimulationJointCommandExecutor(::dart::dynamics::SkeletonPtr
```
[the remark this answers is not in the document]
egordon: I don't think so, since this class will actually edit the Skeleton (i.e. by setting positions/velocities during trajectory execution).

include/aikido/control/detail/KinematicSimulationJointCommandExecutor-impl.hpp
```diff
@@ -0,0 +1,211 @@
...
+}
+
+//==============================================================================
+template <ExecutorType T>
+std::future<int> KinematicSimulationJointCommandExecutor<T>::execute(
+    const std::vector<double> command,
+    const std::chrono::duration<double>& timeout)
+{
+
+  {
```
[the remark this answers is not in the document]
egordon: It's not, this scope was grandfathered in from copied code, removing.

include/aikido/control/detail/KinematicSimulationJointCommandExecutor-impl.hpp
```diff
@@ -0,0 +1,211 @@
...
+    std::lock_guard<std::mutex> lock(mMutex);
+    DART_UNUSED(lock); // Suppress unused variable warning
+    auto promise = std::promise<int>();
+
+    if (command.size() != this->mJoints.size())
```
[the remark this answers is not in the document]
egordon: No, move lock guard to after check.

include/aikido/control/detail/KinematicSimulationJointCommandExecutor-impl.hpp
egordon: Not for the purpose of this class, but the client could be making some calculates based on the joint's velocity, so we want it to be accurately reflected here.

include/aikido/control/detail/KinematicSimulationJointCommandExecutor-impl.hpp
egordon: I don't like comparing doubles directly to 0 when dividing by them (since the result can get big fast), but I guess in this case it should be allowed.

include/aikido/control/detail/KinematicSimulationJointCommandExecutor-impl.hpp
egordon: Sure, just didn't do it due to compile error.

Fixed with scope addition: https://stackoverflow.com/questions/92396/why-cant-variables-be-declared-in-a-switch-statement

include/aikido/control/detail/KinematicSimulationJointCommandExecutor-impl.hpp
```diff
@@ -0,0 +1,211 @@
...
+    std::lock_guard<std::mutex> lock(mMutex);
+    DART_UNUSED(lock); // Suppress unused variable warning
+
+    if (mInProgress)
+    {
+      mInProgress = false;
+      mPromise->set_exception(
+          std::make_exception_ptr(std::runtime_error("Command canceled.")));
+    }
+    this->stop();
```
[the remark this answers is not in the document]
egordon: Yes, required for template class:
https://stackoverflow.com/questions/29390663/error-there-are-no-arguments-to-at-that-depend-on-a-template-parameter-so-a

include/aikido/control/detail/KinematicSimulationJointCommandExecutor-impl.hpp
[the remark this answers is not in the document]
egordon: See above for reason for `this->`

include/aikido/control/Executor.hpp
```diff
@@ -0,0 +1,130 @@
...
+
+#include "aikido/common/ExecutorThread.hpp"
+#include "aikido/common/pointers.hpp"
+
+namespace aikido {
+namespace control {
+
+AIKIDO_DECLARE_POINTERS(Executor)
+
+enum class ExecutorType
```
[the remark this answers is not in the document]
egordon: Done!

egordon:
The base Trajectory Execution has also been tested on libada v0.3.0 in JACO2 simulation.

sniyaz (dismissed):
So I took a high-level look and think I get the general design (I'm taking the fact that Brian's already punched a few holes as proof that it's generally sound).

There are a few high-level issues I've noticed right now- on the next pass I want to deep-dive into the implementation of each executor.

sniyaz:
Also one issue is that this PR is **really really big**. Is there a way we can split this PR into a few dependent commits? I feel like that way it would be easier to review and take an in-depth look at things

sniyaz:
Would splitting up the Trajectory and JointCommand executors here be possible?

egordon:
@sniyaz Unfortunately, at this point, it is too difficult to split up this commit by file. There are too many overlapping commits that touch multiple files (Trajectory and jointcommand executors). So we will need to get his merged as a large PR.

include/aikido/control/TrajectoryExecutor.hpp
```diff
@@ -34,12 +53,9 @@ class TrajectoryExecutor
   virtual std::future<void> execute(const trajectory::ConstTrajectoryPtr& traj)
       = 0;
 
-  /// Step to a point in time.
-  /// \note \c timepoint can be a time in the future to enable faster than
-  /// real-time execution.
-  ///
-  /// \param timepoint Time to simulate to
-  virtual void step(const std::chrono::system_clock::time_point& timepoint) = 0;
+  // Documentation inherited.
```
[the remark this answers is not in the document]
egordon: I thought I was running into linker issues without this, but I'm actually not sure about that.

With that said, even if this is redundant, I actually prefer to have it. Therefore, any dev only needs to look in 1 place to see which pure virtual functions they have to define, not 2.

include/aikido/control/Executor.hpp
```diff
@@ -0,0 +1,138 @@
...
+/// POSITION - commands position to dofs
+/// VELOCITY - commands velocity to dofs
+/// EFFORT   - commands effort (i.e. torque) to dofs
+/// TRAJECTORY - commands a trajectory to dofs
+/// MODE - commands a hardware command mode (e.g. position/velocity/effort)
+/// The following doesn't update the DoF directly:
+/// READONLY
+enum class ExecutorType
+{
+  STATE = -1,
```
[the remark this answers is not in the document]
egordon: I don't think there's a reason. This is a holdover from when this variable was meant to be something akin to an executor acting as a JointStateClient. I've renumbered these.

include/aikido/control/JointCommandExecutor.hpp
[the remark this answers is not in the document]
egordon: Timeout is technically correct in all the implementations so far.

For a position executor, it is a timeout in the traditional sense (i.e. if we don't get to the position in time, we abort).

For velocity/effort, it does function more like a duration. *However*, it is a maximum duration, since the command could be aborted or overridden by future execute commands. In the case of, say, visual servoing, it will actually function more like a timeout (as we will be sending updated commands regularly).

I think timeout makes the most sense given this.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 * Control
 
   * Fixed bug in RosTrajectoryExecutor [#596](https://github.com/personalrobotics/aikido/pull/596)
+  * Add generic and position/velocity/effort Executor types [#602](https://github.com/personalrobotics/aikido/pull/602)
```
brianhou, line 8:
```suggestion
  * Added generic and position/velocity/effort Executor types [#602](https://github.com/personalrobotics/aikido/pull/602)
```

include/aikido/control/Executor.hpp
```diff
@@ -0,0 +1,142 @@
...
+  STATE = 0,
+  POSITION = 1,
+  VELOCITY = 2,
+  EFFORT = 3,
+  TRAJECTORY = 4,
+  MODE = 5,
+  READONLY = 6
+};
+
+// Default rate for ExecutorThread to call step()
```
brianhou, line 46:
```suggestion
/// Default rate for ExecutorThread to call step()
```

include/aikido/control/Executor.hpp
```diff
@@ -0,0 +1,142 @@
...
+
+  /// Step to a point in time.
+  /// \note \c timepoint can be a time in the future to enable faster than
+  /// real-time execution.
+  ///
+  /// \param timepoint Time to simulate to
+  virtual void step(
+      const std::chrono::system_clock::time_point& /* timepoint */)
+  { /* Do Nothing */
+  }
```
brianhou, line 87:
```suggestion
  {
    // Do nothing
  }
```

include/aikido/control/JointCommandExecutor.hpp
```diff
@@ -0,0 +1,69 @@
...
+
+/// Abstract class for executing a command of a single type
+/// on a group of joints.
+/// \tparam T The primary and only type of this executor.
+template <ExecutorType T>
+class JointCommandExecutor : public Executor
+{
+public:
+  /// Constructor
+  /// Documentation inherited.
```
brianhou, line 25:
```suggestion
  /// \copydoc Executor::Executor
```

include/aikido/control/detail/KinematicSimulationJointCommandExecutor-impl.hpp
```diff
@@ -0,0 +1,209 @@
...
+  mExecutionStartTime = std::chrono::system_clock::now();
+  mTimeout = timeout;
+
+  // Set start positions
+  mStartPosition.clear();
+  for (std::size_t i = 0; i < this->getDofs().size(); ++i)
+  {
+    auto dof = this->getDofs()[i];
+    mStartPosition.push_back(dof->getPosition());
+  }
```
brianhou, line 86: Nit: it seems like we're calling `getDofs()` twich for each loop iteration, which could maybe be fixed by referencing `mDofs` directly or calling it once before the loop and referencing the local variable.

brianhou: Same issue below.

include/aikido/control/detail/KinematicSimulationJointCommandExecutor-impl.hpp
brianhou: Resolved offline: timeout of 0 means no timeout.

include/aikido/control/ros/RosJointGroupCommandClient.hpp
```diff
@@ -40,37 +37,44 @@ class RosPositionCommandExecutor
...
+  /// Send command to ROS server for execution.
+  /// \param[in] type Selecting between Position, Velocity, or Effort
+  /// \param[in] goal Vector of target positions for each joint
+  /// \param[in] timeout How long until command should expire
+  std::future<int> execute(
+      ExecutorType type,
+      const std::vector<double>& goal,
+      ::ros::Duration timeout);
 
-  /// \copydoc PositionCommandExecutor::step()
```
brianhou, line 49: Nit: should this be `\copydoc JointCommandExecutor::step()`?

include/aikido/control/ros/detail/RosJointCommandExecutor-impl.hpp
```diff
@@ -0,0 +1,85 @@
...
+extern template class RosJointCommandExecutor<ExecutorType::VELOCITY>;
+
+extern template class RosJointCommandExecutor<ExecutorType::EFFORT>;
+
+// Convert Dof List to Joint name List
+static std::vector<std::string> jointNamesFromDofs(
+    const std::vector<dart::dynamics::DegreeOfFreedom*>& dofs)
+{
+  std::vector<std::string> jointNames;
+  for (auto dof : dofs)
```
brianhou, line 17:
```suggestion
  for (const auto& dof : dofs)
```

src/control/Executor.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+#include <dart/dart.hpp>
+
+#include "aikido/common/ExecutorThread.hpp"
+#include "aikido/common/pointers.hpp"
+
+namespace aikido {
+namespace control {
+
+// Init static membres
```
brianhou, line 15:
```suggestion
// Init static members
```

brianhou (approved):
Looks good! I left some minor nits that would be great to address either now or in one of the (many) followup PRs.

src/control/InstantaneousTrajectoryExecutor.cpp
```diff
@@ -40,9 +44,12 @@ void InstantaneousTrajectoryExecutor::validate(
...
         "Trajectory is not in a MetaSkeletonStateSpace.");
 
-  // TODO: Delete this line once the skeleton is locked by isCompatible
-  std::lock_guard<std::mutex> lock(mSkeleton->getMutex());
-  space->checkIfContained(mSkeleton.get());
+  // Check that traj space is compatible with metaskeleton
+  if (!space->isCompatible(mMetaSkeleton.get()))
+  {
+    throw std::invalid_argument(
+        "Trajectory StateSpace incompatible with MetaSkeleton");
```
amalnanavati, line 51: @egordon FYI this doesn't work on gen2 real. Specifically, with `ada_feeding` on `emprise/gen3-integration`, `libada` on `egordon/executor_hotfix`, and `aikido` on `egordon/executor`, this exception gets called the first time the arm tries to execute (in moveAbovePlate).

src/control/ros/RosTrajectoryExecutor.cpp
```diff
@@ -104,10 +107,20 @@ void RosTrajectoryExecutor::validate(const trajectory::Trajectory* traj)
...
 
   if (!space)
+  {
     throw std::invalid_argument(
         "Trajectory is not in a MetaSkeletonStateSpace.");
+  }
+
+  dart::dynamics::MetaSkeletonPtr metaSkeleton
+      = dart::dynamics::Group::create("check", getDofs());
+  if (!space->isCompatible(metaSkeleton.get()))
```
egordon, line 117: @amalnanavati Thank you for checking on real! The exception I believe that gets triggered should be this one (rather than the one is Instantaneous Traj Executor). I'll double-check that.

amalnanavati:
Ignoring the CodeCov failure due to the challenges of spinning up a ROS master on CI for unit tests. 

Ethan verified that this works on the robot, so I'll merge it in.

PR #606, opened 2021-09-06
Add Docs deployment

egordon:
Our DOCS workflow hasn't actually been deployment to our environment since we migrated to Github Actions in #598

Adding this step should fix this. 

***

- [ N/A, internal ] Summarize this change in `CHANGELOG.md`

### DIFF
--- a/.ci/build_docs.sh
+++ b/.ci/build_docs.sh
@@ -4,9 +4,8 @@ set -ex
 
 AIKIDO_DIR="${HOME}/workspace/src/aikido"
 
-# Github Actions doesn't do a full clone of the repository.
-# We do it ourselves in a separate directory
-# To access all branches and tags.
+# GitHub Actions doesn't do a full clone of the repository, so
+# we clone it again in a separate directory to access all tags.
 mkdir -p ${AIKIDO_DIR}
 git clone "https://github.com/${GITHUB_REPOSITORY}.git" ${AIKIDO_DIR}
 
@@ -15,7 +14,6 @@ mkdir -p ${GITHUB_WORKSPACE}/gh-pages
 
 # Initialize list of API versions
 cat <<EOF > ${GITHUB_WORKSPACE}/gh-pages/README.md
-
 ## API Documentation
 
 EOF
--- a/.ci/build_docs.sh
+++ b/.ci/build_docs.sh
@@ -7,13 +7,6 @@ cd "${HOME}/workspace"
 
 AIKIDO_DIR="${HOME}/workspace/src/aikido"
 
-# For branch builds, Travis only clones that branch with a fixed depth of 50
-# commits. This means that the clone knows nothing about other Git branches or
-# tags. We fix this by deleting and re-cloning the full repository.
-# TODO: Unsure if we still need to do this for Github Actions
-rm -rf ${AIKIDO_DIR}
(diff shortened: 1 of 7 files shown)


.ci/build_docs.sh
```diff
@@ -2,23 +2,20 @@
...
+# To access all branches and tags.
+mkdir -p ${AIKIDO_DIR}
 git clone "https://github.com/${GITHUB_REPOSITORY}.git" ${AIKIDO_DIR}
 
 # Organize into "gh-pages" directory
 mkdir -p ${GITHUB_WORKSPACE}/gh-pages
 
 # Initialize list of API versions
 cat <<EOF > ${GITHUB_WORKSPACE}/gh-pages/README.md
+
```
brianhou, line 18: This produces an unnecessary leading blank line in the README.

.ci/build_docs.sh
```diff
@@ -2,23 +2,20 @@
...
 AIKIDO_DIR="${HOME}/workspace/src/aikido"
 
-# For branch builds, Travis only clones that branch with a fixed depth of 50
-# commits. This means that the clone knows nothing about other Git branches or
-# tags. We fix this by deleting and re-cloning the full repository.
-# TODO: Unsure if we still need to do this for Github Actions
-rm -rf ${AIKIDO_DIR}
+# Github Actions doesn't do a full clone of the repository.
+# We do it ourselves in a separate directory
+# To access all branches and tags.
```
brianhou, line 9:
```suggestion
# GitHub Actions doesn't do a full clone of the repository, so
# we clone it again in a separate directory to access all tags.
```

.github/workflows/docs.yml
```diff
@@ -43,6 +43,12 @@ jobs:
         run: .ci/install.sh
       - name: Script
         run: .ci/script.sh
+      - name: Deploy to GitHub Pages
+        uses: peaceiris/actions-gh-pages@v3
+        with:
```
brianhou, line 48: It looks like the Doxygen-generated HTML currently has timestamps that will change every time despite the content being the same. Consider using the `force_orphan` option to drop the old history. Alternatively, we can try to reconfigure Doxygen to avoid the timestamps, which might make the gh-pages commits actually meaningful. But there are also other things that might change underneath (e.g. Doxygen version), which is why we originally dropped the old versions.

CMakeLists.txt
```diff
@@ -180,11 +180,11 @@ if(${DOXYGEN_FOUND})
   # Optionally download Doxygen tagfiles for dependencies.
   set(DOXYGEN_TAGFILES "")
   if(DOWNLOAD_TAGFILES)
-    set(tagfile_url "http://dartsim.github.io/dart/v6.3.0/dart.tag")
+    set(tagfile_url "http://dartsim.github.io/dart/v6.6.2/dart.tag")
```
brianhou, line 183: I wonder if there's a way to auto-generate this version from the `find_package` invocation above, so there's one fewer checkbox to remember when updating DART versions.

brianhou (changes requested):

CMakeLists.txt
egordon: Yep, we an use ${DART_VERSION}

egordon:
Also manually ran deployment workflow:
https://github.com/personalrobotics/aikido/actions/runs/1211485519

brianhou (approved):
Looks great, thanks for sorting this out!

PR #607, opened 2021-09-10
Added batch options to ConcreteRobot's planToTSR

amalnanavati:
- Added batchSize and maxBatches parameters to planToTSR
- Renamed ConfigurationToTSR's maxSamples to maxSamplingTries to avoid confusion

ConcreteRobot's current planToTSR method samples 100 end-effector poses from the TSR, solves their IK to get a configuration, ranks then based on their distance from the current configuration, and plans in reverse-distance order until it finds a valid plan. However, if IK is slow (e.g., [DART's GradientDescentSolver](https://github.com/dartsim/dart/blob/main/dart/optimizer/GradientDescentSolver.cpp) on ADA's 6DoF arm), this procedure is also slow. This PR adds parameters for `batchSize` and `maxBatches`, where the planner samples `batchSize` end effector poses, ranks then, plans to them in reverse-distance order, and only if it doesn't find a valid plan does it continue to the next batch (up to `maxBatches`). batchSize=100 and maxBatches=1 is equivalent to the old system.

I tested the ADA sim demo with this code and with the current master branches of [libada](https://github.com/personalrobotics/libada) and [ada_feeding](https://github.com/personalrobotics/ada_feeding). I also tested it with modified versions of libada and ada_feeding that modify the batchSize and maxBatches parameters (upcoming PRs in those two repositories) and it worked for multiple configurations of batchSize and maxBatches.

***


amalnanavati:
Q for the reviewer: 

In the existing code, before `maxSamplingTries` is used, [there is an assertion to ensure it is positive](https://github.com/personalrobotics/aikido/blob/77e736ab0a364a9e90a29915033c5852721ff04d/src/constraint/dart/InverseKinematicsSampleable.cpp#L166). If I were to add an assertion for the two new parameters, I believe the best spot for it would be in the constructors of ConfigurationToTSR, which currently do nothing. Let me know if you think I should add assertions into those constructors (lines 29 and 52 of the new ConfigurationToTSR.cpp), or leave it as-is (i.e., trusting the user to pass in positive values for those parameters)

src/planner/dart/ConfigurationToConfiguration_to_ConfigurationToTSR.cpp
```diff
@@ -90,63 +90,69 @@ ConfigurationToConfiguration_to_ConfigurationToTSR::plan(
...
-    // has to terminate even if none are valid.
-    ++samples;
+    // Goal state
+    auto goalState = mMetaSkeletonStateSpace->createState();
 
-    if (!sampled)
-      continue;
+    // Sample valid configurations first.
+    std::size_t samples = 0;
+    while (samples < batchSize && generator->canSample())
```
egordon, line 121: The logic in this loop is a tad confusing. Some possible changes while we're working on it:
* Have an `if(!generator->canSample()) break;` at the start of the loop, which allows
* You to switch this to a for loop, which is less prone to future infinite loop bugs (esp with warnings), no need for the comment below then.

src/planner/dart/ConfigurationToConfiguration_to_ConfigurationToTSR.cpp
```diff
@@ -90,63 +90,69 @@ ConfigurationToConfiguration_to_ConfigurationToTSR::plan(
...
 
-    configurations.emplace_back(goalState.clone());
-  }
+      // Increment even if it's not a valid sample since this loop
+      // has to terminate even if none are valid.
+      ++samples;
 
-  if (configurations.empty())
-    return nullptr;
+      if (!sampled)
```
egordon, line 130: Not a fan of using `continue` or `break` unless necessary, as it can make following code more difficult.

In this case, wrapping `configurations.emplace_back()` in an `if(sampled)` can do away with the `continue` with minimal change.

include/aikido/planner/dart/ConfigurationToTSR.hpp
```diff
@@ -24,7 +24,11 @@ class ConfigurationToTSR : public Problem
   /// current state of when called.
   /// \param[in] endEffectorBodyNode BodyNode to be planned to move to a desired
   /// TSR.
-  /// \param[in] maxSamples Maximum number of TSR samples to plan to.
+  /// \param[in] maxSamplingTries Maximum number of times to try when
+  /// sampling from the TSR.
+  /// \param[in] batchSize Number of TSR samples to include per batch.
```
egordon, line 29: I would consider clarifying that this is a `ranked batch`. Batching itself doesn't add any additional functionality (since we already do `maxSamplingTries` underlying samples) except in that we carry out ranking within a batch.

include/aikido/planner/dart/ConfigurationToTSR.hpp
```diff
@@ -24,7 +24,11 @@ class ConfigurationToTSR : public Problem
   /// current state of when called.
   /// \param[in] endEffectorBodyNode BodyNode to be planned to move to a desired
   /// TSR.
-  /// \param[in] maxSamples Maximum number of TSR samples to plan to.
+  /// \param[in] maxSamplingTries Maximum number of times to try when
```
egordon, line 27: Maybe specify that this is the parameter to the underlying sample generator, similarly to:
https://github.com/personalrobotics/aikido/blob/92befdfe0af887aa23bf817b372c87b3933b789c/include/aikido/constraint/dart/InverseKinematicsSampleable.hpp#L32

include/aikido/planner/dart/ConfigurationToTSR.hpp
```diff
@@ -33,7 +37,9 @@ class ConfigurationToTSR : public Problem
       statespace::dart::ConstMetaSkeletonStateSpacePtr stateSpace,
       ::dart::dynamics::ConstMetaSkeletonPtr metaSkeleton,
       ::dart::dynamics::ConstBodyNodePtr endEffectorBodyNode,
-      std::size_t maxSamples,
+      std::size_t maxSamplingTries,
+      std::size_t batchSize,
+      std::size_t maxBatches,
```
egordon, line 42: In the name of not exploding arguments, thoughts about making this default to 1? Since it effectively just provides a convenient way of avoiding a wrapping for/while loop.

include/aikido/planner/dart/ConfigurationToTSR.hpp
```diff
@@ -53,7 +63,9 @@ class ConfigurationToTSR : public Problem
       statespace::dart::ConstMetaSkeletonStateSpacePtr stateSpace,
       const statespace::dart::MetaSkeletonStateSpace::State* startState,
       ::dart::dynamics::ConstBodyNodePtr endEffectorBodyNode,
-      std::size_t maxSamples,
+      std::size_t maxSamplingTries,
+      std::size_t batchSize,
+      std::size_t maxBatches,
```
egordon, line 68: See prev comment on default argument.

egordon (approved):
Minor comments, but nothing I would consider merge-blocking.

Feel free to address/comment and merge when ready.

amalnanavati:
The code compiles, but I have not been able to test that it works with the ADA feeding demo in simulation (blocked by [libada#64](https://github.com/personalrobotics/libada/pull/64) and a corresponding `ada_feeding` PR.) Converting this into a draft until it can be fully tested with `libada` and `ada_feeding`.

include/aikido/planner/dart/ConfigurationToTSR.hpp
egordon: Defaulting to 1 in Robot.hpp is fine.

egordon:
LGTM at this point, trusting the testing with both main libada / ada_feeding AND the new ada_feeding branch (which will also be merged today).

Issue #604, opened 2021-08-31
Fix XCode CI


egordon:
XCode CI never successfully migrated to Github Actions.

While this is low-priority (since all development is currently happening on Ubuntu), it should be fixed if possible.

PR #610, opened 2021-09-16
Update CI and Documentation to Restore Functionality from Travis

egordon:
Since the move to Github Actions in (#598), some CI functionality has been lost.

This includes:
* Codecov support
* Checking formatting prior to merging

In addition, the simultaneous update to Ubuntu Focal has lead to outdated documentation and unused CI code.

This PR re-adds this lost functionality, cleans up the CI code by removing old statements, and updates the main README to reflect the current use on Ubuntu Focal.

Unfortunately, this has also necessitated an upgrade to `clang-format-10`, which is available on Ubuntu Focal.

The previous version of `clang-format-6.0` had a bug such that the output of `output-replacement-xmls` (which we used to check formatting) did not match the actual replacements made in-file.

The new `clang-format-10` adds a `--dry-run` flag which offers the exact functionality we want (i.e. error if formatting is required). 

***


.ci/install_linux_catkin.sh
```diff
@@ -6,29 +6,14 @@ cd "${HOME}/workspace"
 cp -r "${GITHUB_WORKSPACE}" src
 ./scripts/internal-distro.py --workspace=src distribution.yml --repository "${REPOSITORY}" ${REQUIRED_ONLY}
 
-if [ $BUILD_NAME = TRUSTY_FULL_DEBUG ]; then
-  sudo apt-get install -y clang-format-6.0
-fi
+# Catkin on Focal-only
+$SUDO apt-get install -y clang-format-10
```
amalnanavati, line 10: Do we want to keep the if statements to support earlier versions of Ubuntu, or do we want to only support Focal? (Either way is fine, I just figured that documenting the preference and reasoning in this PR would be useful.)

.ci/install_linux_catkin.sh
```diff
@@ -6,29 +6,14 @@ cd "${HOME}/workspace"
...
 
 if [ "$BUILD_AIKIDOPY" = "ON" ]; then
   $SUDO apt-get -y install python3-dev python3-numpy
   $SUDO apt-get -y install python3-pip -y
   $SUDO pip3 install pytest -U
 
-  if [ $(lsb_release -sc) = "trusty" ] || [ $(lsb_release -sc) = "xenial" ] || [ $(lsb_release -sc) = "bionic" ]; then
```
amalnanavati, line 18: Same Q about whether we want to include if statements to maintain compatibility for earlier versions of Ubuntu or not

.ci/install_linux_cmake.sh
```diff
@@ -39,7 +36,7 @@ if [ "$BUILD_AIKIDOPY" = "ON" ]; then
   $SUDO apt-get -y install python3-pip -y
   $SUDO pip3 install pytest -U
 
-  if [ $(lsb_release -sc) = "trusty" ] || [ $(lsb_release -sc) = "xenial" ] || [ $(lsb_release -sc) = "bionic" ]; then
+  if [ $(lsb_release -sc) = "bionic" ]; then
```
amalnanavati, line 39: Clearly there is some support for earlier versions of Ubuntu. So maybe we should just document which versions of Ubuntu are supported? (Maybe you already did and I haven't seen that file yet)

.ci/install_linux_cmake.sh
```diff
@@ -60,7 +57,3 @@ fi
 if [ $BUILD_NAME = DOCS ]; then
   $SUDO apt-get -qq -y install doxygen
 fi
-
-if [ $BUILD_NAME = TRUSTY_FULL_DEBUG ]; then
-  sudo apt-get install -y clang-format-6.0
```
amalnanavati, line 65: Do we need to replace this with clang-format-10.0 (as you did in `install_linux_catkin`?)

README.md
```diff
@@ -31,6 +31,9 @@ $ sudo add-apt-repository ppa:personalrobotics/ppa
 $ sudo apt-get update
 $ sudo apt-get install libaikido-all-dev
 ```
+-->
+
+<!--- TODO: Test on OS X and re-add to Homebrew
```
amalnanavati, line 36: I just wanted to flag this TODO in case it got overlooked

README.md
```diff
@@ -133,6 +141,7 @@ Michael Koval ([**@mkoval**](https://github.com/mkoval))
 and Pras Velagapudi ([**@psigen**](https://github.com/psigen)).
 It has received major contributions from
 Shushman Choudhury ([**@Shushman**](https://github.com/Shushman)),
+Ethan Gordon ([**@egordon**](https://github.com/egordon)),
```
amalnanavati, line 144: 👏

cmake/ClangFormat.cmake
```diff
@@ -74,7 +74,7 @@ function(clang_format_add_targets)
     )
     add_custom_target(check-format
       COMMAND ${CMAKE_COMMAND} -E echo "Checking ${formatting_files_length} files..."
-      COMMAND ${CMAKE_SOURCE_DIR}/tools/check_format.sh ${CLANG_FORMAT_EXECUTABLE} ${formatting_files}
+      COMMAND ${CLANG_FORMAT_EXECUTABLE} -style=file --dry-run --Werror ${formatting_files} >/dev/null
```
amalnanavati, line 77: Hmm, wouldn't we want to see the output of the format executable (to identify/fix formatting issues) as opposed to redirecting it to /dev/null?

amalnanavati (approved):
LGTM. A few clarifying comments but nothing that should block merging.

cmake/ClangFormat.cmake
egordon: (1) any issues would be output to stderr, not stdout, so outputint stdout to /dev/null won't hide any important information from console.

(2) Having error information at this stage isn't useful, since the user isn't expected to fix these issues manually. This line tripping only means that the user neglected to run `make format` on their code.

README.md
egordon: Hi! :)

Felt comfortable with that at this point. Feel free to add yourself once you have a few hundred merged lines under your belt.

README.md
egordon: Thanks for the flag! It's captured as an issue #604 

But note that, since nobody is on OS X anymore, this is very, very low priority.

.ci/install_linux_cmake.sh
egordon: Nope, clang-format-10 is only needed for the FOCAL_CATKIN_DEBUG_CODECOV (we only need to test formatting in one job), so the `cmake` installation doesn't need it.

.ci/install_linux_catkin.sh
egordon: Let it be documented both here, in the fact that our workflows run on the ubuntu-20.04 Github-Actions machine, and the new README (which changed `Trusty` -> `Focal`), that we are indeed primarily supporting Focal (with deprecated support for Bionic until that goes end-of-life).

We can debate this in #aikido if needed, but I feel like we should generally try building to the current Ubuntu LTS and leave build support for the previous Ubuntu LTS until they go end-of-life.

.ci/install_linux_cmake.sh
egordon: Recall that these are Continuous-Integration workflows. They are only executed on the OS's that we tell them to execute on, and right now, we tell them to execute on bionic and focal. We could leave these statements in, but they will never be used again, since these versions are end-of-life (Github does maintain them in github-actions).

PR #605, opened 2021-09-02
Adds JacobianExecutor

egordon:
Wraps any VelocityExecutor to convert an end-effector SE(3) velocity command into joint velocity commands.

Prerequisite: #602 

***

**Before creating a pull request**

- [x] Document new methods and classes
- [x] Format code with `make format`

**Before merging a pull request**

- [x] Set version target by selecting a milestone on the right side
- [x] Summarize this change in `CHANGELOG.md`
- [x] Add unit test(s) for this change

PR #614, opened 2021-10-05
Add non-trajectory executors to Robot Class

egordon:
Previously, only a single TrajectoryExecutor was accessible from the Robot class.

This PR adds the functionality to have the Robot class manage multiple executors of arbitrary type (with convenience methods for the basic Trajectory, Position, Velocity, and Effort types).

***


RKJenamani (commented):
Not many comments as some of this code is shared with PR(#605)

RKJenamani (approved):

PR #616, opened 2021-12-03
[WIP] Create new AIKIDO IK solver class.

sniyaz:
**Background:** We want to start bringing new IK solvers (like `ikFast`) into our stack. Ideally each `Robot` object would have some associated IK object that we can swap out easily. The key issue right now is that the DART IK solver API is a little confusing and doesn't handle analytic solvers that well (it assumes that everything is a gradient-based solver).

**Proposed Solution:** Make an AIKIDO-specific IK solver class that we use in place of the DART one. This PR takes a first stab at this, and demonstrates how to make an optimization-based solver. This really just hides a lot of the complexity of the DART solver, and is a lot easier to work with.

 **Remaining Issues:** This is a draft PR, and there's a lot to sort out. In no particular order:

1. What should the RV of IK be? Right now I'm just using `Eigen::VectorXd` because I don't think we should use allocated states, but this kind of clashes with the rest of AIKIDO.

2. Where should this be? Right now it's just alongside the other Robot stuff.

3. Naming conventions for sub-classes. Also, calling the base class `InverseKinematics` is a bit confusing since that's what DART does as well.

***


src/robot/GradientDescentIk.cpp
```diff
@@ -0,0 +1,61 @@
...
+      = mSeedConfigSampleable->createSampleGenerator();
+  auto seedState = mArmSpace->createState();
+
+  for (size_t i = 0; i < maxSolutions; i++)
+  {
+    // TODO: What to do if we can't sample a seed config?
+    if (!ikSeedGenerator->sample(seedState))
+      continue;
+
+    mArmSpace->setState(mArm.get(), seedState);
```
amalnanavati, line 49: Just to verify that I'm understanding this right, the only functional difference between this GradientDescentSolver and DART's GradientDescentSolver is that this one samples a random state as the initial guess, right?

amalnanavati: In the code [here](https://github.com/dartsim/dart/blob/7e3db63a07f7fcabd577efbd0487c6cd99e31c34/dart/dynamics/InverseKinematics.cpp#L103) DART set's the problem's initial guess to the current state, and then in the code [here](https://github.com/dartsim/dart/blob/7e3db63a07f7fcabd577efbd0487c6cd99e31c34/dart/optimizer/GradientDescentSolver.cpp#L132) it sets that value to the initial guess

amalnanavati: In our current stack, [randomizing the state is done in AIKIDO's InverseKinematicsSampleable](https://github.com/personalrobotics/aikido/blob/c0a8b7eff286640374631a2c6bddb571b829a73e/src/constraint/dart/InverseKinematicsSampleable.cpp#L209), which calls DART's IK class. So we would have to change InverseKinematicsSampleable accordingly.

(Regardless, if we intend AIKIDO's IK class to become our main IK class as opposed to the DART one, we should change InverseKinematicsSampleable to work with AIKIDO's IK class by default. InverseKinematicsSampleable is effectively the class that takes a pose-solving IK and generalizes it to a TSR-solving IK.)

src/robot/GradientDescentIk.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+  std::vector<Eigen::VectorXd> solutions;
+
+  std::shared_ptr<SampleGenerator> ikSeedGenerator
+      = mSeedConfigSampleable->createSampleGenerator();
+  auto seedState = mArmSpace->createState();
+
+  for (size_t i = 0; i < maxSolutions; i++)
+  {
+    // TODO: What to do if we can't sample a seed config?
```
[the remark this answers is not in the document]
brianhou: I think `SampleGenerator`s have a `canSample` method, and I think it's fine to continue as long as that's true.

src/robot/GradientDescentIk.cpp
```diff
@@ -0,0 +1,61 @@
...
+    if (!ikSeedGenerator->sample(seedState))
+      continue;
+
+    mArmSpace->setState(mArm.get(), seedState);
+    mDartIkSolver->getTarget()->setTransform(targetPose);
+    bool withinTol = mDartIkSolver->solveAndApply(true);
```
amalnanavati, line 51: I believe we should call `solve` instead of `solveAndApply` (both functions can be found [in this file](https://github.com/dartsim/dart/blob/7e3db63a07f7fcabd577efbd0487c6cd99e31c34/dart/dynamics/InverseKinematics.cpp#L133)). Because given that we are calling this in the loop, it seems wasteful to apply the solution each time. 

With that said, since we are returning more than one solutions, there is a question of which solution should get applied. I think either the first one or the last one makes sense, but regardless I think we should apply outside the loop. (And whichever solution we decide to apply, we should mention that in the documentation)

amalnanavati (commented):
Overall looks good, although there are lingering discussions w.r.t. `InverseKinematicsSampleable` and `solveAndApply`

amalnanavati:
In terms of the questions, here are my two cents:

1. I think Eigen::VectorXd makes sense, which is the same effective [retval of DART's IK](https://github.com/dartsim/dart/blob/7e3db63a07f7fcabd577efbd0487c6cd99e31c34/dart/dynamics/InverseKinematics.cpp#L125). Further, we are returning a vector of Eigen::VectorXd, and we don't know which solution the user would want to use. Hence, it doesn't make sense imo to preemptively allocate states for them – we should leave that to the user. 
2. My initial leaning is to have a separate IK folder within the `robot` folder, but I see that you [have a commit where you specifically undid that](https://github.com/personalrobotics/aikido/pull/616/commits/a43beb9157eb3995d9b9e84a210d6ca4c50d5f3c). What made you do that? Understanding the factors you considered there could help us decide where to put this code.
3. Do we need a naming convention for subclasses? I was thinking stuff like `GradientDescentIK`, `IKFast`, and whatever the other _default names_ for IK solvers would work well.
4. I agree that having the same class name within AIKIDO and DART is confusing. Speaking as someone who was just familiarizing myself with AIKIDO a few months ago, I found that it took some time to determine which classes were in AIKIDO and which were in DART. With that said, it's not immediately obvious to me what a better name would be. `RobotInverseKinematics`? But afaik IK is only for robots... Perhaps `InverseKinematicsSolver`?

src/robot/GradientDescentIk.cpp
amalnanavati: Also, another question is should `InverseKinematicsSampleable` be in the same folder/namespace as the rest of the IK?

sniyaz: Yeah! This is really just wrapping DART's IK solver in the new interface and randomizing the initial state each time (again since DART just goes from the current position of the skeleton).

And yeah, we will need to rewrite `InverseKinematicsSampleable`- my goal is to merge this new API in this PR, and then start actually using it in subsequent ones. As for where to put that class I think it belongs where it is with the other sampleable stuff- IK is more of a robot thing IMHO

src/robot/GradientDescentIk.cpp
sniyaz: Agreed, changed over to `solve()`.

Also which solution to apply doesn't actually matter- we're using a state saver here anyway at the beginning, which restores the metaskeleton positions when it gets destructed.

sniyaz:
@amalnanavati Responded to your nits, also some thoughts:

1. Agreed!
2. It made things feel a little cluttered if that makes sense- there aren't that many classes right now. But maybe when we add more our feelings on this will change?
3. Yeah, fair. Perhaps naming convention is over-rated.
4. Yeah I agree- maybe avoiding some degree of confusion is impossible here. I still think on balance this will be better since 1) we can give robots an actual IK method now and 2) we can at least switch between IK types in a consistent manner.

sniyaz:
Whoops, this shouldn't be WIP anymore huh?

sniyaz:
Also I remember that @brianhou said he'd like an example of what an IkFast solver would look like with this API, so I included a first cut of that as well!

amalnanavati (approved):
Looks good to me! @sniyaz, I'm assuming you have written some sort of script to test that the IKFast and GradientDescentIK actually work using this code, right? If so, this PR should be good-to-go.

(Eventually I do think we should add unit tests, at least for GradientDescentIK which doesn't require a binary. See the [unit tests for InverseKinematicsSampleable](https://github.com/personalrobotics/aikido/blob/master/tests/constraint/test_InverseKinematicsSampleable.cpp). I know your goal in this PR is just to approve the interface, so I think this PR is good to go, but I feel like the next one should add unit tests.)

amalnanavati:
Or well looks like CI failed, so should be good to merge after that. I'm happy to re-review then.

PR #619, opened 2022-01-22
Fix CI Issue Due to Deprecated DART Function

amalnanavati:
A CI failure recently started cropping up because DART deprecated [parseSkeleton](https://github.com/dartsim/dart/blame/main/dart/utils/urdf/DartLoader.hpp#L160) in 6.12 (see this [DART PR](https://github.com/dartsim/dart/pull/1605)). This PR addresses that.

I've verified that it still works on DART 6.11. The CI will check whether it works on DART 6.12

***


include/aikido/common/detail/parseSkeleton-impl.hpp
```diff
@@ -0,0 +1,33 @@
+#ifndef AIKIDO_COMMON_DETAIL_PARSESKELETON_IMPL_HPP_
```
egordon-HRI, line 1: Please move this to src/common/parseSkeleton.cpp.

Header impl's are supposed to be for template classes (which cannot be pre-compiled into the library). Functions like this don't have that issue, and will unnecessarily slow down the compilation time of all sub-libraries that use it.

include/aikido/common/parseSkeleton.hpp
```diff
@@ -0,0 +1,18 @@
...
+namespace common {
+
+const dart::dynamics::SkeletonPtr parseSkeleton(
+    const dart::common::ResourceRetrieverPtr& retriever,
+    const dart::common::Uri& uri);
+
+} // namespace common
+} // namespace aikido
+
+#include "aikido/common/detail/parseSkeleton-impl.hpp"
```
egordon-HRI, line 16: See above, flagging this line for removal.

egordon-HRI (changes requested):
(1) Do we want to change these calls to be to the already-existing call to aikido::io::loadSkeletonFromURDF()?
See: https://github.com/personalrobotics/aikido/blob/c015cea62017b6253264460c70b28df9119c69c0/src/io/util.cpp#L11

I would recommend making this update there for a smaller, easier PR.

(2) If we do decide to go with this change, see comment about moving it to a cpp compilation unit instead of a header impl.

include/aikido/common/detail/parseSkeleton-impl.hpp
amalnanavati: Thanks! Yeah I wasn't sure what requires putting some code in an impl and others not -- this helps.

egordon (changes requested):
Left a review from the wrong account (which will be deleted in April). Feel free to dismiss and address for this account.

amalnanavati:
Great insight @egordon ! I addressed the changes and it's ready to re-review. 

@hsnemlekar when you have a moment can you verify that the `addBodyFromURDF` in the Python bindings still works as expected? The transform was earlier set with `(skeleton->getJoint(0))->setTransform(transform)` and is now set with `(skeleton->getRootJoint())->setTransform(transform)`. My hunch is that those functions do the same thing, but that wasn't immediately obvious by looking at the DART code. So if you have code that calls that function, could you ensure that it still works as expected?

egordon (approved):
If everything works (and you don't get const warnings, LGTM. Thanks!

amalnanavati:
Heramb verified that the changes have not affected the Python `addBodyFromURDF`. Merging.

PR #621, opened 2022-05-11
Controller Switching through RosRobot

RKJenamani:
Builds on top of #614. Adds corresponding controller operations (load, start, stop) to the executors operations (register, activate, deactivate). Corresponding hardware mode switching takes place if a joint mode controller is specified.

***


src/robot/ros/RosRobot.cpp
```diff
@@ -38,6 +52,262 @@ RosRobot::RosRobot(
...
+}
+
+//=============================================================================
+bool RosRobot::switchControllers(
+    const std::vector<std::string> startControllersNames, const std::vector<std::string> stopControllersNames)
+{
+  if (!mRosSwitchControllerServiceClient)
+  {
+    throw std::runtime_error(
+        "ROS controller manager service client not instantiated.");
```
egordon, line 264: Minimize the use of runtime errors when not necessary. Print a warning and return false.

src/robot/ros/RosRobot.cpp
```diff
@@ -38,6 +52,262 @@ RosRobot::RosRobot(
...
+  return mRosSwitchControllerServiceClient->call(srv) && srv.response.ok;
+}
+
+//=============================================================================
+bool RosRobot::switchControllerMode(
+    const hardware_interface::JointCommandModes jointMode)
+{
+  if (!mRosJointModeCommandClient)
+  {
+    throw std::runtime_error(
```
egordon, line 290: Same comment as elsewhere. No need for a runtime error. Print warning and return False

src/robot/ros/RosRobot.cpp
```diff
@@ -38,6 +52,262 @@ RosRobot::RosRobot(
...
+        "ROS joint mode controller actionlib client not instantiated.");
+  }
+
+  // Currently we only send the same control mode to each joint
```
egordon, line 294: Add TODO comment here and create an issue so we can address this later (ideally, we need to be able to set the joint modes for only the joints used by the Executor)

src/robot/ros/RosRobot.cpp
```diff
@@ -38,6 +52,262 @@ RosRobot::RosRobot(
...
+  mRosJointModeCommandClient = rosJointModeCommandClient;
+}
+
+//=============================================================================
+bool RosRobot::loadController(const std::string loadControllerName)
+{
+  if (!mRosLoadControllerServiceClient)
+  {
+    throw std::runtime_error(
+        "ROS controller manager service client not instantiated.");
```
egordon, line 235: See elsewhere. Warning, return False

include/aikido/robot/ros/RosRobot.hpp
```diff
@@ -18,21 +21,198 @@ AIKIDO_DECLARE_POINTERS(RosRobot)
...
       const bool addDefaultExecutors,
       const dart::common::ResourceRetrieverPtr& retriever
-      = std::make_shared<aikido::io::CatkinResourceRetriever>());
+      = std::make_shared<aikido::io::CatkinResourceRetriever>(),
+      std::shared_ptr<::ros::NodeHandle> node = nullptr);
+
+  ///
+  /// Construct a new RosRobot as subrobot.
+  ///
+  /// \param[in] refSkeleton The metaskeleton defining the robot.
```
egordon, line 45: Consider `/// Documentation Inherited` after first descriptor line

include/aikido/robot/ros/RosRobot.hpp
```diff
@@ -18,21 +21,198 @@ AIKIDO_DECLARE_POINTERS(RosRobot)
...
+      dart::dynamics::ReferentialSkeletonPtr refSkeleton,
+      const std::string& name) override;
+
+  ///
+  /// Stops the controller corresponding to this executor and deactivates
+  /// executor.
+  void deactivateExecutor() override;
+
+  ///
+  /// Loads the controller. Throws a runtime_error if controller cannot be
```
egordon, line 86: Just return empty string on failure

include/aikido/robot/ros/RosRobot.hpp
```diff
@@ -18,21 +21,198 @@ AIKIDO_DECLARE_POINTERS(RosRobot)
...
+  /// loaded. Add an (executor, controller, control mode) to the inactive
+  /// (executor, controller, control mode) list Releases DoFs held by executor
```
egordon, line 88: Add period after end of sentence.

include/aikido/robot/ros/RosRobot.hpp
```diff
@@ -18,21 +21,198 @@ AIKIDO_DECLARE_POINTERS(RosRobot)
...
+  /// if held.
+  ///
+  /// \param[in] executor The Executor to add to the inactive list.
+  /// \param[in] controllerName The corresponding controller to load and add to
+  /// the inactive list. \param[in] controllerMode The corresponding control
```
egordon, line 93: \param[in] and \return need to be on new lines. Note `make format` will often screw this up the first time.

include/aikido/robot/ros/RosRobot.hpp
```diff
@@ -18,21 +21,198 @@ AIKIDO_DECLARE_POINTERS(RosRobot)
...
+
+  ///
+  /// Loads the controller. Throws a runtime_error if controller cannot be
+  /// loaded. Add an (executor, controller) to the inactive (executor,
+  /// controller, control mode) list Operations on this executor do not affect
+  /// the control mode. Releases DoFs held by executor if held.
+  ///
+  /// \param[in] executor The Executor to add to the inactive list.
+  /// \param[in] controllerName The corresponding controller to load and add to
+  /// the inactive list. \return A robot-unique non-negative ID (negative
```
egordon, line 110: \return on new line

include/aikido/robot/ros/RosRobot.hpp
```diff
@@ -18,21 +21,198 @@ AIKIDO_DECLARE_POINTERS(RosRobot)
...
+  /// mode to add to the inactive list. \return A robot-unique non-negative ID
+  /// (negative implies failure)
+  std::string registerExecutor(
+      aikido::control::ExecutorPtr executor,
+      std::string controllerName,
+      hardware_interface::JointCommandModes controllerMode,
+      std::string desiredName);
+
+  ///
+  /// Loads the controller. Throws a runtime_error if controller cannot be
```
egordon, line 103: Flagging change of functionality (empty string on failure)

include/aikido/robot/ros/RosRobot.hpp
egordon: Also: change to `empty implies failure` as per the changes to Robot

include/aikido/robot/ros/RosRobot.hpp
```diff
@@ -18,21 +21,198 @@ AIKIDO_DECLARE_POINTERS(RosRobot)
...
+      std::string controllerName,
+      std::string desiredName);
+
+  ///
+  /// Add an executor to the inactive executors list.
+  /// Operations on this executor do not affect the controller or control mode.
+  /// Releases DoFs held by executor if held.
+  ///
+  /// \param[in] executor The Executor to add to the inactive list.
+  /// \return A robot-unique non-negative ID (negative implies failure)
```
egordon, line 123: Flagging "negative implies failure"

include/aikido/robot/ros/RosRobot.hpp
```diff
@@ -18,21 +21,198 @@ AIKIDO_DECLARE_POINTERS(RosRobot)
...
+  /// Deactivates the current active executor.
+  /// Starts the corresponding controller and switches to corresponding control
+  /// mode. Sets an executor from the inactive executor list to be the active
+  /// executor. Holds and releases DoFs as needed.
+  ///
+  /// \param[in] id of executor on executor list
+  /// \return True if successful. If false, all executors are inactive.
+  bool activateExecutor(std::string id) override; 
+
+  using Robot::activateExecutor; // This un-hides Robot::activateExecutor(const aikido::control::ExecutorType type)
```
egordon, line 136: Move comment to preceding line.

include/aikido/robot/ros/RosRobot.hpp
```diff
@@ -18,21 +21,198 @@ AIKIDO_DECLARE_POINTERS(RosRobot)
...
+  ///
+  /// \param[in] rosControllerServiceClient ros controller service client.
+  void setRosLoadControllerServiceClient(
+      const std::shared_ptr<::ros::ServiceClient>& rosLoadControllerServiceClient); 
+
+  ///
+  /// Sets the ros controller service client which enables controller switching.
+  ///
+  /// \param[in] rosControllerServiceClient ros controller service client.
+  void setRosSwitchControllerServiceClient(
```
egordon, line 149: I would not have an external class own the service clients. Have a single function `setRosControllerService(string cm_name)` that takes in the controller manager namespace. Then this class can construct its own service clients based on that namespace.

include/aikido/robot/ros/RosRobot.hpp
```diff
@@ -18,21 +21,198 @@ AIKIDO_DECLARE_POINTERS(RosRobot)
...
+      const std::shared_ptr<::ros::ServiceClient>& rosSwitchControllerServiceClient);
+
+  ///
+  /// Sets the ros joint mode command client which enables controller mode switching.
+  ///
+  /// \param[in] rosJointModeCommandClient ros joint mode command client.
+  void setRosJointModeCommandClient(
```
egordon, line 156: Similarly, I would not have an external class own the joint mode clients.

Have this function be `setRosJointModeClient(string serverName, vector joints = empty, other args = default)` and initialize the client internally. If not provided, use the list of joint names from the robot metaskeleton.

include/aikido/robot/ros/RosRobot.hpp
```diff
@@ -18,21 +21,198 @@ AIKIDO_DECLARE_POINTERS(RosRobot)
...
+      const std::shared_ptr<aikido::control::ros::RosJointModeCommandClient>& rosJointModeCommandClient);
+
+  ///
+  /// Starts the controller with name startControllerName.
+  ///
+  /// \param[in] startControllerName name of controller to start.
+  /// \return True if successful. Otherwise false.
+  bool startController(const std::string startControllerName);
```
egordon, line 164: start/stop/load/switch controllers should be private functions. We want external classes only handling controllers via the executor interface.

include/aikido/robot/ros/RosRobot.hpp
```diff
@@ -18,21 +21,198 @@ AIKIDO_DECLARE_POINTERS(RosRobot)
...
+  /// \return True if successful. Otherwise false.
+  bool switchControllerMode(
+      const hardware_interface::JointCommandModes jointMode);
+
+protected:
+  std::unordered_map<std::string, hardware_interface::JointCommandModes> mRosControllerModes;
+  std::unordered_map<std::string, std::string> mRosControllerNames;
+
+  // Ros node associated with this robot.
+  std::shared_ptr<::ros::NodeHandle> mNode;
```
egordon, line 204: As in other classes, don't have this be a shared pointer. Have it be a non-pointer object, and just initialize it in the constructor using the passed in object (if provided). If not provided, create it manually.

include/aikido/robot/ros/RosRobot.hpp
```diff
@@ -18,21 +21,198 @@ AIKIDO_DECLARE_POINTERS(RosRobot)
...
+
+  // Ros load controller service client.
+  std::shared_ptr<::ros::ServiceClient> mRosLoadControllerServiceClient;
+
+  // Ros switch controller service client.
+  std::shared_ptr<::ros::ServiceClient> mRosSwitchControllerServiceClient;
+
+  // Ros joint mode command client.
+  std::shared_ptr<aikido::control::ros::RosJointModeCommandClient>
+      mRosJointModeCommandClient; // default is null ptr so mode switching is
```
egordon, line 214: Move comment to next line. Also: `null ptr` -> `nullptr`

src/control/ros/RosJointModeCommandClient.cpp
```diff
@@ -0,0 +1,168 @@
...
+using std::chrono::milliseconds;
+
+//==============================================================================
+RosJointModeCommandClient::RosJointModeCommandClient(
+    ::ros::NodeHandle node,
+    const std::string& serverName,
+    std::vector<std::string> jointNames,
+    const std::chrono::milliseconds connectionTimeout,
+    const std::chrono::milliseconds connectionPollingPeriod)
+  : mNode(std::move(node))
```
egordon, line 19: Look at other classes. Don't move the node. Instead, copy it.

src/robot/ros/RosRobot.cpp
```diff
@@ -38,6 +52,262 @@ RosRobot::RosRobot(
...
+
+//=============================================================================
+void RosRobot::deactivateExecutor()
+{
+  if (!mActiveExecutor.empty())
+  {
+    if (!mRosControllerNames[mActiveExecutor].empty()
+        && !stopController(mRosControllerNames[mActiveExecutor]))
+    {
+      throw std::runtime_error(
```
egordon, line 133: Verify that this does not throw if the controller is already stopped.

Ideally, this would only throw if the controller is started, but cannot shutdown (which is a safety issue). If the controller is already stopped, this should just ignore the error.

src/robot/ros/RosRobot.cpp
```diff
@@ -38,6 +52,262 @@ RosRobot::RosRobot(
...
+//=============================================================================
+std::string RosRobot::registerExecutor(
+    aikido::control::ExecutorPtr executor,
+    std::string controllerName,
+    hardware_interface::JointCommandModes controllerMode,
+    std::string desiredName)
+{
+  if (!controllerName.empty() && !loadController(controllerName))
+  {
+    throw std::runtime_error("Could not load controller: " + controllerName);
```
egordon, line 149: As elsewhere. Warning, return false

src/robot/ros/RosRobot.cpp
```diff
@@ -38,6 +52,262 @@ RosRobot::RosRobot(
...
+    aikido::control::ExecutorPtr executor, 
+    std::string controllerName,
+    std::string desiredName)
+{
+  return registerExecutor(
+      executor, controllerName, hardware_interface::JointCommandModes::ERROR, desiredName);
+}
+
+//=============================================================================
+std::string RosRobot::registerExecutor(aikido::control::ExecutorPtr executor, std::string desiredName)
```
egordon, line 170: Remove this function

include/aikido/robot/ros/RosRobot.hpp
```diff
@@ -18,21 +21,198 @@ AIKIDO_DECLARE_POINTERS(RosRobot)
...
+      std::string desiredName);
+
+  ///
+  /// Add an executor to the inactive executors list.
+  /// Operations on this executor do not affect the controller or control mode.
+  /// Releases DoFs held by executor if held.
+  ///
+  /// \param[in] executor The Executor to add to the inactive list.
+  /// \return A robot-unique non-negative ID (negative implies failure)
+  std::string registerExecutor(aikido::control::ExecutorPtr executor, std::string desiredName = "") override;
```
egordon, line 124: Remove this. It should just call the parent class (no need to load a controller or register any controller names / modes)

include/aikido/robot/ros/RosRobot.hpp
```diff
@@ -18,21 +21,198 @@ AIKIDO_DECLARE_POINTERS(RosRobot)
...
+  /// loaded. Add an (executor, controller, control mode) to the inactive
+  /// (executor, controller, control mode) list Releases DoFs held by executor
+  /// if held.
+  ///
+  /// \param[in] executor The Executor to add to the inactive list.
+  /// \param[in] controllerName The corresponding controller to load and add to
+  /// the inactive list. \param[in] controllerMode The corresponding control
+  /// mode to add to the inactive list. \return A robot-unique non-negative ID
+  /// (negative implies failure)
+  std::string registerExecutor(
```
egordon, line 96: Have this interface generally match the parent class: `string registerExecutor(executor, desiredName, controllerName, controllerMode = ERROR)`.

This is not ambiguous: if there are 2 args, use the parent class. Else use this function.

include/aikido/robot/ros/RosRobot.hpp
```diff
@@ -18,21 +21,198 @@ AIKIDO_DECLARE_POINTERS(RosRobot)
...
+  /// Loads the controller. Throws a runtime_error if controller cannot be
+  /// loaded. Add an (executor, controller) to the inactive (executor,
+  /// controller, control mode) list Operations on this executor do not affect
+  /// the control mode. Releases DoFs held by executor if held.
+  ///
+  /// \param[in] executor The Executor to add to the inactive list.
+  /// \param[in] controllerName The corresponding controller to load and add to
+  /// the inactive list. \return A robot-unique non-negative ID (negative
+  /// implies failure)
+  std::string registerExecutor(
```
egordon, line 112: Remove this function. It should be handled by the suggestion above to make the 4-arg function have a default argument of ERROR.

src/robot/ros/RosRobot.cpp
```diff
@@ -38,6 +52,262 @@ RosRobot::RosRobot(
...
+  {
+    throw std::runtime_error(
+        "Could not start controller: " + mRosControllerNames[id]);
+  }
+
+  if (mRosControllerModes[id]
+          != hardware_interface::JointCommandModes::ERROR
+      && !switchControllerMode(mRosControllerModes[id]))
+  {
+    throw std::runtime_error(
```
egordon, line 199: Warning, return false

src/robot/ros/RosRobot.cpp
```diff
@@ -38,6 +52,262 @@ RosRobot::RosRobot(
...
+  // Validate input
+  if (mExecutors.find(id) == mExecutors.end())
+  {
+    return false;
+  }
+
+  if (!mRosControllerNames[id].empty()
+      && !startController(mRosControllerNames[id]))
+  {
+    throw std::runtime_error(
```
egordon, line 191: warning, return false

src/robot/ros/RosRobot.cpp
```diff
@@ -38,6 +52,262 @@ RosRobot::RosRobot(
...
+//=============================================================================
+bool RosRobot::activateExecutor(std::string id)
+{
+  // Deactivate active executor
+  deactivateExecutor();
+
+  // Validate input
+  if (mExecutors.find(id) == mExecutors.end())
+  {
+    return false;
```
egordon, line 185: add a warning here (not sure if Robot does this, but would be a good idea there as well)

egordon (changes requested):
In addition to the comments below. This breaks test_Robot (because of the interface changes). This will also break `libada` (due to the interface change)

(1) Fix this PR
(2) Fix the test, and make sure that test coverage of `Robot` is not reduced
(3) Make a PR in libada that *just* adjusts to this interface (i.e. minimal code changes to prevent breakage)

egordon:
@RKJenamani please resolve conflicts

egordon:
So when I was talking about ownership of the Client objects, I wasn't talking necessarily about getting rid of the setters. You completely got rid of the ability to modify any parameters about the controller service clients.
You can have shared_ptr to both mode and controller objects and still have them owned by the RosRobot class.

Here is what I would recommend (and I'll review to this effect):
(1) Keep separate setters for both mode and controller manager service clients. But don't have the argument just be a shared_ptr (or keep it as a way for maximum customization, but don't make it the only option, see below)
(2) The only parameter that matters for the controller_manager service clients is the namespace (i.e. the prefix before "cotnroller_manager/*"). So have those clients default to nullptr (which disables controller switching) and have the setter just ask for the namespace (default "")
(3) Have the setter for the mode switch that takes takes in the 2 necessary parameters: controller name and string list of joints. I would completely get rid of the "joint_mode" string default (since that is definitely non-standard).
(4) Have a single boolean flag in registerSubrobot that, if true, forwards mode and controller clients, and otherwise starts them as nullptr.

egordon:
With that said, I think this is actually okay to merge as-is (and move the above to an issue / future PR) just to keep our PRs small and easy to review.

I'll make sure this doesn't break Gen2 demos and then we can approve and merge

egordon:
Does not break Feeding Demo, will merge after CI is fixed (see #631 )

egordon (approved):
See above comment, approving for now since it doesn't break existing demos, but may edit later

PR #622, opened 2022-09-08
Check ability of Robot class to handle R0 joints

egordon:
Previously, the Robot class only worked with "controlled MetaSkeletons" by default (i.e. the MetaSkeleton stripped of all fixed and R0 joints). However, our planners do not make that distinction by default. Therefore, there were number-of-joint mismatches between many generated trajectories and their corresponding MetaSkeletons when checking state space compatibility.

The solution was to either:
(1) Update the codebase to strip all R0 joints from any MetaSkeletons we plan on.
OR
(2) Just remove the concept of joints entirely from the MetaSkeletonStateSpace and only deal with DoFs.

Because DART already deals only with DoFs in functions like "getPositions", this push goes for the latter.

As a consequence, some conversion code that claims to only work with R1 and SO2 joints was updated to allow R0 joints (which make no change to code execution anyway).

***

### DIFF
--- a/include/aikido/robot/util.hpp
+++ b/include/aikido/robot/util.hpp
@@ -29,7 +29,7 @@ namespace util {
 const std::vector<double> defaultVFParams{
     0.03,  // distanceTolerance
     0.004, // positionTolerance
-    0.004, // angularTolerance
+    0.01,  // angularTolerance
     0.01,  // initialStepSize
     1e-3,  // jointlimitTolerance
     1e-3,  // constraintCheckResolution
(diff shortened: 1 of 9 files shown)


include/aikido/robot/util.hpp
```diff
@@ -29,7 +29,7 @@ namespace util {
 const std::vector<double> defaultVFParams{
     0.03,  // distanceTolerance
     0.004, // positionTolerance
-    0.004, // angularTolerance
+    0.01,  // angularTolerance
```
amalnanavati, line 32: Just flagging that this change to default params got included in the push, in case it was unintentional

amalnanavati (approved):
LGTM

include/aikido/robot/util.hpp
egordon: Yeah, the old tolerance was too strict for JACO 2 (it was good for Gen3)

PR #629, opened 2022-12-30
Remove bug where Kunz hangs on empty trajectory

egordon:
Previously, in the case of an empty trajectory (which is possible for an Offset planner with 0 offset or a configuration planner with goal == start), the resulting trajectory was length 0.

In this case, any postprocessor should return a valid (but empty) trajectory.

However, Kunz's Trajectory class initially timed the trajectory using the velocity without checking for 0 velocity. If there is 0 velocity but non-0 position, the trajectory is invalid (this should never happen). But if there is 0 velocity AND 0 change in position, the trajectory is valid but empty, and Kunz shouldn't perform the division (which results in a NaN endTime, which in turn leads to a *VERY LARGE* sequence of waypoints when naively constructed with aikido::common::StepSequence, which in turn causes the program to loop nearly infinitely).

In addition, the Robot class has been modified to gracefully handle valid and empty trajectories (namely: not executing them).

***


amalnanavati (approved):
LGTM

PR #633, opened 2023-03-10
Add control and control_ros dependencies to ros robot

vbhatt-cs:
After the controller switching support was added, packages that created a `RosRobot` had linker errors during `catkin build` which is fixed by these changes to `CMakeLists.txt`

The changes were tested by building a [simple example](https://github.com/vbhatt-cs/aikido_tutorials/tree/master/src/aikido_tutorials) that instantiated a `RosRobot`.

***


egordon (approved):
LGTM, thanks!

